PR #488, opened 2015-03-21
unix LogImpl: ensure pFile is NULL if no filename was given

stromnet:
Otherwise we use an uninited pFile in Write, potentially causing crash


Fishwaldo:
Could you also please update the ChangeLog with your change and resubmit the request? (just follow the format in there already).

Fishwaldo:
(until we get someother form of Commit to Changelog that is reasonable, please do so for all future commits as well... cheers)

stromnet:
There!

PR #494, opened 2015-03-30
adjust spelling of receive

alteholz:


bytedreamer:
'i' before 'e' except after 'c'

Fishwaldo:
As this changes log messages, it would also require updates to the LogAnalyzer definitions. I'll get to it eventually but it will be low priority as it's nothing that is user facing.

PR #495, opened 2015-03-30

nechry:
a new definition for Figaro FGD211


ualex73:
The changelog was missing, so i manually added it to the repo.

PR #500, opened 2015-04-04
Update fgk001.xml

nechry:
according to the official doc, the old parameter 10 is wrong, the parameter 11 not exist. and a new paramter 15 is present
http://www.fibaro.com/manuals/en/FGK-10x/FGK-10x-EN-A-v1.01.pdf

### DIFF
--- a/config/fibaro/fgk001.xml
+++ b/config/fibaro/fgk001.xml
@@ -45,14 +45,12 @@
 			<Item label="Group 1 not sent, Group 2 sent." value="2" />
 		</Value>
 
-		<Value type="byte" genre="config" instance="1" index="10" label="Interval between successive readings of temperature sensors" min="1" max="255" value="20">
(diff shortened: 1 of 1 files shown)


nechry:
HI, 
the new id for this device <Product type="0700" id="4000" name="FGK101 Door Opening Sensor" config="fibaro/fgk001.xml" /> have been introduced last year.
I thin we can manage to set a specific configuration file for this release.

In two revision the parameter 11 did not exist, we have to remove it.

see initial v2.1- v2.3:
http://www.fibaro.com/manuals/en/FGK-10x/FGK-10x-EN-A-v1.00.pdf
and the updated revision v2.5
http://www.fibaro.com/manuals/en/FGK-10x/FGK-10x-EN-A-v1.01.pdf

the v2.5 introduce parameters 10 and 15.

Fishwaldo:
Hi,

Thanks

PR #504, opened 2015-04-06
implemented CRC16Encap support

xs4:
Implemented CRC16Encap support based on the CRC-CCITT algorithm.

### DIFF
--- a/cpp/src/command_classes/CRC16Encap.cpp
+++ b/cpp/src/command_classes/CRC16Encap.cpp
@@ -35,8 +35,25 @@
 
 using namespace OpenZWave;
 
+//
+// CRC-CCITT (0x1D0F)
+//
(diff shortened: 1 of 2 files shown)


Fishwaldo:
Which Devices are actually using the CEC16Encap? I've seen it appear on NIF frames, but never actually seen it in use.

xs4:
The BeNext Heating control (http://www.benext.eu/producten/heating-control/) uses the CEC16Encap.
I used that device during development and testing. 

(From what I read on the openhab mailing list the fibaro smoke sensor fgss001 also uses the CEC16Encap. see https://groups.google.com/forum/#!topic/openhab/xg_j2tyGwLc )

killercode:
NorthQ also uses it

Fishwaldo:
Merged. Thanks

Issue #416, opened 2015-03-14
ValueShort initializes the Value Base class as a Byte


GoogleCodeExporter:
```
in the constructor of ValueShort:


Value( _homeId, _nodeId, _genre, _commandClassId, _instance, _index, 
ValueID::ValueType_Byte, _label, _units, _readOnly, _writeOnly, false, 
_pollIntensity ),


We should change to ValueID::ValueType_Short but I need to check what if 
anything, will break here. :(
```

Original issue reported on code.google.com by `jus...@dynam.ac` on 12 Dec 2014 at 3:30

xs4:
ValueSchedule also has the same issue. ie uses ValueID::ValueType_Byte iso ValueID::ValueType_Schedule as argument in the constructor.

Fishwaldo:
Argh... Simple Fix be Not!

The Value Class doesn't handle a lot of the different ValueTypes. Particulary in the VerifyRefreshedValue function.

Fishwaldo:
This should be fixed now.

PR #506, opened 2015-04-06
ValueShort: use short iso byte

xs4:
The value class ValueShort uses a ValueID::ValueType_Byte as argument for the _type argument of the Value constructor. This should have been a ValueType_Short.

### DIFF
--- a/cpp/src/value_classes/ValueShort.cpp
+++ b/cpp/src/value_classes/ValueShort.cpp
@@ -56,7 +56,7 @@ ValueShort::ValueShort
 	int16 const _value,
 	uint8 const _pollIntensity
 ):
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Already tracked as part of Issue #416. Just need some confirmation that its not breaking any existing setup etc.

Frayman67:
I agree with this commit.
In addition, in the same class, the SetFromString() method implementation will reject a string with negative value such as "-10000", because of the uint32 cast of the atoi result and the comparison to 32768. Are negative values forbidden?

Fishwaldo:
The z-wave protocol is all unsigned. There are a few vendors that screwed up early on and did signed values, but technically the protocol doesn't support it.

Fishwaldo:
This is fixed in the commit I just did.

PR #508, opened 2015-04-07
Update fgfs101.xml association group 3

nechry:
According to documentation, only one device for groupe 3:
3-rd Association Group reports the device status and allows for assigning single device only (the main controller by default - the device reports its status to the main controller).


Fishwaldo:
Merged. Thanks

Issue #111, opened 2015-03-14
OZW goes into 100% CPU load if the Aeon USB Stick 2 is unplugged (only restarts helps)


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Start OZW e.g. with MinOZW and get the devices discovered
2. Unplug the USB Stick
3. Check with "top" and the MinOZW is consuming 100% CPU load

What is the expected output? What do you see instead?
Not 100% CPU load

What version of the product are you using? On what operating system?
Unbuntu 10.04
Open Z-Wave r556

Please provide any additional information below.


```

Original issue reported on code.google.com by `uAle...@gmail.com` on 3 Nov 2012 at 11:37
- Blocking: #189

dsoulayrol:
Hi. It seems this issue was abandoned half-fixed. As reported earlier, there are two problems. First, a 100% Cpu spin in SerialImpl, and then the absence of an event towards the user of the Manager object.

I need to address both those difficulties, so I'd like to know if anyone has made progress upon the proposed patches. Thanks.

dsoulayrol:
Hello.

[Here](https://github.com/OpenZWave/open-zwave/files/277690/serialerror.txt) is a refreshed version of the patch proposed on 2015 March, the 14th. It still cannot be considered complete because I do not use the recover capabilities of the Driver and the Controller for now. When the stream is broken, the Driver emits a DriverFailure notification, and my program stops upon reception of this event.

However, I think there is not much to do so as to get the Driver up again. I experience no more CPU spins, but I had some problems on reconnection and I suspect some purge is missing before commands can be transmitted correctly again.

Fishwaldo:
Thanks. I'll look over it soon.

I think trying to recover from a "failed" serial port is too hard. I'll probably just try to tweek it a bit, so we also unload the Driver as well...

julienw:
I'm very interested by this issue, please tell me if I can help at something.

julienw:
note: the link to the issue in https://github.com/OpenZWave/open-zwave/wiki/FAQ points incorrectly to google code.

dhylands:
The real secret is that the serial port must be closed once it has "gone away". I've been using udev in other programs to detect serial ports being added and removed, and then just close the handle when the close is detected. This will typically cause any reads/writes which happen after the close to fail, and if timeouts are used, then in-progress reads/writes will timeout.

[...]

PR #513, opened 2015-04-12
Fix for CPU-spinning on device disconnection

stromnet:
Hi,
this is an attempt to fix the issue with disconnected devices, which currently isn't handled very well.

If a serial port device was disconnected, we got stuck in read loop burning CPU cycles and having a unusable device.

On disconnect (at least on FreeBSD), select() will return 1 indicating
FD ready to read. The subsequent read() would however return 0. This would would go on until the process was killed, yielding unusable device, and high CPU usage.
With this patch, we break out of the read loop and re-init the connection if this condition occurs.

This has been tested on FreeBSD 10.1 with an Aeon Labs 2 dongle. On this setup, disconnecting and then reconnecting the device works without any problems, and communication seems to be restored. I'm not sure if there are any higher-level issues with this though, there may be buffers which should be flushed on re-init or similar?

Testing should be done on other supported platforms to verify that it works.


Fishwaldo:
Hi,
This issue has already been discussed in Issue 111. You patch only addresses a small part of it. We still need to propagate the disconnect up the stack (to the Driver and then to the Manager, and of course consider a notification to the Application).

Just re-opening the Serial Port is not the complete fix here. We would probably need to also re-init the Controller. 

Also, I think most times, the issue here is going to be more along the lines of the USB stick going unresponsive or falling out. At the moment, if the stick is not re-inserted, we will just spin, and instead of consuming CPU, we will start consuming memory and the message queues build up. :)

I will ultimately merge this, but I think it needs more work to address the issue. Thanks for getting started on this.

stromnet:
Ok, kind of what I expected, but this is a start at least!

An alternative to the current behaviour would just be to log "Crap, we're dead" and just stop reading (but not reconnect). Then at least we have notified the user that there is a problem, and we won't waste CPU/power for no good reason.

Fishwaldo:
My plan is to issue a Driver Failed notification and unload the Driver. There are a few race conditions though that make it tricky. (One being the driver thread handles the dispatch of notifications etc.) 

We will probably have to put a condition variable in the Serial API (as it's threaded) and in the main loop of the driver check it and then call the notification, and subsequently signal the exit event.

stromnet:
Sounds like a good idea, this way the application can re-init the library (if possible, not too familiar with the details. Otherwise just restart).

Fishwaldo:
Hi,

Thanks

stromnet:
Leaving this as-is as this seems discussed in #111 which is still open.

PR #516, opened 2015-04-12
Patch for Philio pst02-1c

ugoviti:
This device still not working because https://github.com/OpenZWave/open-zwave/issues/352 worked in branch https://github.com/OpenZWave/open-zwave/tree/Issue-352-Security

Anyway I pull the patches to add the support for later merge.

King Regards


Fishwaldo:
Merged. Thanks. FYI - No need to specify instances, and You should try to keep the Group names short (as most apps display it as a control). Also, I set Group 2 auto to false, as we should get the real CC reports from Group 1.

PR #524, opened 2015-04-20
Non-blocking open() on OS X

tternes:
When using the Aeon Labs Z-Stick on OS X, it appears to be necessary to call `open()` with `O_NONBLOCK` specified. Otherwise, the `open()` call never seems to return.

From the testing I've done, this doesn't appear to have any adverse affects, but some feedback here would be great, if perhaps I've overlooked something.

### DIFF
--- a/cpp/src/platform/unix/SerialControllerImpl.cpp
+++ b/cpp/src/platform/unix/SerialControllerImpl.cpp
@@ -184,7 +184,7 @@ bool SerialControllerImpl::Init
 	
 	Log::Write( LogLevel_Info, "Trying to open serial port %s (attempt %d)", device.c_str(), _attempts );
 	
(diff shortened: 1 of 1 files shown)


Fishwaldo:
This doesn't seem right.... the SerialPortImpl runs in a dedicated "blocking" thread to read/write to the serial port... Setting it to Non-Blocking would screw up those semantics. 

Something else is going wrong with your install... Permissions perhaps? is another application opening the serial port etc?

tternes:
You're right, @Fishwaldo. I was unaware of the subtle differences between `cu.` and `tty.` devices (I was attempting to use `/dev/tty.SLAB_USBtoUART`).

Switching my configuration to `/dev/cu.SLAB_USBtoUART` works great.

Thanks for the nudge in the right direction.

Fishwaldo:
No Problem. Glad you figured it out. :)

PR #531, opened 2015-04-26
Update recessed_doorsensor.xml

nechry:
The parametrer 121 seem wrong. I set as door window sensor v2. With right size and max value


Fishwaldo:
Is this for the Gen5 Door Sensor?
If so, can you redo the pull request against the Issue-352 Branch and the recessed_doorsensor_gen5.xml instead?

nechry:
is only the max="65535" and size="4" So I will close this one and create a new pull request

PR #534, opened 2015-04-27
Adds new product IDs for Aeon Labs Recessed Door Sensor

xxv:
I'm not sure why this is different than the existing entry, but it matches the datasheet linked linked off https://github.com/OpenZWave/open-zwave/wiki/Technical-Documents and the hardware that I have. The device is labeled "Gen 5". There don't appear to be any configuration changes.

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -48,6 +48,7 @@
 		<Product type="0002" id="001c" name="Home Energy Meter G2" config="aeon_labs/hemg2.xml"/>
 		<Product type="0002" id="001d" name="Door/Window Sensor" config="aeon_labs/doorwindow.xml"/>
 		<Product type="0002" id="0036" name="Recessed Door Sensor" config="aeon_labs/recessed_doorsensor.xml"/>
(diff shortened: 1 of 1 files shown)


config/manufacturer_specific.xml
```diff
@@ -63,6 +64,10 @@
 		<Product type="0003" id="004b" name="Smart Energy Switch v2" config="aeon_labs/ses2.xml"/>
 		<Product type="0004" id="0025" name="Z-Wave Repeater"/>
 		<Product type="0004" id="0050" name="DSD31 Siren Gen5" config="aeon_labs/dsd31.xml"/>
+        <!-- US-market products start with type prefix "01" -->
```
xxv: I added a comment here instead of putting it in the name, as the product country distinction is likely unimportant to the end-user who will be seeing the "name" value displayed.

Fishwaldo: thats fine. I might drop Aeon Labs a note and ask to get all their product codes etc.

Fishwaldo:
I pulled this into the Issue-352 Branch and updated it as a Gen5 entry instead. Thanks :)

PR #537, opened 2015-05-01
Update ses2.xml

nechry:
is suppose to have 2 associations group


Fishwaldo:
this is already done in the Issue-352 Branch.

PR #541, opened 2015-05-02
Add swiidIter configuration file

nechry:
voilà, the right way.


Fishwaldo:
Just FYI, Group Auto's should generally be true - Otherwise OZW will not get reports about status changes from the device.

nechry:
In documentation I read just assing the group 2 to the controller. Is why I only set auto at true for this group.

PR #542, opened 2015-05-02
Update recessed_doorsensor.xml

nechry:
fix the max="65535" and size="4" attributs for parameter 121.
a default value of 256 whit a max set to 1... I also add the size as wrote in doc.

### DIFF
--- a/config/aeon_labs/recessed_doorsensor.xml
+++ b/config/aeon_labs/recessed_doorsensor.xml
@@ -30,7 +30,7 @@
               Default setting: 86640
             </Help>
         </Value>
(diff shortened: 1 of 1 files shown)


Fishwaldo:
wrong file still... 
I'm assuming the recessed_doorsensor.xml is for devices pre-Gen5, so I created a new recessed_doorsensor_gen5.xml for the new Gen5 device. :)

nechry:
this for regular and gen5 recessed doorsensor. is just the attributs.  

Or I misunderstood something

Fishwaldo:
I wanted to keep the old device and new gen5 device config files seperate (in case a change is needed to one of them, we wont break the other one).

Anyway, I tested this tonight, and updated the config file, so no need to pull this request anymore...

Issue #461, opened 2015-03-14
Dotnet: GetValueAsDecimal is culture sensitive


GoogleCodeExporter:
```


ZWManager::GetValueAsDecimal throws FormatException on a computer that uses "." 
as thousands separator.

Not sure what the defined format is of the underlying protocol but adding 
NumberStyles and a en-GB culture solved the problem for me. Like this:

String^ decimal = gcnew String(value.c_str());
CultureInfo^ culture = gcnew CultureInfo("en-GB");
o_value = Decimal::Parse( decimal, NumberStyles::Float | 
NumberStyles::AllowThousands, culture); 

Using the latest snapshot 1.3.1021


```

Original issue reported on code.google.com by `and...@anderspersson.com` on 18 Feb 2015 at 10:10

Fishwaldo:
@OpenZWave/openzwave-dotnet-uwp You guys might want to look at this if you have not already.

dotMorten:
@Fishwaldo Since this was deprecated, I removed the method altogether.

dotMorten:
Actually question on this:

> In the C++ code, we query the current locale and set the relevant NumberStyle 
there

Does this mean, the string value we get back changes with the locale? I'm curious why you say this too: 

> hardcoding the Locale  is not the way to go

IMHO that IS the way to go, because it makes the value returned predictable and easily parseable. The alternative would be to actually return a number type, rather than a string.

Fishwaldo:
Deprecated? It defiantly should not be deprecated. :) (but history has it as GetValueAsFloat... idk why the naming Decimal/Float is mixed up). 

Regarding Locale - Eventually Id like to support localization etc of the texts etc in OZW. 

So, yes, the string returned should be "correct" with the current Locale. If you are trying to do something "computational" with a Decimal Value, you should use the GetValueAsFloat. The "translation" to a string for ValueDecimal's is a convience function IMHO.

dotMorten:
Right. The "old" .NET library had it mentioned for deprecation, and there is no GetValueAsDecimal in the wrapped Manager class, so I didn't really see the need to have it in the .NET wrapper either? I'd prefer things are one-to-one.

GetValueAsFloat is indeed missing, and I'll get that added ASAP

dotMorten:
@Fishwaldo Actually I'm a little confused, the doc for GetValueAsFloat, but the value type is Decimal? 
[...]

PR #551, opened 2015-05-10
Fix : Dotnet GetValueAsDecimal

bufferUnderrun:
Hi,

Following the issue https://github.com/OpenZWave/open-zwave/issues/461, could you integrate this patch in order to fix the FormatException with the dotnet wrapper.

Thank,


Fishwaldo:
As noted in the issue comment hard coding to en-gb is not the right solution. What if someone is using the library in a different region. You need to detect the current local and use that instead. 

Hence, I won't merge this. Someone with better understanding of dotNet needs to figure that out. 

(Hint here? http://stackoverflow.com/questions/2288383/where-is-the-system-locale-culture-set-for-net)

PR #553, opened 2015-05-11
Issue 352 security - Updated .Net Wrapper and OZWForm to new methods

JKlessens:
Edited the .Net wrapper and OZWForm to work with the new methods. I have added a new menu item for adding devices securely. I have no other devices, so I can't test if everything works correctly or not. For now my changes seem to work as they should, however I have had some problems with assertions failing in "output.c", but this does not seem to fail every time. I believe this is not due to my code, so I let it be. 

Finally, my code should be tested more extensively, by someone with more devices and (maybe) more experience with the project, to test whether everything actually works as it should!


Fishwaldo:
Cheers. But can you remove all the Generated Objects/Bin files etc from your commit (git rm)
eg the Debug, src/bin etc

PR #554, opened 2015-05-13
Reverted and readded changed files

JKlessens:
Update:
Had to revert to a new branch and re-add all my changed files, because my working branch of my fork got completely messed up. This new branch should now only contain the necessary files and updates.

Old updates: 
(from new to old)
*Updated OZWForm with a new menu item to add secure devices

*Edited .Net wrapper and OZWForm to use the new methods:
Edited the .Net wrapper and OZWForm to work with the new methods. For
now I have used only one way of adding nodes, which always uses the
security enabled way. I have no other devices, so can't test if this is
right or not. Furthermore, I split the implementation and definition of
the methods between the ZWManager.cpp and ZWManager.h. But the methods
are just one line calls to the manager, and as some other methods like
that in ZWManager are just implemented in the header file, this might be
part of the coding standard? If so, I'll have to change that.

When BeginControllerCommand is fully deprecated everything should be
looked over and edited further. For now my changes seem to work as they
should, but I have had some problems with assertions failing in
"output.c", when I was trying to remove a node, but it does not seem to
fail every time. I believe this is not due to my code, so I let it be.

Finally, my code should be tested more extensively, by someone with more
devices and (maybe) more experience with the project, to test whether
everything actually works as it should!

*Update .Net wrapper for new Manager API's:
Added Methods to ZWManager corresponding to the new API methods in
Manager.
ToDo: Change the examples so they use the new API methods (Use
Notification::Type_ControllerCommand as the reply for the API methods.
The (OpenZWave::Driver::ControllerState)_notification->GetEvent(),
(OpenZWave::Driver::ControllerError)_notification->GetNotification() can
be used as parameters for the original callback method, I believe)!

### DIFF
--- a/dotnet/examples/OZWForm/src/ControllerCommandDlg.cs
+++ b/dotnet/examples/OZWForm/src/ControllerCommandDlg.cs
@@ -9,24 +9,29 @@ using OpenZWaveDotNet;
 
 namespace OZWForm
 {
-	public partial class ControllerCommandDlg : Form
-	{
-        static private ZWManager m_manager;
-        static private ControllerCommandDlg m_dlg;
-        static private UInt32 m_homeId;
-        static private ManagedControllerStateChangedHandler m_controllerStateChangedHandler = new ManagedControllerStateChangedHandler(ControllerCommandDlg.MyControllerStateChangedHandler);
-		static private ZWControllerCommand m_op;
-        static private Byte m_nodeId;
-		static private DialogResult result;
+    public partial class ControllerCommandDlg : Form
+    {
+        private static ZWManager m_manager;
+        private static ControllerCommandDlg m_dlg;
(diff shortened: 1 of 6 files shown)


Fishwaldo:
Thanks. This is merged. 
2 quick questions: Is there a way to mark a method as depreciated in dotNet code? 
and did you see my last update where I added the remaining ControllerCommands (I think so right?)

JKlessens:
For marking a method as deprecated/obsolete
http://stackoverflow.com/questions/1759352/how-do-i-mark-a-method-as-obsolete-deprecated

I did see the update and will add the rest of the controllercommands to the .Net wrapper ASAP, somewhere this week. I will let you know and make a PR when all the CC are handled in the new way by the wrapper!

PR #555, opened 2015-05-13
Issue 352 security

JKlessens:
Updated the .Net wrapper and OZWForm to use the last of the new ControllerCommand methods, fully deprecating/obsoleting the BeginControllerCommand method.

Also added/generated headers and comments for all OZWForm files and some wrapper files.


cpp/build/windows/winversion.cpp
```diff
@@ -1,3 +1,3 @@
-unsigned short ozw_vers_minor = 1;
```
Fishwaldo: Drop this file from your commit. Its auto Generated.

dotnet/examples/OZWForm/src/ConfigurationWakeUpDlg.cs
```diff
@@ -1,14 +1,39 @@
+//-----------------------------------------------------------------------------
```
Fishwaldo: Where did you copy the header from? 
The example apps should be under GPL license, and only the libary code under LGPL

Fishwaldo:
See my comments above about the headers.... might need to revert or fix them up.

JKlessens:
Apologies, I just copied them from the library code I think. Didn't realize the example apps were under another license. I will fix it on Friday and do another PR. 

FYI the winversion.cpp is not autogenerated by Visual Studio. VS2013 won't even build without it! But if necessary I'll remove it. I just thought someone had committed it wrongly as it was committed as winversion.tmpl.

Fishwaldo:
In the 2010 vs project file, we call the following to generate the winversion.cpp:
CALL $(SolutionDir)\..\GIT-VS-VERSION-GEN.bat "$(ProjectDir)\" "..\winversion.cpp

the GIT-VS-VERSION-GEN.bat creates the winversion.cpp based on the latest git hash etc. 

I guess when you update the project file to VS2013 it didn't migrate over? From member, it was called as part of a "PreBuild" stage or something.

JKlessens:
Modified the headers, and removed the winversion.cpp. FYI I checked and MinOZW also has the LGPL header. Not sure if it is considered an example app, but I thought I would let you know.

I will look at what is going wrong with the generation of winversion.cpp when I have time. I think its either because of the convert to VS2013, or because OZWForm is in a different solution, which might make $(SolutionDir) different.

Fishwaldo:
yeah, I just noticed there is a non fatal error on the windows build logs:
http://bamboo.my-ho.st/bamboo/download/OZW-OZWWXP0-JOB1/build_logs/OZW-OZWWXP0-JOB1-28.log

at 14-May-2015 23:12:17, it works fine when building the c++ library (\cpp\build\windows\vs2010\\) , but for some reason, at build	14-May-2015 23:13:53 it fails. In fact it shouldn't be called the 2nd time, as thats the dotnet build (\dotnet\build\vs2010\\), which I don't think incorporates the version info. (that might be something missing from the dotnet API as well, I can't remember!)

It might be we are using the wrong macro's?

JKlessens:
Do you mean that you are getting a non fatal error due to my changes in the code? Because that shouldn't be possible. Besides, the problem with winversion.cpp and the assertion error were already occurring for me before I started editing anything!

Furthermore, I don't know if I'll have time to look at the winversion.cpp problem this week, as I am gonna be fairly busy. I am gonna try looking at it as soon as possible, but it is probably best if you merge this request before that.

Fishwaldo:
No - Sorry, I didn't mean you changes caused that error, only that I started noticing the problem. Anyway, I've merged a fix for wrong paths just now from Douglasbeck. Hopefully that fixes the error I noticed in the Log Files.

Feel free to take your time. We are all pretty busy :)

PR #560, opened 2015-05-15
[RFC] MultiInstanceAssociation support

xs4:
This is an initial attempt to add MultiInstanceAssociation support, and not intended to actually be pulled. 
Since there is some impact on the manager and driver classes I would like some feedback on what the best approach is to continue this work.
Thus far adding a instance association works. 

Some questions:
- do we want both version of 
```
		void AddAssociation( uint8 const _nodeId );
		void AddAssociation( uint8 const _nodeId, uint8 const _instance );
```
or only the last where passing 0x00 as instance would give you the original behavior?

- do we want method signature like this: 
```
		void AddAssociation( uint8 const _nodeId, uint8 const _instance );
```
or
```
		void AddAssociation( InstanceAssociation const _association);
```

On my todo list is:
- testing removing
- restoring COMMAND_CLASS_ASSOCIATION_COMMAND_CONFIGURATION support
- reporting association with instances in manager/driver 
- check why zwcfg_0xxxxx.xml does not contain instances associations
- ???

### DIFF
--- a/cpp/src/Group.cpp
+++ b/cpp/src/Group.cpp
@@ -106,7 +106,7 @@ Group::Group
 {
 	int intVal;
 	char const* str;
-	vector<uint8> pending;
+	vector<InstanceAssociation> pending;
 
 	if( TIXML_SUCCESS == _groupElement->QueryIntAttribute( "index", &intVal ) )
 	{
@@ -137,8 +137,18 @@ Group::Group
 		char const* elementName = associationElement->Value();
 		if( elementName && !strcmp( elementName, "Node" ) )
 		{
-			if (associationElement->QueryIntAttribute( "id", &intVal ) == TIXML_SUCCESS) 
-				pending.push_back( (uint8)intVal );
+			
+			if( associationElement->QueryIntAttribute( "id", &intVal ) == TIXML_SUCCESS )
+			{
+				InstanceAssociation association;
+				association.m_nodeId = (uint8)intVal;
+				if( associationElement->QueryIntAttribute( "instance", &intVal ) == TIXML_SUCCESS )
+					association.m_instance = (uint8)intVal;
+				else
+					association.m_instance = 0x00;
+
+				pending.push_back( association );
+			}
 		}
 
 		associationElement = associationElement->NextSiblingElement();
@@ -177,7 +187,7 @@ void Group::WriteXML
 		
 		snprintf( str, 16, "%d", it->first.m_nodeId );
 		associationElement->SetAttribute( "id", str );
-//		if (it->first.m_instance != 0)
+		if (it->first.m_instance != 0)
 		{
 			snprintf( str, 16, "%d", it->first.m_instance );
 			associationElement->SetAttribute( "instance", str );
--- a/cpp/src/Group.cpp
+++ b/cpp/src/Group.cpp
@@ -231,7 +231,6 @@ void Group::AddAssociation
 	{
 		if( Node* node = driver->GetNodeUnsafe( m_nodeId ) )
 		{
-			// TODO add options to use Association when _instance == 0
 			if( MultiInstanceAssociation* cc = static_cast<MultiInstanceAssociation*>( node->GetCommandClass( MultiInstanceAssociation::StaticGetCommandClassId() )))
 			{
 				cc->Set( m_groupIdx, _nodeId, _instance );
@@ -263,7 +262,6 @@ void Group::RemoveAssociation
 	{
 		if( Node* node = driver->GetNodeUnsafe( m_nodeId ) )
 		{
-			// TODO add options to use Association when _instance == 0
 			if( MultiInstanceAssociation* cc = static_cast<MultiInstanceAssociation*>( node->GetCommandClass( MultiInstanceAssociation::StaticGetCommandClassId() ) ) )
 			{
 				cc->Remove( m_groupIdx, _nodeId, _instance );
@@ -403,7 +401,7 @@ uint32 Group::GetAssociations
 		return 0;
 	}
 
-	uint8* associations = new uint8[numNodes]; // TODO
+	uint8* associations = new uint8[numNodes]; // room for all associations, we only need room for the associations without instance
 	uint32 i = 0;
 	for( map<InstanceAssociation,AssociationCommandVec>::iterator it = m_associations.begin(); it != m_associations.end(); ++it )
 	{
--- a/cpp/src/Group.cpp
+++ b/cpp/src/Group.cpp
@@ -193,14 +193,13 @@ void Group::WriteXML
 //-----------------------------------------------------------------------------
 bool Group::Contains
 (
-	uint8 const _nodeId
+	uint8 const _nodeId,
+	uint8 const _instance
 )
 {
-//	map<uint8,AssociationCommandVec>::iterator it = m_associations.find( _nodeId );
-//	return( it != m_associations.end() );
 	for( map<InstanceAssociation,AssociationCommandVec>::iterator it = m_associations.begin(); it != m_associations.end(); ++it )
 	{
-		if (it->first.m_nodeId == _nodeId)
+		if ((it->first.m_nodeId == _nodeId) and (it->first.m_instance == _instance))
 		{
 			return true;
 		}
@@ -214,28 +213,6 @@ bool Group::Contains
 //-----------------------------------------------------------------------------
 void Group::AddAssociation
 (
-	uint8 const _nodeId
-)
-{
-	if( Driver* driver = Manager::Get()->GetDriver( m_homeId ) )
-	{
-		if( Node* node = driver->GetNodeUnsafe( m_nodeId ) )
-		{
-			if( Association* cc = static_cast<Association*>( node->GetCommandClass( Association::StaticGetCommandClassId() ) ) )
-			{
-				cc->Set( m_groupIdx, _nodeId );
-				cc->QueryGroup( m_groupIdx, 0 );
-			}
-		}
-	}
-}
-
-//-----------------------------------------------------------------------------
-// <Group::AddAssociation>
-// Associate a node with this group
-//-----------------------------------------------------------------------------
-void Group::AddAssociation
-(
 	uint8 const _nodeId,
 	uint8 const _instance
 )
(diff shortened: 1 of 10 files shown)


Fishwaldo:
interesting... :)
are you on the mailing list? I'd prefer to discuss there!

xs4:
No, only as a reader, I have not yet subscribed.

openzwave-dev-watcher:
Ok, i have subscribed to the mailing list. 

Op vrijdag 15 mei 2015 17:18:25 UTC+2 schreef xs4:

Fishwaldo:
ok. Guiding principle when adding/modifying stuff right now is to maintain API compatibility. 
we could do 
```
void AddAssociation( uint8 const _nodeId, uint8 const _instance = 0x00);
``` 
as the signature, and nothing broken. :)

My bigger question though is how this interacts with our existing "auto associate" groups for reports with the normal association class. 

I dont "think" I have a multiinstance association device... and I see this class is probably more aimed at setting up associations with other node rather than gateways (as we are interested in all associations).

So I'm wondering if this should instead be just a stand alone CommandClass, or really integrated into our groups. 

(take the rest of the converstation to this... email is easier for me at the moment!)

openzwave-dev-watcher:
Op vrijdag 15 mei 2015 17:33:55 UTC+2 schreef Justin Hammond:
>
> ok. Guiding principle when adding/modifying stuff right now is to maintain 
> API compatibility. 
> we could do 
>
> void AddAssociation( uint8 const _nodeId, uint8 const _instance = 0x00);
>
> as the signature, and nothing broken. :)
>
> My bigger question though is how this interacts with our existing "auto 
> associate" groups for reports with the normal association class. 
>
> I dont "think" I have a multiinstance association device... and I see this 
> class is probably more aimed at setting up associations with other node 
> rather than gateways (as we are interested in all associations).
>
> So I'm wondering if this should instead be just a stand alone 
> CommandClass, or really integrated into our groups. 
>
> (take the rest of the converstation to this... email is easier for me at 
> the moment!)
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/OpenZWave/open-zwave/pull/560#issuecomment-102433664>.
>
From a users perspective, I think the api should not be verry different 
when adding a node to a group compared to a add  instance / node to a group.
The MultiInstanceAssociation Command class also reports the associations 
that where created with the Association Command class.
And the MultiInstanceAssociation Command class is also able to create node 
only associations. 
So I would think that when a device supports the MultiInstanceAssociation 
Command class we use that for all associations related manipulations.

The "auto associate" would just keep on working, but only setup with a 
different Command class. This is the route that I have taken thus far.
How would we expose a send alone Commandclass to the users?

openzwave-dev-watcher:
On Fri, 2015-05-15 at 08:56 -0700, xs4 wrote:
> 
> 
> Op vrijdag 15 mei 2015 17:33:55 UTC+2 schreef Justin Hammond:
>         ok. Guiding principle when adding/modifying stuff right now is
>         to maintain API compatibility. 
>         we could do 
>         
>         void AddAssociation( uint8 const _nodeId, uint8 const _instance = 0x00);
>         
>         as the signature, and nothing broken. :)
>         
>         My bigger question though is how this interacts with our
>         existing "auto associate" groups for reports with the normal
>         association class. 
>         
>         I dont "think" I have a multiinstance association device...
>         and I see this class is probably more aimed at setting up
>         associations with other node rather than gateways (as we are
>         interested in all associations).
>         
>         So I'm wondering if this should instead be just a stand alone
>         CommandClass, or really integrated into our groups. 
>         
>         (take the rest of the converstation to this... email is easier
>         for me at the moment!)
>         
>         —
>         Reply to this email directly or view it on GitHub.
>         
> From a users perspective, I think the api should not be verry
> different when adding a node to a group compared to a add  instance /
> node to a group.

At a high level, yep, I agree, as long as it functions almost identical
to the existing class. I'm very weary of breaking apps.... :)


> The MultiInstanceAssociation Command class also reports the
> associations that where created with the Association Command class.
> And the MultiInstanceAssociation Command class is also able to create
> node only associations. 

Ok.
so what happens when a device supports the Association Class and the
MultiInstance Association?

I guess the current code would add a 2 Associations? 1 via the Old
Association Class and 1 via the MultiInstance Association?

Looking at your code so far, the old Association Commands use the
original class, and your new methods trigger the MultiAssociation
commands right? 

I think we probably have to put some logic and and make it all uniform.
If the node Supports MultiAssociation CC, use that by default for all
Association Related commands and if not, fall back to the Original?


> So I would think that when a device supports the
> MultiInstanceAssociation Command class we use that for all
> associations related manipulations.
> 
Agreed. With possibly a config option to fall back if its broken on a
device :)

Ok, but is that on a TODO list as well? I don't see it that way at the
moment... 


> The "auto associate" would just keep on working, but only setup with a
> different Command class. This is the route that I have taken thus far.
> How would we expose a send alone Commandclass to the users?
>  

via the normal ValueID's. We would have to think best way to expose it
though... lets go there only if we need to. 


What devices support this? I might have to buy one as well *grin*

> 
> -- 
> You received this message because you are subscribed to the Google
> Groups "OpenZWave" group.
> To unsubscribe from this group and stop receiving emails from it, send
> an email to openzwave+unsubscribe@googlegroups.com.
> For more options, visit https://groups.google.com/d/optout.

openzwave-dev-watcher:
2015-05-15 18:22 GMT+02:00 Justin Hammond <justin@dynam.ac>:

> On Fri, 2015-05-15 at 08:56 -0700, xs4 wrote:
> >
> >
> > Op vrijdag 15 mei 2015 17:33:55 UTC+2 schreef Justin Hammond:
> >         ok. Guiding principle when adding/modifying stuff right now is
> >         to maintain API compatibility.
> >         we could do
> >
> >         void AddAssociation( uint8 const _nodeId, uint8 const _instance
> = 0x00);
> >
> >         as the signature, and nothing broken. :)
> >
> >         My bigger question though is how this interacts with our
> >         existing "auto associate" groups for reports with the normal
> >         association class.
> >
> >         I dont "think" I have a multiinstance association device...
> >         and I see this class is probably more aimed at setting up
> >         associations with other node rather than gateways (as we are
> >         interested in all associations).
> >
> >         So I'm wondering if this should instead be just a stand alone
> >         CommandClass, or really integrated into our groups.
> >
> >         (take the rest of the converstation to this... email is easier
> >         for me at the moment!)
> >
> >         —
> >         Reply to this email directly or view it on GitHub.
> >
> > From a users perspective, I think the api should not be verry
> > different when adding a node to a group compared to a add  instance /
> > node to a group.
>
> At a high level, yep, I agree, as long as it functions almost identical
> to the existing class. I'm very weary of breaking apps.... :)
>
> Well for creating the associations that will be no problem I think.I will
changes the signature as you proposed, ie :
  void AddAssociation( uint8 const _nodeId, uint8 const _instance = 0x00);

And what about the retrieval function?

uint32 GetAssociations( uint32 const _homeId, uint8 const _nodeId, uint8
const _groupIdx, uint8** o_associations );

would it be a problem to add a new one like this?

uint32 GetAssociations( uint32 const _homeId, uint8 const _nodeId, uint8
const _groupIdx, InstanceAssociation** o_associations );

(InstanceAssociation is defined in group.h)


>
> > The MultiInstanceAssociation Command class also reports the
> > associations that where created with the Association Command class.
> > And the MultiInstanceAssociation Command class is also able to create
> > node only associations.
>
> Ok.
> so what happens when a device supports the Association Class and the
> MultiInstance Association?
>
> I guess the current code would add a 2 Associations? 1 via the Old
> Association Class and 1 via the MultiInstance Association?
>
> No there is no need to use both. The  MultiInstance Cc can do what the
Association CC can and a little bit more, so no need to use them both.


Looking at your code so far, the old Association Commands use the
> original class, and your new methods trigger the MultiAssociation
> commands right?
>
> Yes correct. But if we have only one method that needs to change.


> I think we probably have to put some logic and and make it all uniform.
> If the node Supports MultiAssociation CC, use that by default for all
> Association Related commands and if not, fall back to the Original?
>
> Yes that would work.

>
> > So I would think that when a device supports the
> > MultiInstanceAssociation Command class we use that for all
> > associations related manipulations.
> >
> Agreed. With possibly a config option to fall back if its broken on a
> device :)
>
> Ok, but is that on a TODO list as well? I don't see it that way at the
> moment...
>
> I will put the "rework after review" on my TODO list. :-)
My priorities where first have some basic version that demonstrates the
basic principle, next have some discussions
and then take it from there.

> The "auto associate" would just keep on working, but only setup with a
> > different Command class. This is the route that I have taken thus far.
> > How would we expose a send alone Commandclass to the users?
> >
>
> via the normal ValueID's. We would have to think best way to expose it
> though... lets go there only if we need to.
>
>
> What devices support this? I might have to buy one as well *grin*
>
> The "FIBARO System FGS221 Double Relay Switch 2x1.5kW" and "FIBARO System
FGBS001 Universal Binary Sensor" both support this.

> --
> > You received this message because you are subscribed to the Google
> > Groups "OpenZWave" group.
> > To unsubscribe from this group and stop receiving emails from it, send
> > an email to openzwave+unsubscribe@googlegroups.com.
> > For more options, visit https://groups.google.com/d/optout.
>
>
> --
> You received this message because you are subscribed to the Google Groups
> "OpenZWave" group.
> To unsubscribe from this group and stop receiving emails from it, send an
> email to openzwave+unsubscribe@googlegroups.com.
> For more options, visit https://groups.google.com/d/optout.
>

xs4:
ok I have done some rework and testing. Changes are:
- methods use default parameter now
- added new manager method:
```
		uint32 GetAssociations( uint32 const _homeId, uint8 const _nodeId, uint8 const _groupIdx, InstanceAssociation** o_associations );
```
- restored COMMAND_CLASS_ASSOCIATION_COMMAND_CONFIGURATION support
- reworked the add and remove associations logic in Groups 

So my TODO list is empty now :smile: . What do you think about the InstanceAssociation in Group.h? Maybe its better to move that to Defs.h? In stead of including groups.h is manager/driver? 

I think I have covered all your remarks except maybe the introduction of the option to use the Assocation CC in case of a broke device. Do you think thats needed? If so I will add that.

Fishwaldo:
I'll review tomorrow. 

One favour, do you think you could update OZW control panel as well? (It gets hairy on JavaScript so I'll understand if you can't).

Fishwaldo:
(If you can, don't rush just yet, let me look over what you have done so far).

cpp/src/Group.cpp
```diff
@@ -200,18 +223,27 @@ bool Group::Contains
...
 		if( Node* node = driver->GetNodeUnsafe( m_nodeId ) )
 		{
-			if( Association* cc = static_cast<Association*>( node->GetCommandClass( Association::StaticGetCommandClassId() ) ) )
+			if( MultiInstanceAssociation* cc = static_cast<MultiInstanceAssociation*>( node->GetCommandClass( MultiInstanceAssociation::StaticGetCommandClassId() )))
 			{
-				cc->Set( m_groupIdx, _nodeId );
+				cc->Set( m_groupIdx, _nodeId, _instance );
 				cc->QueryGroup( m_groupIdx, 0 );
 			}
+			else
```
Fishwaldo: this could be a else if and the final else, I would log a message about no supported Association CC's found.

cpp/src/Group.cpp
```diff
@@ -222,17 +254,26 @@ void Group::AddAssociation
...
 		if( Node* node = driver->GetNodeUnsafe( m_nodeId ) )
 		{
-			if( Association* cc = static_cast<Association*>( node->GetCommandClass( Association::StaticGetCommandClassId() ) ) )
+			if( MultiInstanceAssociation* cc = static_cast<MultiInstanceAssociation*>( node->GetCommandClass( MultiInstanceAssociation::StaticGetCommandClassId() ) ) )
 			{
-				cc->Remove( m_groupIdx, _nodeId );
+				cc->Remove( m_groupIdx, _nodeId, _instance );
 				cc->QueryGroup( m_groupIdx, 0 );
+			} 
+			else
```
Fishwaldo: same as AddAssociation - else if, and final else being a Log Message.

Fishwaldo:
I've made some comments, have a look over it. 

For the fallback, I'd like to at least put the provision in for it, rather than later hitting a buggy device and having to do it then...
I'd create a "assicationccfallback" variable in the Protocol Section of ReadXML in the NodeClass. (Add it near the nodeinfosupported tests would be the right place to put it). Then we would have to expose this via a private Get* method to the Group/Driver classes etc, and in your "if CC exists, do, else, if newCC exists do" blocks, add a test for the fallback flag as well.

Finally, also remember to update WriteXML as well (and bonus points update the config/device_configuration.xsd file that does our syntax checking).

other than those comments and this above, I think everything else is fine. This is very much appriciated. 

Final request though is that once thats done, could you do a new PR against the New-CC branch rather than the Issue-352-Security branch?

cpp/src/Group.cpp
xs4: fixed

cpp/src/Group.cpp
xs4: fixed

xs4:
Thanks for reviewing. I added logging.

I though about adding a option to disable part of the MultiInstansAssociation CC but came to the conclusion that it makes no sense to have half of it's behavior. Ie only adding but no removing, or only querying. It all or nothing, otherwise you will end up with messy end result.

So would removing the CC completely like this: 
'''

<CommandClass id="142" action="remove" />
'''
not be sufficient?

I will open a new pull request based on the New-CC branch.

Fishwaldo:
the remove option - Ya... I didn't think about this approach...

PR #570, opened 2015-05-20
Z-Wave Plus Info Report Command Class

xs4:
initial implementation for the COMMAND_CLASS_ZWAVE_PLUS_INFO command class.
Tested on the Issue-352-Security branch with a Aeon Labs Recessed Door Sensor gen5.
The icons could also be a value list but I do not have sufficient information at this time. 

The config will look like this:
```
                        <CommandClass id="94" name="COMMAND_CLASS_ZWAVE_PLUS_INFO" version="1" request_flags="4" innif="true">
                                <Instance index="1" />
                                <Value type="byte" genre="system" instance="1" index="0" label="version" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="255" value="1" />
                                <Value type="list" genre="system" instance="1" index="1" label="role" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" vindex="6" size="5">
                                        <Item label="Central Controller" value="0" />
                                        <Item label="Sub Controller" value="1" />
                                        <Item label="Portable Controller" value="2" />
                                        <Item label="Portable Reporting Controller" value="3" />
                                        <Item label="Portable Slave" value="4" />
                                        <Item label="Always On Slave" value="5" />
                                        <Item label="Reporting Sleeping Slave" value="6" />
                                        <Item label="Listening Sleeping Slave" value="7" />
                                </Value>
                                <Value type="list" genre="system" instance="1" index="2" label="node" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" vindex="0" size="0">
                                        <Item label="Z-Wave+ node" value="0" />
                                        <Item label="Z-Wave+ ip router" value="1" />
                                        <Item label="Z-Wave+ ip gateway" value="2" />
                                        <Item label="Z-Wave+ ip client and ip node&#x09;" value="3" />
                                        <Item label="Z-Wave+ ip client and zwave node&#x09;" value="4" />
                                </Value>
                                <Value type="short" genre="system" instance="1" index="3" label="installerIcon" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="-32768" max="32767" value="3072" />
                                <Value type="short" genre="system" instance="1" index="4" label="userIcon" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="-32768" max="32767" value="3072" />
                        </CommandClass>

```

ps works best in combination with #506


Fishwaldo:
This is a good start. It gets more "hairy" though as this CC also defines a "Device Type" which is a more strict implementation of Device Classes. We can start with this, but eventually will have to tie it in with the AddMandatoryCommandClasses method in the Node Class, which means integrating with the QueryStage_NodeInfo code etc (but ZWavePlus will always be the first CC returned for devices that support it so the test should be reasonably simple)

I'd also consider exporting the node and role values back to the GetNodeType and/or add necessary Manager methods so it is similar to the old way of picking up Node Details. If we do it that way, I don't think we need to export anything via ValueID's (in fact I'd even create GetNodeIcon on the manager class) and treat the ZWavePlus CC similar to Manafacturer_Specific CC

I've also got a URL at home to either smartthings or another web api that lists all the Icon types... I'll dig it up tonight.

Fishwaldo:
here is the icon list:
http://220.135.186.178/zwave/example/ZWAVEPLUS%20INFO/Icon_Type/index.html

xs4:
Thanks for the list. I created a ValueList for the generic type. The specific type, which depends on the generic type is just a valueByte.

Basic question is how to continue. I get  that you eventually want to use the Zwave Plus device info in combination with the device_classes.xml handling, but lets take small steps. And maybe only start with the retrieval of the values. That brings me the the next question ValueClasses versus exposing via manager api calls? 

I could create method like:
```
    uint8 GetNodeZWPlusRole( uint32 const _homeId, uint8 const _nodeId );
    uint8 GetNodeZWPlusNode( uint32 const _homeId, uint8 const _nodeId );
    // usericon only?
    uint8 GetNodeIcon( uint32 const _homeId, uint8 const _nodeId );
```
Just the uint8 variant or also a String version? And remove the ValueClasses?

xs4:
Ok I started exposing the Zwave+ info via Manager. I started by adding these methods:
```
        bool IsNodePlusInfoReceived( uint32 const _homeId, uint8 const _nodeId );
        uint8 GetNodeIcon( uint32 const _homeId, uint8 const _nodeId );
        string GetNodeIconName( uint32 const _homeId, uint8 const _nodeId );
```
For now I have disabled the ValueClasses support.   
Next I will look at exposing the role and/or type info via GetNodeType.

xs4:
Created a new QueryStage NodePlusInfo that  will retrieve the ZWave+ info when available and will update the mandatory device classes based on what is configured in device_classes.xml.
I think I covered all your point.

Fishwaldo:
I will review all this soon I promise. We need to see how the new QueryStage is going to "integrate" with the updates in the Security-352 Branch, but other than that, it looks pretty good.

Fishwaldo:
I'm trying to merge in the master branch here... but a ton of merge conflicts. once merged, I'll get to this review. :+1:

xs4:
Ok, I will update this pull request when there are merge conflicts after you merged master.

Fishwaldo:
ok... Here come the initial comments:

Your IconType is actually the DeviceType as far as I understand it (Icon's map directly to DeviceTypes).

Lets make the DeviceTypes a XML database in the DeviceClasses file as well.. (like you did for roles). I don't think we need to "export" the DeviceType as the IconNameString. Just the Integer should be sufficient. 

as far as CommandClasses go - Mandatory CommandClasses are defined by DeviceType, and not NodeType or RoleType - So we need to update that part, but I'd still pass RoleType to the SetDeviceClasses as we might be able to use that to determine if we should add the WakeUP CC. 

lets also export the Role/DeviceType via the Manager class as well. 


Now, a bigger "change" we can think about:

If we get the zwave+ CC when we recieve a NIF, I *think* we can skip processing the MandatoryCommandClasses there, and just request for the zwave+ info, and use the DeviceType there instead of the Generic/Specific Types passed in the NIF. (Lets get this current PR updated and then I can work with you on that)

Cheers

xs4:
ok, I think I processed all your point. I updated the Mandatory CommandClasses processing, it is now base on the DeviceType, (and also on Role but since no command_classes are configured it just prints a log line but could also be removed)

Fishwaldo:
Ok. I think we still got the DeviceType stuff messed up a bit. Its not making sense to me based on what I've seen.... but I'll pull this in very soon and try to fix it up.

PR #588, opened 2015-06-04
Fixed sending color command, Disable other colors when sending value

gizmocuz:
Fixed sending color command (after ColorCmd_Set we should send the amount of color pairs)

Implemented, disabling other colors (When sending for example a RGB command, then we have to turn of the possible white colors)

### DIFF
--- a/cpp/src/command_classes/Color.cpp
+++ b/cpp/src/command_classes/Color.cpp
@@ -25,10 +25,6 @@
 //
 //-----------------------------------------------------------------------------
 
-#include <iostream>
-#include <iomanip>
-
-
 #include "command_classes/CommandClasses.h"
 #include "command_classes/Color.h"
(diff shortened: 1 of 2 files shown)


Fishwaldo:
1) - when doing PR - Please try to follow the coding style. There is so many style changes here its hard to see what you actually changed. 
2) - This seems more like a policy than a requirement. If a application sends 4 colors and then later 3 colors, then I think the application needs to be fixed.

gizmocuz:
Okey, must be plugin i am using in the IDE.

most important changes:

		msg->Append(3 + (nocols * 2)); //length
		msg->Append(GetCommandClassId());
		msg->Append(ColorCmd_Set); //cmd
--->>>>    msg->Append(nocols); //number of color pairs
		msg->Append(COLORIDX_RED);

PR #594, opened 2015-06-06
Update minimote.xml

nechry:
the parameter 250 is suppose to be read and write. the default value is right "Group". Like the key fob gen5 a list is more appropriate for this parameter

### DIFF
--- a/config/aeon_labs/minimote.xml
+++ b/config/aeon_labs/minimote.xml
@@ -44,8 +44,12 @@
       <Item label="Learn Mode" value="5" />
     </Value>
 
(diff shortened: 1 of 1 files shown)


ualex73:
Is it an idea to keep the following in the Help:

Button 1 (tap)=1, Button 1 (hold)=2, Button 2 (tap)=3, Button 2 (hold)=4, Button 3 (tap)=5, Button 3 (hold)=6, Button 4 (tap)=7, Button 4 (hold)=8.

nechry:
From my point of view, is something evident, but if you thinks is better to keep this part is totally right.

PR #595, opened 2015-06-07
update smart switch gen5

nechry:
remove parameter 1
adapte parameter 2, with more details help and default value
add parameter 3 Current Overload Protection
add parameter 20 Output Load Status
adjust some labels and default values
add parameter 200 Partner ID
add parameter 252 Configuration Locked
rename group 2 and remove auto attribute

### DIFF
--- a/config/aeon_labs/ses2.xml
+++ b/config/aeon_labs/ses2.xml
@@ -1,51 +1,52 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Product xmlns='http://code.google.com/p/open-zwave/'>
+  <!-- Aeon Labs Smart Switch Gen5 -->
+  <!--Source: http://ozw.my-ho.st/Aeon/Engineering%20Spec%20-%20Aeon%20Labs%20Smart%20Switch%20Gen5_V3.25.pdf-->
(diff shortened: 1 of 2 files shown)


Gert05:
You're skipping parameter 81. Is that intentional?

FYI
<Value type="list" index="81" genre="config" label="Configure the state of the LED" units="" min="0" max="1" size="1" value="0">
<Item label="The LED will follow the status (on/off) of its load. (Default)" value="0"/>
<Item label="When the state of the Switch changes, the LED will follow the status (on/off) of its load, but the LED will turn off after 5 seconds." value="1"/>
</Value>

ualex73:
@nechry: if you can answer Gert05 question ... The configuration looks good, so in principle i am happy to merge after this nswer ;-)

nechry:
hello @Gert05 I missed this value when redo ... so yes parameter 81 must be in this specific configuration file.

@ualex73 , you can merge and I patch again to add this configuration item for the next race or you can change my PR ?

I fought for many defaults , according to the source it has many different values ​​for many items, I tried to keep the Aeon-labs reference. So we can think of other changes could happen over time, the sulfide and as we move conntraitre this module

PR #602, opened 2015-06-17
AssociationGroupInfo Command Class

xs4:
Initial version for the AssociationGroupInfo(AGI) Command Class.
Retrieving the group names is working and is used to set the association group names.
The  AGI Command_List is also working but is only used to output the information in the log file. The AGI Info report needs more work. The users are requested, in the log file, to report  AGI Info messages so we can continue this work. 

(This pull request also merged in master)


cpp/src/Node.cpp
```diff
@@ -498,6 +501,42 @@ void Node::AdvanceQueries
...
+				if( Association* cc = static_cast<Association*>( GetCommandClass( Association::StaticGetCommandClassId() ) ) )
+				{
+					cc->RequestState( CommandClass::RequestFlag_Static, 1, Driver::MsgQueue_Query );
+					m_queryPending = true;
+					addQSC = true;
+				}
+				else
+				{
+					// if this device doesn't support Associations, move to retrieve Session information
+					m_queryStage = QueryStage_GroupNames;
```
Fishwaldo: Shouldn't we move onto QueryStage_Static instead?

cpp/src/Node.cpp
```diff
@@ -498,6 +501,42 @@ void Node::AdvanceQueries
 				}
 				break;
 			}
+			case QueryStage_Groups:
```
Fishwaldo: Wouldn't the QueryStage_Association get us this info when it completes?
(Groups == Associations right?)

cpp/src/command_classes/AssociationGroupInfo.cpp
```diff
@@ -0,0 +1,336 @@
...
+// Request current state from the device
+//-----------------------------------------------------------------------------
+bool AssociationGroupInfo::RequestState
+(
+	uint32 const _requestFlags,
+	uint8 const _instance,
+	Driver::MsgQueue const _queue
+)
+{
+	if( /*_requestFlags & RequestFlag_Static ||*/ _requestFlags & RequestFlag_Dynamic )
```
Fishwaldo: this info should be static right?

cpp/src/command_classes/AssociationGroupInfo.cpp
```diff
@@ -0,0 +1,336 @@
...
+			{
+				// update the name of the group
+				group->SetLabel( name );
+			}
+		}
+		return true;
+	} 
+	else if( AssociationGroupInfoCmd_Info_Report == _data[0] )
+	{
+		Log::Write( LogLevel_Info, GetNodeId(), "Received AssociationGroupInfoCmd_Info_Report command report from node %d", GetNodeId() );
```
Fishwaldo: see the link I sent you for some hints about this message type.

Fishwaldo:
I'm wondering if we can get away without doing additional QueryStages here? If we just set it up to trigger on RequestFlag_Static we don't loose any functionality right (only the group might change right?)

I'd also just try to be consistent (to avoid confusion) and make sure all log messages etc say Association Group rather than just group (group == Association group right?)

you branched this from master branch did you?

cpp/src/Node.cpp
xs4: Yes if the number of groups is not yet retrieved QueryStage_Association would also retrieve this number of groups. 
Groups != Associations. Groups hold associations. A group can hold a number of associations. 

QueryStage_Association retrieves the actual associations and updates the routes.  
Here we also add the controller as association based on config options(auto=true). This could be enhanced to also take into account the group profile that is retrieved by the AGI class. 
(the group profile is not yet retrieved)

Thats why I have separate states for the number of groups, group info and associations.
Putting it all in  QueryStage_Static would not enforce the needed sequence.

cpp/src/command_classes/AssociationGroupInfo.cpp
xs4: Yes, once  AssociationGroupInfoCmd_Info_Report is done, static is sufficient.

xs4:
I agree that it would be nice if all info could be retrieved in QueryStage_Static state. But If we want to used the AGI info to decide if we need to add the controller in a group based on ie profile or supported command (Battery), we need better control on the order in witch the CC retrieve there data.  For now we must have the number of groups before AGI can retrieve it's info.  

I will review the log messages for consistency.

Yes I did branched this from master, I hope thats not an issue.

cpp/src/command_classes/AssociationGroupInfo.cpp
xs4: found it. :smile:

xs4:
I moved the group name retrieval to QueryStage_Static, so now only one extra QueryState is needed, QueryStage_Group, to determine the number of groups.

The Group Info report is now also supported and used to extract the (association) group profile which is stored in the Group object.

Fishwaldo:
I probably need you to rebase this on the New-CC class. I think its just gonna end up as conflicts when we try to merge back with master.

Fishwaldo:
(my gitfu is not that great though... so if you think it wont be a issue let me know)

xs4:
I don't think it will be a problem. If you merge master into New-CC,  this pull request will show only my work. If you refer to pull from a rebased version, that will work as well. I rebased it at https://github.com/xs4/open-zwave/tree/New-CC-AGI-rebase. I will also open an new Pull Request.

Fishwaldo:
I've merged master. Both this and 613 have conflicts. I've not investigated further yet tho. If you have the time maybe you can look at it.

Fishwaldo:
I'm not so sure on the auto associate function right now. Based on current code, we auto associate group 1, so doing this on subsequent groups might lead to duplicate notifications. (Eg, hail from group 1, which triggers a refresh of al dynamic vales, then a report message from group 2, which should be covered by group 1 message.). 

Maybe to start with, just log a message. If we can be assured of some consistency on zwave+ devices, we can disable auto=true for group 1 on them, and use this todo our "auto" association.

xs4:
ok lets close this one and continue in #613.

Regarding auto associate, lets take it one step at a time. So will change it so that is will only log a message. Later we can investigate if we can introduce auto associate for (only)zwave+ devices.

PR #604, opened 2015-06-17
SceneActuatorConf and SceneControllerConf Command Class

xs4:
Initial verions for SceneActuatorConf and SceneControllerConf Command Classes.

- A scene can now be started via a Manager call (via a broadcast message) 
- SceneActuatiors can be bound to a scene via a Manager call.
- Buttons on a SceneController can be configured to trigger a scene via Value classes.

This is all independent from the OZW scene handling, but could be integrated if desired.


cpp/src/Manager.h
```diff
@@ -2279,6 +2279,10 @@ OPENZWAVE_EXPORT_WARNINGS_ON
 		 */
 		bool ActivateScene( uint8 const _sceneId );
 
+		// register scene at actuator
+		bool SetActuatorScene( uint8 const _sceneId, ValueID const& _valueId, uint8 const _value, uint8 const _duration );
```
Fishwaldo: doco please!

cpp/src/command_classes/SceneActuatorConf.cpp
```diff
@@ -0,0 +1,164 @@
...
+	// once a actuator has activated the scene it can only be undone by a reset
+	// TODO _bReplyRequired is false else we get a timeout. Why is this needed?
+	Msg* msg = new Msg( "SceneActuatorConfCmd_Set", GetNodeId(), REQUEST, FUNC_ID_ZW_SEND_DATA, true, false, FUNC_ID_APPLICATION_COMMAND_HANDLER, GetCommandClassId() );
+	msg->Append( GetNodeId() );
+	msg->Append( 6 );
+	msg->Append( GetCommandClassId() );
+	msg->Append( SceneActuatorConfCmd_Set );
+	msg->Append( _sceneId );
+	msg->Append( _duration );
+	msg->Append( 0x80 ); // use our value not the current nodes level
```
Fishwaldo: why?

Fishwaldo:
is there anything (apart from doco) that lists how many scenes a device supports? or do they all support 255 scenes?
Should we consider a device config option to specify the number of scene's a device supports and throw a error if a scene is attempted to be set on a index thats out of range?

cpp/src/Manager.h
xs4: If you agree that this is the way to go, I will add documentation.

cpp/src/command_classes/SceneActuatorConf.cpp
xs4: I was hoping you could shed some light on this. What I see when bReplyRequired is true, the request times out, but the get that follows afterwards returns the new value. Maybe this command could also broadcasted ??!??

xs4:
I think that a device that supports scenes must always support 255 scenes. So I don't think we need a config option.

cpp/src/command_classes/SceneActuatorConf.cpp
Fishwaldo: Yeah - But broadcast would mean that all nodes that support this CC would save this scene. I'm not sure if thats the right option. 

Or - Like some classes, Set doesn't require a reply. In that case, you can drop all the rest of the variables after the false.

cpp/src/command_classes/SceneActivation.cpp
```diff
@@ -76,3 +77,30 @@ bool SceneActivation::HandleMsg
...
+	Driver * driver,
+	uint8 const _scene,
+	uint8 const _duration
+)
+{
+	Log::Write( LogLevel_Info, "Activating scene %d, duration %d", _scene, _duration );
+
+	if ( driver != NULL )
+	{
+		Msg* msg = new Msg( "SceneActivationCmd_Set", 0xFF, REQUEST, FUNC_ID_ZW_SEND_DATA, true , true);
```
Fishwaldo: replyrequired is probably false here if we are broadcasting.

Fishwaldo:
I'm still thinking about the best way to expose this API without confusing the existing Scene classes. :)

xs4:
Any thoughts on how to continue this?  Do you prefer separate pull requests for the SceneActuatorConf and SceneControllerConf Command Classes? Or move the current scene handling to a separate support library first?

Fishwaldo:
its on the "roadmap" so to say. The implementation of the CC's looks ok, its just how to co-exist with the existing Scene Class we have now without it getting confusing (and in all honesty, I've thought about dropping that Scene class when we go for a API breaking release (say version 2.0), so its probably not worthwhile integrating it)

Fishwaldo:
Looking into this more, we appear to be using the number of groups to query the number of scenes. But as I understand it, there is no correlation between the two. Any device must support upto 255 scenes... so we need to reload at the initilzation there where we do a node->GetNumGroups() call, as that sounds wrong....

justfalter:
FWIW, I believe this would be a valuable addition to the project. Without support for SCENE_ACTUATOR_CONF and SCENE_CONTROLLER_CONF, I had spend $140 on a Cooper RFHDCSG to set up the association between a Cooper RFWC5 and other devices.

Fishwaldo:
Hi,

Thanks

PR #607, opened 2015-06-20
Add Aeotec DSC24 and Everspring AD147

sleveque:
(first PR ever, sorry in advance if I messed up something :) )

Adding support for two devices I own, Aeotec DSC24 and Everspring AD147.
AD147 gets a special config file to switch between dimming and on/off mode.

Please review and comment


sarakha63:
The config is wrong for the everspring AD147/ The second association group has 4 max associations allowed. And this second group shouldn't be set to auto true

Fishwaldo:
please open a new issue, otherwise this will get lost.

PR #614, opened 2015-06-27
Update SensorBinary CommandClass

xs4:
* added a list of sensor names
* updated logging when receiving a SensorBinary message with version > 1 info
* when the configuration provides a sensor map use that, otherwise use sensor auto discovery (to maintain backward compatibility)


xs4:
Any thoughts about this? Maybe remove the SensorMap support? and  have only auto discovery? or use the auto discovery to generate/update the SensorMap? Or the current version?

At the moment we have SensorBinary V2 devices without proper configuration (and are not working correctly now because of it), we should have a solution for those. My preference would be that they work without the need for a configuration file, so use some sort of auto discovery.

Fishwaldo:
Sorry, real life has been a pain these last three weeks. I will get to it shortly.

At the moment, V2 should still work with our V1 class using the first two values in the Report message. The issue here is that we wouldn't know what those represent as it vendor dependant. V2 added as you know more types of defined reports.

What I want to think about is how to incorporate V2 "defined" messages while retaining V1 compatibility from a existing API perspective. It might not be possible, but I've not given it a lot of thought. 

(I want to nail the AGI merge I started soon, then get to this... the Scene work you did, I'm still thinking about the right approach as far as API goes there)

xs4:
I think you are talking about the alarm cc v2? At the moment the SensorMap handling is already V2 info we are processing.

Fishwaldo:
No, I'm talking about SensorBinary. Even with the sensormap as it is now, its arbitrary values that are sent over with. We still expose that arbitary values in with V2 devices (as its backwards compatible) but with V2, we now have Defined Sensors right? Technically, the V2 data should replace the V1 ValueID's (to remain backwards compatible, otherwise applications have to start guessing at Version info, which I think should be abstracted away).

eg, if your App supports SensorBinary right now, you expose the one static existing ValueID's to your users, and then random ValueID's (based on SensorMaps). If we update to V2, we should continue to use those ValueID's where possible, rather than have both a V1 ValueID(s) and a V2 ValueID(s)

(i'm ok adding new ValueID's to classes when its introducing new functionality, but not when version changes are duplicating "basically" the same info)

One thing we could (and should do) is expose the SensorMap as a ReadOnly ValueList as well, so we can determine what type SensorBinary report we are dealing with (as i understand it, a single SensorBinary CC instance on a device can expose "multiple" sensors - eg, motion and tamper for example)

so my question is - how to do this without breaking applications...

Fishwaldo:
Oh, on this point:
when the configuration provides a sensor map use that, otherwise use sensor auto discovery (to maintain backward compatibility) 

I think reverse that logic - When we get a V2 SupportedReport message, then drop any config file sensormap. AutoDiscovery is the way to go! (but of course, reload it when the zwcfg_*.xml file is present)

xs4:
Well the current implementation has a bug that lets you think there is a random value, the code has an off by one issue and is evaluating the last crc byte.

xs4:
hmm, I see were the off by one is coming from; When the sensor binary message is send via a MultiCmd the length is passed as length-1; while when received  directly the length is passed through as is. What do you think should I change MultiCmd to passing the length as is? I can't oversee the impact of that, but if we don't we can never trust the length.....

xs4:
Ok, I reworked it so that ico V2 we always use AutoDiscovery. If we find a SensorMap ico V2 we warn and remove it. 
I updated  MultiCmd so that the length in the message is passed along, just like in Node::ApplicationCommandHandler. This way we have consistent behavior.  

To recap. V1 has only one value True or False. When the SensorBinary message received directly we passed the length as reported in the message causing the code to report a random value(being the crc code). When the SensorBinary message was received in MultiCmd the length changed and the random value was not reported.

Fishwaldo:
Hi,

Thanks

PR #641, opened 2015-08-02
config ZMNHBA2: correct instance/endpoint mapping

xs4:
without the mapping the device reports tree in stead of two switches due to a incorrect endpoint mapping.

### DIFF
--- a/config/qubino/ZMNHBA2.xml
+++ b/config/qubino/ZMNHBA2.xml
@@ -68,5 +68,8 @@
 
 	<!-- Remove COMMAND_CLASS_BASIC -->
 	<CommandClass id="32" action="remove" />
+	
+	<!-- Map endpoints to instances -->
+	<CommandClass id="96" mapping="endpoints" />
 
 </Product>


nechry:
Nice catch! Thank

Egregius:
The endpoint mapping is still not good. 
With this change it works good for I1 but not for I2 (still reporting as Q1 and Q3 active)

nechry:
Yes, I also got new wrong behavior with this patch. the behavior is different from button vs zwave command.

Egregius:
I learned that a refresh of the device in ozwcp get's the correct status. 
I'm trying to fix this externally by scripts but can't seem to get it stable.
Any ideas?

Fishwaldo:
Can you send logs to the mailing list? Please remove any zwcfg_*.xml and restart OZWCP before doing so

Egregius:
Can you tell me in short wich steps to take?

nechry:
Hello, I want go forward with this issues. you need one log with Map endpoints to instances activated and one without? 
I got the same issues with the ZMNHBDx (zwave+ revision of the ZMNHBAx)
This switch have 3 instances on Master, S1 and S2. With Map endpoints to instances activated, we see only 2 instances, the reels one. But the first one receive the S2 state. and yes with a manual refresh we restore the S1 value as expected.

Warp95:
Hi,
I got the same issues with the Qubino ZMNHBD2 Flush 2 Relays module !!
Help us, please....

Issue #446, opened 2015-03-14
Import of std namespace in headers leads to VS2015 (and earlier?) build errors


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Fetch latest source (I'm using openzwave-1.3.1001)
2. Build using VS 2015 (and perhaps earlier versions)

What is the expected output? What do you see instead?
I'd hope for no compilation errors, but instead I get: 
ZWManager.cpp(242): error C2976: 'std::array': too few template arguments
...

What version of the product are you using? On what operating system?
openzwave-1.3.1001 on Windows 10, using VS 2015 (CTP)

My thinking is that this is caused because we are including the std namespace 
in headers rather than in the cpp files (via the using namespace std in 
Defs.h).  This is causing a conflict with the Visual Studio "array" interop 
class which predates std.

My short-term fix is to replace the "using namespace std" in Defs.h with more 
specific usings:
#include <string>
#include <map>
#include <list>
#include<vector>
#include <strstream>
#include <set>

// Declare the OpenZWave namespace
namespace std {}
namespace OpenZWave
{
	// Include the STL namespace
	//using namespace std;
	using std::string;
	using std::map;
	using std::list;
	using std::vector;
	using std::stringstream;
	using std::set;
}

This lets everything build.  A more complete fix might be to remove the "using 
std" from the headers.  Its been some time since I did any serious C++, so I 
may well be talking complete nonsense, or there may be an easier way to do 
things.  In any event if someone wants to build using a more recent version of 
Visual Studio this does work.
```

Original issue reported on code.google.com by `dam...@mehers.com` on 26 Jan 2015 at 2:09

JKlessens:
I can confirm that this is still a problem!

Fishwaldo:
I commited a less intrusive fix where we qualified the namespace of the array we would like to use. This should be fixed.

PR #643, opened 2015-08-06
Adding WinRT implementation of OpenZWave stack

jplafonta:
Adding a winRT implementation of the platform dependent classes.

Adding a VS 2015 .sln/.vcxproj to build a lib file that can be consumed by WinRT applications (specifically UWP apps).  Note that UWP apps using this library will need to include the SerialCommunication capability (required by the WinRT serial APIs).

This library is currently used by the Windows ZWave Device system bridge.  More information about that can be found here: http://ms-iot.github.io/content/en-US/win10/samples/ZWaveTutorial.htm


Fishwaldo:
1) - You need to add a copyright for yourself into the new files (or do you want to "assign" copyright to OZW?
2) - Any possibility to merge the Visual Studio Solution Files into one? right now, every-time we add new files, we have to modify the Unix MakeFiles and VS2013 files (I've ignore the older VS solution files, and should probably delete them). Adding a new set of WinRT files would make it even more tedious. :)
3) What is the "NetworkMonitorInterval" config option? 
4) what is the "ThreadTerminateTimeout" config option?
5) Putting in a ControllerPath into the stock standard Options.xml file isn't a great Idea. Please drop it.
6) can you explain the m_initMutex in the Driver class and why its needed?
7) Why do you include "platform/winRT/HidControllerWinRT.h" in Driver.cpp instead of using the implementation class platform/HidController.h
8) Could you rebase against the Dev branch instead of Master.

Other than that, thanks very much for working on this.

Fishwaldo:
Any update on my questions?

jplafonta:
1) I will need to get back to you about this.  Assuming we can get legal approval (and approval from your end) I will add copyright Microsoft.
2) Would prefer to keep this separated, unless you are ok with getting rid of the VS2013 .sln and .vcxproj files and just using VS2015 versions.  In that case I would have different configurations for the winRT, and windows libs.  This may be a little tricky too as the platform depending *Impl files would need to be conditionally compiled depending on the configuration.
3) Will removed it. This is specific to our application and doesn't belong in OZW.
4) In WinRT, it is not possible to terminate a thread. Thus ThreadImpl::Terminate will wait a configurable amount of time for the thread to finish on its own, and then return whether it actually does or not.
5) Will remove this.
6) The following ordering of events was happening causing an access violoation (accessing freed object)
a) Our application called Manager::Create() -> causes driver to be created and initialized
     b) Driver starts "DriverThread".
     c) Application calls attempts and fails to initialize another component (alljoyn in this case), which fails. As a result the application bails and calls Manager::Destroy, which then causes the driver to be destroyed -> driver destructor calls thread::Stop on the polling thread (it has not been started), deletes that thread

     d) DriverThread resumes execution, calls thread::Start on polling thread which has now been deleted.

7) Unlike the other platform specific classes, there is HidController.*, but no corresponding HidControllerImpl.*. Instead there are a couple of platform specific implementations of hidapi.h which HidController.cpp relies on. However the way you interact with a hid device in WinRT is significantly different (more controlled by the platform giving you less freedom to do arbitrary reads/writes), making implementing hidapi.h difficult and even impossible in some cases. Hence the WinRT specific HidControllerWinRT.h.
8) Yes, I will do this

jplafonta:
I think I have addressed all the feedback at this point.  Please let me know if there is anything else you would like me to change.

Fishwaldo:
yep. I will merge soon.

Fishwaldo:
Question: How did you overcome issue #446 error with your code? Or doesn't WinRT suffer the same issues?

Fishwaldo:
(can you also please drop me a email at justin@dynam.ac)

jano42:
@Fishwaldo do you plan to integrate this pull request into master branch ?
I've seen that's the last remaining thing for you to publish (very expected) 1.4 version.

Fishwaldo:
yes I do. I'm just trying to get a CI environment running for it

Fishwaldo:
@jplafonta: what do I need installed/configured to compile this? I have Visual Studio 2015 installed, with the Win10 SDK etc, yet I'm getting a error when trying to compile:
```
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets
(2048,5): error MSB3774: Could not find SDK "Microsoft.VCLibs, Version=14.0". [
C:\Users\fish\Documents\GitHubVisualStudio\open-zwave\cpp\build\winRT\vs2015\Op
enZWave.vcxproj]
```
For the life of me, I can't get this to work. Reinstalled VS2015 many times already.

Fishwaldo:
I think I got this fixed. Thanks! Sorry it took so long, took a while to get my Windows Skills upto par!

PR #645, opened 2015-08-10
config for Fibaro FGR221 (firmware 109)

stephanlascar:
It's just a new specification for the popular Fibaro FGR221 using firmware 109.


Fishwaldo:
Many Thanks

PR #655, opened 2015-08-31

ewgast:
updated benext product types
added benext heating controller
added panicwatch+button


Fishwaldo:
thanks!

PR #665, opened 2015-10-06
Added Qubino ZMNHBD2 Flush 2 Relays module

emdioh:
It's similar, to the ZMNHBA2, but id and internal parameters are different


Fishwaldo:
Thanks!

nechry:
I got same wrong behavior with this config https://github.com/OpenZWave/open-zwave/pull/641

PR #668, opened 2015-10-10
PR: add some useful variables in pkg-config

ekarak:
Precise git version is needed to be able to scan and enable the new security API in openzwave-shared. I've also added the shared conf dir (PREFIX/etc/openzwave) and documentation (PREFIX/share/doc/openzwave-<version>)


ekarak:
Please refer to [the gyp binding](https://github.com/OpenZWave/node-openzwave-shared/blob/security/binding.gyp) that uses these variables. I also need to be able to query pkg-config so as to pass the device DB XML path (usually /usr/local/etc/openzwave) to my library (in detail, preprocessor macro OPENZWAVE_ETC)

PR #677, opened 2015-10-21
Added configuration files for Linear dimmers

vexofp:
Two configuration parameters, and an override for the Level's max value. Both the module and the in-wall switch work the same.


Fishwaldo:
thanks!

PR #678, opened 2015-10-21
Fix includes to not assume the base include directory

vexofp:
Currently, the headers assume that you are adding the openzwave include directory directly to the compiler's search path (e.g. -I /pathto/include/openzwave & #include "Manager.h").

When this is not the case, includes in the headers like "#include platform/Stream.h" from platform/Controller.h will break, because platform is not in the search path, nor in the directory of the current file.

Making the includes relative to the current file's directory fixes this and gives the flexibility to set up your includes differently (e.g., -I /pathto/include & #include "openzwave/Manager.h")

### DIFF
--- a/cpp/src/platform/Controller.cpp
+++ b/cpp/src/platform/Controller.cpp
@@ -24,9 +24,9 @@
 //	along with OpenZWave.  If not, see <http://www.gnu.org/licenses/>.
 //
 //-----------------------------------------------------------------------------
-#include "Defs.h"
+#include "../Defs.h"
 #include "Driver.h"
-#include "platform/Controller.h"
+#include "Controller.h"
 
 using namespace OpenZWave;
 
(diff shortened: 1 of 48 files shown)


Fishwaldo:
You should have discussed this on the list before going ahead and doing all the work. 

I currently see no issue with using the correct Include Path with building, and I'm unlikely to accept this as it has the potential to break builds for existing applications that incorporate OZW already.

In fact, if you use pkg-config or the ozw_confg --CFlags in your build scripts (Makefiles or whatnot) then the include paths should be setup correctly already. 

Its a lot of churn for very little return. Sorry :(

vexofp:
No worries, was just sending up changes I'd already made locally for my own build environment. Just a few quick sed commands. ;)

I would still say that the fact that the simplest possible case
```
cd open-zwave
make && sudo make install
g++ -o test -x c++ - -lopenzwave <<<"#include <openzwave/Manager.h> ..."
```
won't compile (when it easily could) without dragging pkg-config into the mix, violates the behavior expected of most libraries.

Issue #687, opened 2015-11-03
pkg-config stuff is missing for libopenzwave1.3-dev package


pastukhov:
Seems like libopenzwave.pc generation is wrong and missing in libopenzwave1.3-dev package creation spec.

Fishwaldo:
Wrong? In what way. Can you provide a bit more detail than a one liner please?

pastukhov:
I'm trying to build packages for Debian jessie directly on my Banana Pi:
```dpkg-buildpackage -b```
* libopenzwave.pc file is missing from libopenzwave1.3-dev_0.9_armhf.deb.
I found it in ./debian/tmp/usr/local/lib/arm-linux-gnueabihf/pkgconfig/libopenzwave.pc:
```
prefix=/usr
exec_prefix=/usr/bin
libdir=/usr/lib/arm-linux-gnueabihf/
includedir=/usr/include/openzwave/
sysconfdir=/usr/etc/openzwave/
gitversion=0.9.-1
docdir=/usr/share/doc/openzwave-0.9.0/

Name: libopenzwave
Description: A Open Source implementation of the ZWave Serial API
URL: http://www.openzwave.org/
Version: 0.9
Libs: -L${libdir} -lopenzwave
Cflags: -I${includedir} 
```
* So it has been generated in wrong place and having wrong version number.

Fishwaldo:
Could you please post the output of "pkg-config --variable pc_path pkg-config"

Fishwaldo:
Pending on a fix from PR #689

jjhuff:
Curious if anyone has a solution for this. I'm using the Debian stretch package, and it's missing libopenzwave.pc (https://packages.debian.org/stretch/amd64/libopenzwave1.5-dev/filelist)

Thanks!

PR #689, opened 2015-11-07
Update Debian packaging

lnussbaum:
This set of commits reworks the Debian packaging of open-zwave, bringing it mostly up-to-par with what's expected for integration into Debian.

The two changes outside of debian/ are the first two commits.

i've added a debian/TODO file listing the remaining work. Some of it is on the debian packaging side, some of it is on the upstream side. This work is only strictly required if we want to push open-zwave to Debian. But some of it might be worth it as general cleanup moves.

Lucas


Fishwaldo:
Thanks. If you could do some sanity checks on the deb snapshots at http://old.openzwave.com/snapshots/ (when it finishes building them in about a hour or two) that would be great.

Fishwaldo:
another question. what about pkg-config .pc files:
See issue #687

lnussbaum:
ah, yes, pkg-config files should be added. also, there should probably be a dependency in openzwave to libopenzwave*. I'll fix those two in the next few days.

PR #690, opened 2015-11-07
Fixed crash in CommandClass::ReadValueRefreshXML and corrected 3 type castings

gizmocuz:
Fixed crash in CommandClass::ReadValueRefreshXML that was caused by QueryIntAttribute which takes an int pointer as argument and a uint8 pointer was provided

Corrected 3 type castings that where causing a compile warning

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -3871,7 +3871,7 @@ void Driver::CommonAddNodeStatusRequestHandler
 			state = ControllerState_Completed;
 			if( m_currentControllerCommand != NULL && m_currentControllerCommand->m_controllerCommandNode != 0xff )
 			{
-				InitNode( m_currentControllerCommand->m_controllerCommandNode, true, m_currentControllerCommand->m_controllerCommandArg, m_currentControllerCommand->m_controllerDeviceProtocolInfo, m_currentControllerCommand->m_controllerDeviceProtocolInfoLength );
+				InitNode( m_currentControllerCommand->m_controllerCommandNode, true, m_currentControllerCommand->m_controllerCommandArg != 0, m_currentControllerCommand->m_controllerDeviceProtocolInfo, m_currentControllerCommand->m_controllerDeviceProtocolInfoLength );
 			}
 
 			// Not sure about the new controller function here.
--- a/cpp/src/Manager.cpp
+++ b/cpp/src/Manager.cpp
@@ -998,7 +998,7 @@ bool Manager::IsNodeZWavePlus
 		version = driver->IsNodeZWavePlus( _nodeId );
 	}
 
-	return version;
+	return (version != 0);
 }
 
 
@@ -1290,7 +1290,7 @@ uint8 Manager::GetNodeDeviceType
 {
 	if( Driver* driver = GetDriver( _homeId ) )
 	{
-		return driver->GetNodeDeviceType( _nodeId );
+		return (uint8)driver->GetNodeDeviceType( _nodeId );
 	}
 
 	return 0x00; // unknown
--- a/cpp/src/Node.cpp
+++ b/cpp/src/Node.cpp
@@ -3207,7 +3207,7 @@ void Node::ReadDeviceClasses
 			if( keyStr )
 			{
 				char* pStop;
-				uint16 key = (uint16)strtol( keyStr, &pStop, 16 );
+				uint8 key = (uint8)strtol( keyStr, &pStop, 16 );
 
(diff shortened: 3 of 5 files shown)


cpp/src/Driver.cpp
```diff
@@ -3871,7 +3871,7 @@ void Driver::CommonAddNodeStatusRequestHandler
 			state = ControllerState_Completed;
 			if( m_currentControllerCommand != NULL && m_currentControllerCommand->m_controllerCommandNode != 0xff )
 			{
-				InitNode( m_currentControllerCommand->m_controllerCommandNode, true, m_currentControllerCommand->m_controllerCommandArg, m_currentControllerCommand->m_controllerDeviceProtocolInfo, m_currentControllerCommand->m_controllerDeviceProtocolInfoLength );
+				InitNode( m_currentControllerCommand->m_controllerCommandNode, true, m_currentControllerCommand->m_controllerCommandArg != 0, m_currentControllerCommand->m_controllerDeviceProtocolInfo, m_currentControllerCommand->m_controllerDeviceProtocolInfoLength );
```
Fishwaldo: This is ok

cpp/src/Manager.cpp
```diff
@@ -998,7 +998,7 @@ bool Manager::IsNodeZWavePlus
 		version = driver->IsNodeZWavePlus( _nodeId );
 	}
 
-	return version;
+	return (version != 0);
```
Fishwaldo: Its easier to change version to a bool rather than do this.  :)

cpp/src/Manager.cpp
```diff
@@ -1290,7 +1290,7 @@ uint8 Manager::GetNodeDeviceType
 {
 	if( Driver* driver = GetDriver( _homeId ) )
 	{
-		return driver->GetNodeDeviceType( _nodeId );
+		return (uint8)driver->GetNodeDeviceType( _nodeId );
```
Fishwaldo: DeviceType can be larger than a uint8.... so change the Prototype to a uint16

cpp/src/Node.cpp
```diff
@@ -3207,7 +3207,7 @@ void Node::ReadDeviceClasses
 			if( keyStr )
 			{
 				char* pStop;
-				uint16 key = (uint16)strtol( keyStr, &pStop, 16 );
+				uint8 key = (uint8)strtol( keyStr, &pStop, 16 );
```
Fishwaldo: The Key can be as large as a uint16, so this is wrong.

cpp/src/command_classes/CommandClass.cpp
```diff
@@ -329,12 +329,15 @@ void CommandClass::ReadValueRefreshXML
 	char const* str;
 	bool ok = false;
 	const char *genre;
+	int32 intVal;
```
Fishwaldo: why did you do this? I don't see any warnings on VS2015 and smells like a broken compiler rather than a problem.

cpp/src/value_classes/ValueList.cpp
```diff
@@ -355,12 +355,12 @@ bool ValueList::GetItemLabels
 // Get the Item at the Currently selected Index
 //-----------------------------------------------------------------------------
 ValueList::Item const *ValueList::GetItem() const {
-	try {
-		return &m_items.at(m_valueIdx);
-	} catch (const std::out_of_range& oor) {
+	if ((m_valueIdx < 0) || (m_valueIdx >= (int32)m_items.size()))
```
Fishwaldo: This just seems redundant for no good reason? Both achieve the same thing...

cpp/src/Manager.cpp
gizmocuz: this function expects a uint8 as return, so then this function header should be changed

Fishwaldo:
I merged a few of these fixes, but some don't seem correct.

cpp/src/command_classes/CommandClass.cpp
gizmocuz: QueryIntAttribute expects an int32 type, as you do in the rest of the code (same method)
You parse a uint8 pointer, and this caused a crash

cpp/src/value_classes/ValueList.cpp
gizmocuz: I had situations where m_ValueIdx was not assigned, and it had a value of -82385624
This case it checks also this

gizmocuz:
Thanks, please see a few reactions that i replied

PR #693, opened 2015-11-13
constify Notification::GetAsString() 

ekarak:
`const` needs to be added to the function signature (it returns a one-use message string anyway: no mutations whatsoever), otherwise I cannot call it from the callback handler, as I'm getting a [passing const discards qualifiers](http://stackoverflow.com/questions/5973427/error-passing-xxx-as-this-argument-of-xxx-discards-qualifiers?rq=1)

this is because the callback handler signature defines `_pNotification` as `const`
```
 typedef void(* 	pfnOnNotification_t) (Notification const *_pNotification, void *_context)
```

### DIFF
--- a/cpp/src/Notification.cpp
+++ b/cpp/src/Notification.cpp
@@ -37,7 +37,7 @@ using namespace OpenZWave;
 // Return a string representation of OZW
 //-----------------------------------------------------------------------------
 
-string Notification::GetAsString() {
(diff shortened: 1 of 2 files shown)


Fishwaldo:
Looks fine. Will merge tomorrow

Fishwaldo:
Done. Thanks

PR #700, opened 2015-11-24
Updated Linear switch configs and added Enerwave ZWN-SC7 config.

vexofp:


Fishwaldo:
Thanks!

MYeager1967:
Is the Enerwave ZWN-SC7 scene controller compatible with OZW? It pairs, but it doesn't seem to be functional past that point. I know with Vera, you assign scenes to each button but I can't find a way to do that with OZW...

PR #701, opened 2015-11-25
FreeBSD fixes

stromnet:
Some minor fixes to make build work on FreeBSD again.

Not sure on backwards compatibility on the iconv thing, but other FreeBSD ports seems to have been fully changed (i.e no conditionals): https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=199099, so I'd say openzwave should do the same?


Fishwaldo:
Thanks!

PR #712, opened 2015-12-12
add manager interface and value list method to retrieve list of integer values from value list

out4b:
This may be useful to upper SW stack or bindings to retrieve the list of integer values from list type values. This could make it more convenient for subsequent ozw.setValue() call, instead of getting string labels of the value list.


Fishwaldo:
Thanks!

PR #721, opened 2015-12-29
Create POP009303.xml

armaesiea:
Add POPP Battery Wall Switch Z-Wave+


armaesiea:
Add POPP Battery Wall Switch Z-Wave+

PR #723, opened 2016-01-01
Add brk smoke C02 Detector

joshughes:
This is my first try of adding devices to openzwave, so I am sorry if I did anything incorrectly. I will be more than happy to fix it up. That being said in my `zwcfg_*.xml` file I am seeing the following... 

```xml
<Node id="7" name="" location="" basic="4" generic="161" specific="0" type="Alarm Sensor" listening="false" frequentListening="false" beaming="true" routing="true" max_baud_rate="40000" version="4" query_stage="Probe">
                <Manufacturer id="" name="">
                        <Product type="" id="" name="" />
                </Manufacturer>
                <CommandClasses>
                        <CommandClass id="32" name="COMMAND_CLASS_BASIC" version="1" after_mark="true" mapping="113">
                                <Instance index="1" />
                        </CommandClass>
                        <CommandClass id="132" name="COMMAND_CLASS_WAKE_UP" version="1" request_flags="2">
                                <Instance index="1" />
                                <Value type="int" genre="system" instance="1" index="0" label="Wake-up Interval" units="Seconds" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="-2147483648" max="2147483647" value="3600" />
                        </CommandClass>
                </CommandClasses>
        </Node>
```

So I followed the instructions and found the device at http://www.pepper1.net/zwavedb/device/458 and created this PR. The device is still not autodetecting. I am not sure if its because its asleep or if I made a mistake adding the device. Any guidance is greatly appreciated.

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -147,6 +147,9 @@
 	</Manufacturer>
 	<Manufacturer id="0023" name="Boca Devices">
 	</Manufacturer>
+	<Manufacturer id="0138" name="BRK">
+		<Product type="0001" id="0002" name="Smoke and Carbon and Monoxide Alarm"/>
+	</Manufacturer>
 	<Manufacturer id="002d" name="Broadband Energy">
 	</Manufacturer>
 	<Manufacturer id="0123" name="Iwatsu Electric Co">


joshughes:
So looking at another issue I see that I need to figure out how to wake up the smoke detector so openzwave can finish the interview process...

I will work on that.

Fishwaldo:
Please read this: https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices

joshughes:
Thanks I did read that. I think the misunderstanding is that you only have al limited time with devices that sleep. In my case the smoke detector goes to sleep after 30 seconds in inclusion. Where I would unplug my zwave stick for switch inclusion I need to use openzwave to include it into my network so it can do the interrogation after inclusion. 

In any case I believe the xml i came up with is correct... unless there is something your trying to point out to me specifically from that doc.

Fishwaldo:
You should only need to include the Configuration CC variables, and association information. Copying for zwcfg_*.xml isn't necessary as 99% of it is automatically discovered.

Fishwaldo:
and if the device is asleep when you restart your application, there is usually a way to wake up a device (by hitting a button 3 times etc).

joshughes:
Ya reading online these smoke detectors are hard to include... seems like they don't wake up correctly but if you get them included correctly to begin with they work. I will do some experimentation with some exclusion and inclusion today. Is there anything incorrect with this PR? I think other people would like to have this device mapped with open-zwave.

Fishwaldo:
Have a look at the other config files. As mentioned, you only need to specify the CommandClass Variables for the Configuration CommandClass, and Association Details. The rest of the CommandClass information is not required as its autodiscovered.

joshughes:
Sorry took me a while to understand auto discovery of these sleepy devices and how it all works. I appreciate your help! I am going to be working on getting my security enabled lock working next. I did see the documentation on the security stuff. So far I have it included but I don't see messages being published when I lock and unlock... so I may have to try the inclusion again i guess. 

Anyway, I hope to contribute more in the future. Thanks for everything!

Fishwaldo:
So there are no configuration values for this device (can usually be found in the device manual or sometimes you have to ask their support for their tech specs?)

joshughes:
http://www.pepper1.net/zwavedb/device/458

There is a configuration for double alarm according to this.

Fishwaldo:
Then that config item should be added to the Config File.

Fishwaldo:
Any update?

Fishwaldo:
Closing. When/if you add the configuration information, please open a new pull request.

PR #726, opened 2016-01-05
Small update

PiterEL:


Fishwaldo:
this is mostly cosmetic.. Sorry, I don't see the point...

PiterEL:
But i want to update to the same as in manual from fibaro. Should be the same.

PR #733, opened 2016-01-11
Add support for Wenzhou TKB TSM02 4 in 1 multi-sensor.

### LINKED ISSUE #719: Wenzhou TKB TSM02 4 in 1 Door/Window Sensor
I have a "Wenzhou TKB TSM02 4 in 1 Door/Window sensor", but with latest Open-ZWave it's being identified as a "TSP01 3 in 1 PIR Motion Sensor", which causes issues with door/window and PIR sensors.

TSM02 is exactly the same as "PST02-A 4 in 1 Multi-Sensor", so I made the following change and the sensor works as expected:

In file`config/manufacturer_specific.xml`, change line:

    <Product type="0002" id="0002" name="TSP01 3 in 1 PIR Motion Sensor" config="wenzhou/tsp01.xml" />

to

    <Product type="0002" id="0002" name="TSM02 4 in 1 Door/Window Sensor" config="philio/pst02.xml" />

This of course will cause TSP01 sensors not work as expected. I'm pretty new to Open-ZWave so I couldn't figure out how to add a device instead of replace, specifically what should be written to the `type` and `id` values...

ertanden:
Should resolve #719 without breaking support for TSP01, hopefully...

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -712,7 +712,7 @@
 	<Manufacturer id="0118" name="Wenzhou TKB Control System">
 		<Product type="0001" id="0001" name="TZ88 Smart Energy Plug-in Switch" config="wenzhou/tz88.xml"/>
 		<Product type="0002" id="0001" name="SM103 Door/Window Sensor" config="wenzhou/sm103.xml"/>
-		<Product type="0002" id="0002" name="TSP01 3 in 1 PIR Motion Sensor" config="wenzhou/tsp01.xml" />
+        <Product type="0002" id="0002" name="TSP01/TSM02 Multi-Sensor" config="wenzhou/tsp01.xml" />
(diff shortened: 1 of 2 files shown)


Fishwaldo:
Actually, I can't add this, as it will break TSP01 devices. (it will cause timeouts when trying to retrieve Index 3 values from a TSP01 device (that doesn't have it).

As mentioned, go bitch at the manufacturer. Trying to support something as broken as this isn't on my high list of priorities.

ben423423n32j14e:
Fishwaldo, Sir I have about 15 of these devices, they work properly in HomeSeer but I have recently moved to Domoticz using OpenZWave, I am prepared to pay you $100 USD if you can find a way to get native support for this device in OpenZWave either via PayPal or a escrow service. It makes more sense for me to be able to use these devices which cost over $1000 than have to buy all new ones.

Keep up the great work.

Cheers

PR #735, opened 2016-01-12
update dotnet wrapper

bufferUnderrun:
Hi Justin,

Can you accept this patch, it adds the Manager::GetChangeVerified() function to the dotnet wrapper to keep sync with the c++ codebase.

Thanks,

Also could you explain me the purpose of theses functions (SetChangeVerified() and GetChangeVerified()) ? 

Because i don't understand the using case of theses function : by default, all changes are notified by the controller using a Notification::ValueChanged, isn't it ?


bufferUnderrun:
excuse me, wrong pull. you can delete this.

PR #736, opened 2016-01-12
update dotnet wrapper

bufferUnderrun:
Hi Justin,

Can you accept this patch, it keeps dotnet wrapper to keep sync with the c++ codebase.

Thanks,

Also could you explain me the purpose of theses functions (SetChangeVerified() and GetChangeVerified()) ?

Because i don't understand the using case of theses function : by default, all changes are notified by the controller using a Notification::ValueChanged, isn't it ?

### DIFF
--- a/dotnet/src/ZWManager.h
+++ b/dotnet/src/ZWManager.h
@@ -1042,6 +1042,14 @@ namespace OpenZWaveDotNet
 		bool IsValueReadOnly( ZWValueID^ id ){ return Manager::Get()->IsValueReadOnly(id->CreateUnmanagedValueID()); }
 
 		/**
+		 * \brief Test whether the value is write-only.
+		 * \param _id The unique identifier of the value.
+		 * \return true if the value can only be written to and not read.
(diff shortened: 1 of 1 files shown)


bufferUnderrun:
sorry for no rebasing all the commits, git is not installed on my localhost and i don't know how to do it with github.

Fishwaldo:
SetChangeVerified causes openzwave to requery a Value when we recieve a report and check the result is the same as the initial value reported. This helps with devices that send updates before they have reached a final value. eg:
A Dimmer is at 0% and you set it to 100% (either via the device, or via ZWave). Some devices will send a report back where the dimmer value might be anywhere between 0 and 100%. With SetChangeVerified, we will requery the device, and if the value is the same, then a ValueChanged Notification goes out. If its not then we keep querying the device till the value is the same between two queries.

Fishwaldo:
Thanks. Merged into Dev Branch.

bufferUnderrun:
thanks for you answer about SetChangeVerified.

PR #737, opened 2016-01-12
 ZW15S config params

chadparry:
The ZW15S doesn't advertise its configuration parameters, so it requires manufacturer-specific data.

### DIFF
--- a/config/enerwave/zw15s.xml
+++ b/config/enerwave/zw15s.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Enerwave ZW15S -->
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+  <!-- Configuration Parameters -->
+  <CommandClass id="112">
+    <Value type="byte" index="1" genre="config" label="Toggle LED mode" min="0" max="1" value="0" size="1">
+      <Help>Set to 1 for LED to be in sync with switch.</Help>
+      <Item label="0: LED is on when switch is off" value="0" />
+      <Item label="1: LED is on when switch is on" value="1" />
+    </Value>
+    <Value type="byte" index="2" genre="config" label="Toggle button mode" min="0" max="1" value="0" size="1">
+      <Help>Set to 1 for "up" to turn the load off.</Help>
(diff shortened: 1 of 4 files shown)


nechry:
ZW15S param are ok but you mist the Association groups
http://enerwaveautomation.com/wp-content/uploads/products/zw15s/ZW15S-11052014.pdf

ZW20R have Association groups too? it's not mentioned in doc

chadparry:
I'll do some experimentation with the ZW15S and see what I need to add. I didn't realize that the Association class could be part of this config. The ozwsh tool is able to access groups with this config already, which is good. The ZW20R also supports the association, so I'll look at both of them.

chadparry:
I verified that the Association groups are detected automatically. For example, here is how the node looks after it is added to my network: https://gist.github.com/chadparry/7994df768fe2bcce7495. From what I understand, these config overrides are only to be used for information that was not detected correctly. So this patch is probably correct as-is.

nechry:
It's a good practice to add the associations group, also set a good name to the group according to device documentation in this case is the Lifeline. In some case we must or not auto associate it to the controler a attribut auto can be set, in your case you don't gave to set. 
So please add the association group for you config files.

chadparry:
Sure, I added the associations. I also threw in one more product type that I recently connected.

Fishwaldo:
what about ZW20R? No Associations Groups on that device?

chadparry:
@Fishwaldo, correct, I don't think there are Associations Groups on the ZW20R. I'll look into this again to make sure. No association classes were detected in my installation. But the spec sheet says that Association is one of the supported functions. It could mean that this outlet can be controlled by an Association, rather than that it can directly create an Association.

chadparry:
Technical support from www.enerwaveautomation.com has just confirmed for me that ZW20R does not support Associations Groups. That means that this pull request is complete.

chadparry:
Another email from Enerwave Home Automation clarified that the device can only be controlled through an association, as I suspected.

> ZW20R can act as associated slave, which means zwave device with association feature can link to our ZW20R, and control zw20R through other zwave device.

Fishwaldo:
Ok. Very strange...

PR #740, opened 2016-01-13
Added US version of Fibaro FGMS-001

technosf:
Hello,

The US version of the Fibaro fgms has a different id than the two existing. Added it and tested in Domoticz.

Thank you!


Fishwaldo:
Thanks!

PR #744, opened 2016-01-17
Integration of Danfoss room sensor 014G0160

Quax1507:
XML file and the addition to manufarturer_specific.xml to get Danfoss room sensor supported

### DIFF
--- a/config/danfoss/rs.xml
+++ b/config/danfoss/rs.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+
+        <CommandClass id="67" name="COMMAND_CLASS_THERMOSTAT_SETPOINT" version="1" request_flags="4" override_precision="2" innif="true" base="0">
+                <Instance index="1" />
+                <Value type="decimal" genre="user" instance="1" index="1" label="Furnace" units="C" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="0.0" />
+                <Value type="decimal" genre="user" instance="1" index="2" label="Cooling" units="C" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="0.0" />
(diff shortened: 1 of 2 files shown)


config/danfoss/rs.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+
+        <CommandClass id="67" name="COMMAND_CLASS_THERMOSTAT_SETPOINT" version="1" request_flags="4" override_precision="2" innif="true" base="0">
```
Fishwaldo: Why do you need to add the Thermostat CC in the config file? Does the device not auto-discover correctly?

Quax1507:
You are right. Discovery is incorrect. Do You need any details?

config/danfoss/rs.xml
Quax1507: You are right. Discovery is incorrect. Do You need any details?

Quax1507:
You are right. Discovery is incorrect. Do You need any details?

alexus1211:
it uses "COMMAND_CLASS_THERMOSTAT_SETPOINT" version "3" and not version "1" as described in rs.xml.
Did this patch fix the discovery problem? BTW, I have the same problem in Razberry (Zway).

Quax1507:
With that fix, the Thermostat works without any issues!
The discovery returns wrong value "4" for "furnace". Correct vaue is "1" instead!
Will also occur with Zway :-(

alexus1211:
Can you send me the logs for the auto discovery? What I meant with "3" is the version of the Command Class "thermostat_setpoint" that the RS supports, and not the payload data that has been transmitted.

Fishwaldo:
The version reported in the zwcfg_*.xml file is the version that OZW supports, not what the device reports. All CommandClasses are backwards compatible, so if we don't support version 3, it should still work. 

please attach the logs to this ticket after starting OZW without any zwcfg_*.xml file.

Fishwaldo:
@alexus1211 any update?

Fishwaldo:
Closing as no response.

PR #750, opened 2016-01-23
Multi-platform compatible FreeBSD build fix for iconv

### LINKED ISSUE #748: hid.c regression on FreeBSD
With commit d78c7d08c95779c6ecaa4b846fbb600896f9fa9e the FreeBSD 10.2+ support was broken again, as fixed in #701.

@durin42: What FreeBSD version was that commit tested on? You only stated clang version

stromnet:
Revert hid.c change in d78c7d0, and enforce older FreeBSD versions to
use libiconv from ports instead of old non-posix from base.

Fixes #748, tested on both 10.1 and 10.2. Should work on 9.3 too.


Fishwaldo:
Thanks!

PR #758, opened 2016-01-31
Add support for Fakro Roof Windows Roller Shutter ARZ Z-Wave

pepeEL:


config/fakro/arz-z-wave.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+  <!-- Configuration -->
+  <CommandClass id="112">
```
Fishwaldo: If there is no configuration items, then this section can be dropped.

config/fakro/arz-z-wave.xml
```diff
@@ -0,0 +1,14 @@
...
+  </CommandClass>
+
+  <!-- Association Groups -->
+  <CommandClass id="133">
```
Fishwaldo: You should put in Association Information.

config/fakro/arz-z-wave.xml
```diff
@@ -0,0 +1,14 @@
...
+    <Associations>
+    </Associations>
+  </CommandClass>
+
+</Product>
```
Fishwaldo: If the file really ends up "empty" then we can drop the file entirely.

config/fakro/arz-z-wave.xml
pepeEL: But from where i can get Association Group ? Becuase now when i use it in domoticz it has no association group.
ass

config/fakro/arz-z-wave.xml
pepeEL: And from where i can get option to configuration ? In manual from roller i have no option to config.

nechry:
Hi the ARZ Roof Window Roller Shutter don't have association groups or configuration parameters. we tested this device. see:
 https://github.com/jeedom/plugin-openzwave/blob/master/ressources/openzwave/config/manufacturer_specific.xml#L265

the file config/fakro/arz-z-wave.xml is not require.

pepeEL:
Ok thanks i correct it in new PR #760

PR #759, opened 2016-02-01
Qubino Submitted Config Files

Fishwaldo:
Merge in Qubino Supplied Config Files...

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -582,17 +582,23 @@
 	</Manufacturer>
 	<Manufacturer id="0159" name="Qubino">
 		<Product type="0001" id="0001" name="ZMNHDA2 Flush Dimmer" config="qubino/ZMNHDA2.xml"/>
-		<Product type="0001" id="0051" name="ZMNHDD1 Z-Wave Plus Flush Dimmer" config="qubino/ZMNHDD1.xml"/>
 		<Product type="0002" id="0001" name="ZMNHBA2 Flush 2 Relays" config="qubino/ZMNHBA2.xml"/>
-		<Product type="0002" id="0051" name="ZMNHBD2 Flush 2 Relays" config="qubino/ZMNHBD2.xml"/>
 		<Product type="0002" id="0002" name="ZMNHAA2 Flush 1 Relay" config="qubino/ZMNHAA2.xml"/>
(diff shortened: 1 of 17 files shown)


PatrickEekhout:
Hi there,

I noticed this activity about the Qubino config files and would like to discuss some matters with you. I don't know if this is the right path, if not, I hope you can point me in the right direction.

As I have some problems getting the Qubino Flush 2 Relays ZMNHBD to properly work in Domoticz together with manually switching the device through I1 and I2, my search led me to here.

I studied the Qubino manuals for all their devices. It seems that they use different "order codes" for each frequency. Perhaps you can say that the first part of the order code defines the device/product-code and the last digit defines the frequency-version.

For example, the Qubino Flush 2 Relays, has the following ordering codes according to the manual:
ZMNHBD1 868,4 MHz
ZMNHBD2 921,4 MHz
ZMNHBD3 908,4 MHz
ZMNHBD4 869,0 MHz
ZMNHBD5 916,0 MHz
ZMNHBD8 865,2 MHz

In fact it is the same device (with the same device-id?) so no need for different config-files I guess. Suggested options for naming the config files in my opinion are: ZMNHBD or ZMNHBDx

At the moment there are multiple config files for the same device. For example: ZMNHBD1.xml and ZMNHBD2.xml. Where ZMNHBD2.xml is incomplete, as it misses the temperature-id entries.

There are multiple occurrences in the Qubino config files with label <Help/>. This is incorrect: the opening command <Help> is missing and in the closing command <Help/> the slash should be at the beginning instead of at the end.

This results in the following issues:
Issue 1: Naming of config files
Issue 2: Multiple config files for the same device
Issue 3: Incomplete ZMNHBD2.xml (temperature)
Issue 4: Incorrect label <Help/>

Hope to hear from you soon.

Regards,
Patrick

nechry:
Hi Patrick it's also my approach for zone free naming see 
https://github.com/jeedom/plugin-openzwave/blob/master/ressources/openzwave/config/manufacturer_specific.xml#L590

For parameters it's little fight, we got many version of config file, and we try to improve day by day like aeon-labs or fibaro config files.

But in regard of config files, the main issue with qubino still with CC implementation. 

thank for sharing your point, I don't think I'm fully right in my working config but there one are tested with many users

Regards,
Nechry

PR #762, opened 2016-02-01
Create scsc17.xml

russdan:
Fixes issues with the thermostat setpoints which have the incorrect values by default. The attached file has been tested with my horstmann thermostat and can now correctly set the thermostat via openzwave.


Fishwaldo:
You shouldn't do a straight copy paste from the zwcfg_*.xml file. I'll fit it up.

Also, you can do everything in one pull request, rather than a seperate PR for the manufacturer_specific.xml file.

PR #767, opened 2016-02-03
Correct update param. from manual for Fibaro FGS221

pepeEL:


Fishwaldo:
For future references, could you please try to keep the label strings shorter? It looks horible with long labels on some applications!

PR #780, opened 2016-02-19
Add D-Link dch-z120 PIR/motion sensor

mattwire:


Fishwaldo:
You cannot have negative numbers in Z-Wave. (The protocol doesn't support signed values). So your first entry inbthe config file should be redone please.

mattwire:
Hi, I took that straight from the manual from dlink.  Have now updated to remove the negative number.

Issue #771, opened 2016-02-08
Header files in subfolders have incorrect includes


julienw:
Most headers include other headers using `#includes "Header.h"`. But the quote includes look in the same directory, and in directories specified by -I then.

But the headers in the subdirectories (eg: value_classes, or platform), that most include other open-zwave header using eg `#include "Defs.h"` fail when we don't use `-I path-to-openzwave-headers` to build.

This is too bad because this could be easy to fix by using relative paths instead.

julienw:
Includes like `#include "value_classes/Value.h"` in `value_classes/ValueList.h` yield the same issue.

julienw:
Note this is especially useful with the headers installed globally (for example using the debian `-dev` package). If this bug is fixed, we can just use `#include <openzwave/Options.h>" etc and we don't need any `-I` options, which feels right to me.

Fishwaldo:
I'm well aware of the benifits, but this  would break existing builds for a lot of 3rd party apps that use OZW and give instructions to their users how to compile etc. 

Read the mailing lists/other issues for previous discussions on it. The change wont happen in a 1.x timeframe

julienw:
Actually there are 2 different issues here. Let me try to explain.

1. The "main" header files (eg: Options.h, Manager.h). For these ones you need -I if they're not in the usual locations. And... why not. It would be a lot easier if the repo's layout would have the installed layout (ie: a `openzwave` subdirectory where the headers are). But I agree this is your choice.

2. The subheader files that are not used directly by clients (the ones in the subfolders `value_classes`, or `platform`, etc). These ones have incorrect includes, and using `#include "../Defs.h"` (for example) would just work without needing a -I clause, because the preprocessor is looking relatively from the including file.

Fishwaldo:
1) It's been this way since the inception of OZW. Not my choice. 
2) as mentioned, changing any include files/paths will break compiles for dozens of third party Apps based on OZW and make me unpopular with those authors 
3) you can always use pkg-config (or ozw_config) to setup the right include paths for your Makefiles etc, which means even if users install in weird paths (like opt) then it still works.

julienw:
I'm focussing on the issue 2. here. What I'm saying is that changing these should have zero impact on users.

[...]

PR #786, opened 2016-03-02
Issue #771 - We need to have "includedir" into the -I. Do that with pkg-config

hfiguiere:
This should address issue #771 without breaking other people's build.

Until the root problem gets fixed.

Also had to sanitize white-space in Makefile as Emacs was complaining.


Fishwaldo:
As discussed, I wont merge this.

PR #797, opened 2016-03-12
Device Support: Leviton VRMX1-1LZ - additional hardware id

shanemadden:
Attach the right name and config to the version of the VRMX1-1LZ that I got (newer rev maybe? Ordered at http://smile.amazon.com/Leviton-VRMX1-1LZ-Universal-Magnetic-Voltage/dp/B005Y8JC6Q).


Fishwaldo:
Thanks!

PR #799, opened 2016-03-12
Device Support: GE 12724 Dimmer configuration params

shanemadden:
Add configuration support for the GE 12724 3-Way Dimmer Switch.

It supports slightly different parameters than ge/dimmer.xml, so I split it out into its own file.


Fishwaldo:
Please confirm there is no association groups for this device.

shanemadden:
@Fishwaldo Confirmed, no association groups on here - which makes sense, since this thing has no support for sending any kind of update or hail packet and must be polled.

Fishwaldo:
weird.. if no association groups, it can't report updates when someone manually changes the device. Poor Design Choice... 

But I'll merge it anyway soon.

shanemadden:
@Fishwaldo Yeah, as I understand it there's a US patent which covers local change status reporting, so some manufacturers just don't bother with any sort of reporting.

But, hold off on merging this, since I just ran into a bit in a random forum that suggested this model should support sending hail messages.. so let me dig some more on this first.

shanemadden:
@Fishwaldo I'm not sure what I had wrong on my attempt at getting it to take a group 1 the first time, but it *does* support group 1 - and that group gets `UPDATE_STATE_NODE_INFO_RECEIVED` for instant status updates! My rating of these switches just went up considerably. I'll add a new commit with that association group later on today.

shanemadden:
@Fishwaldo  Sorry for the delay on this, the association group implementation on this device seems.. buggy at best?

It seems to never respond to the `AssociationCmd_Get` call - is there anything that would cause this in normal operation that I might have messed up, or does that just mean it might actually not have association group support implemented?

```
2016-04-22 22:12:49.070 Info, Node009, Sending (Send) message (Callback ID=0x3c, Expected Reply=0x04) - AssociationCmd_Get (Node=9): 0x01, 0x0a, 0x00, 0x13, 0x09, 0x03, 0x85, 0x02, 0x01, 0x25, 0x3c, 0x73
2016-04-22 22:12:49.078 Detail, Node009,   Received: 0x01, 0x04, 0x01, 0x13, 0x01, 0xe8
2016-04-22 22:12:49.078 Detail, Node009,   ZW_SEND_DATA delivered to Z-Wave stack
2016-04-22 22:12:49.102 Detail, Node009,   Received: 0x01, 0x07, 0x00, 0x13, 0x3c, 0x00, 0x00, 0x03, 0xd4
2016-04-22 22:12:49.102 Detail, Node009,   ZW_SEND_DATA Request with callback ID 0x3c received (expected 0x3c)
2016-04-22 22:12:49.102 Info, Node009, Request RTT 32 Average Request RTT 32
2016-04-22 22:12:49.103 Detail,   Expected callbackId was received
2016-04-22 22:12:59.071 Error, Node009, ERROR: Dropping command, expected response not received after 1 attempt(s)
2016-04-22 22:12:59.072 Detail, Node009, Removing current message
2016-04-22 22:12:59.073 Detail, Node009, Notification: Notification - TimeOut
```

I have two of these devices, one currently sending instant updates and one not (with matching configuration classes, unresponsive association groups, and no apparent way to get the behavior to change on either device via anything exposed in ozwcp), and while I *think* it was the attempt at setting the association group that triggered it, I'm not certain. I'm hoping to have some time this weekend to factory reset them a few times and nail down the behavior of the device a bit better, but I'd appreciate your input on what you've seen with other devices' association group behavior!

(and thanks again!)

shanemadden:
Actually, that was a very simple mystery once I tried that one last thing.

The switch that was not working is installed as a 3-way switch, with the 12723 add-on switch in the 3-way location. When I remembered that the further away switch was the 12724, and the one I'd been using to locally trigger the switch was the 12723, the behavior became clear.

Instant updates behavior for the GE 12724 seems to be this:

 - The device does not support any association groups - the hail has a static target of the network's primary controller, apparently?
 - The device sends a 'hail' which in OZW triggers UPDATE_STATE_NODE_INFO_RECEIVED when the hardware button on the 12724 is pushed. Critically, **the packet is *not* sent if the hardware button push was on a remote-switch in a 3-way installation**. Polling seems to be required for these installations. Bummer.

 So, disappointing results, but, this is ready to merge as is; my final answer is that there truly are no association groups. Thanks!

Fishwaldo:
ok.

PR #800, opened 2016-03-13
Device Support: Enerwave ZW20RM configuration params

shanemadden:
Add configuration parameters for Enerwave ZW20RM 20A Smart Meter Duplex Receptacle


Fishwaldo:
Could you confirm there is no Association Groups for this device?

Fishwaldo:
Thanks

PR #802, opened 2016-03-15
Various changes around the Debian package

julienw:
* properly generate the pkgconfig file in libdir if there is a pkgconfig directory
* Simplified the awk command in the process
* include the pkgconfig file in the Debian package
* update the Debian changelog to get a 1.4 version

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -13,9 +13,9 @@ endif
 
 DEBVERS := $(shell dpkg-parsechangelog | sed -n -e 's/^Version: //p')
 VERSION := $(shell echo '$(DEBVERS)' | sed -e 's/^[\[:digit:]]*://' -e 's/[-].*//')
-MAJOR := $(shell echo '$(VERSION)' | awk -F . {'print $$1'})
-MINOR := $(shell echo '$(VERSION)' | awk -F . {'print $$2'})
-REV := $(shell echo '$(VERSION)' | awk -F . {'print $$3'})
+MAJOR := $(shell echo '$(VERSION)' | awk -F '[.+]' {'print $$1'})
+MINOR := $(shell echo '$(VERSION)' | awk -F '[.+]' {'print $$2'})
+REV := $(shell echo '$(VERSION)' | awk -F '[.+]' {'print $$3'})
 DEB_HOST_ARCH := $(shell dpkg-architecture -qDEB_HOST_ARCH)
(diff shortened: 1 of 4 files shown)


julienw:
@lnussbaum maybe you can r? the debian package parts, since you did the initial work ?

debian/rules
```diff
@@ -13,9 +13,9 @@ endif
 
 DEBVERS := $(shell dpkg-parsechangelog | sed -n -e 's/^Version: //p')
 VERSION := $(shell echo '$(DEBVERS)' | sed -e 's/^[\[:digit:]]*://' -e 's/[-].*//')
-MAJOR := $(shell echo '$(VERSION)' | awk -F . {'print $$1'})
-MINOR := $(shell echo '$(VERSION)' | awk -F . {'print $$2'})
-REV := $(shell echo '$(VERSION)' | awk -F . {'print $$3'})
+MAJOR := $(shell echo '$(VERSION)' | awk -F '[.+]' {'print $$1'})
+MINOR := $(shell echo '$(VERSION)' | awk -F '[.+]' {'print $$2'})
+REV := $(shell echo '$(VERSION)' | awk -F '[.+]' {'print $$3'})
```
julienw: This is because I switched to using a "+" for the version, eg `1.4.1+79~gfaea7dd`. This means 79 commits after the 1.4.1 tag and this seems to be what's recommended for git snapshots these days. This is found using `git --describe --tags`

Fishwaldo: The Rev Value generated by the Makefile etc is already "Commits" since tag... 
so 1.4.1 is 1 commit after 1.4 was tagged on master branch.

julienw: ah I see; so I guess I should use the same format as previously: 1.4.79. Thx I'll update this.

julienw:
Added a commit to revert to the previous version format.

julienw:
Hey, is there a specific procedure to get this reviewed and merged ?

Thanks !

lnussbaum:
@julienw hi! I'm pretty busy at the moment, but added this to my TO-DO list. I wouldn't mind if someone beat me to it ;)

Fishwaldo:
@julienw I'll merge soon. Just come off a rather crazy travel schedule with work.

Fishwaldo:
Thanks. Merged into Dev Branch.

PR #805, opened 2016-03-16
New device: Everspring AN181 Miniplug with meter function

FredrikFornstad:
Tested with latest snapshot of Domoticz and works reasonable OK. 

All control seems to work ok, but the meter support does not work for kWh or power factor.
Not sure if the Notification reports (device powered on, or overload situation) works.
If communication is temporarily lost, or device experience power failure: then Domoticz reports the switch as ON when communication is restored even though it may be in OFF position... -> Is the Notification Report (device powered on) misinterpreted as the switch is in ON position?


Fishwaldo:
Thanks!

PR #810, opened 2016-03-29
Add Enerwave ZWN-BPC configuration

robbiet480:
Add a configuration for the Enerwave ZWN-BPC Ceiling Mounted PIR Motion Sensor.


Fishwaldo:
Thanks

PR #811, opened 2016-03-29
#PR-811

halsafar:
- added dimmer.xml config for GE 12724 3-Way Dimmer Switch
- updated changelog


halsafar:
Noticed by GE 12724 3-Way Dimmer had no config options in Domoticz.  This turned out to be the cause, no config info for it.

Fishwaldo:
Thanks

PR #817, opened 2016-04-06
Remove multi-arch fat binaries as it's not supported in MacOS X anymo…

### LINKED ISSUE #814: There is no easy way to disable generating a fat binary on MacOS X
I'd like to generate a thin binary instead of a fat binary on MacOS X (for static linking with Rust). This is currently not easily possible because the LIBS make variable has more things than just `-arch` and it would be error prone to set it from the outside.

julienw:
…re (Fixes #814)


julienw:
This is maybe quite drastic, but I believe this is not useful anymore in today's world. Tell me what you think.

Fishwaldo:
Hi,

Thanks

PR #818, opened 2016-04-06
Use generic `cc` and `c++` instead of `gcc` and `g++` (Closes #816)

### LINKED ISSUE #816: Building on MacOS X 10.10 is broken when gcc is installed -- needs to use clang instead of gcc
After hacking around to fix https://github.com/OpenZWave/open-zwave/issues/814 on MacOS X 10.10, we have this additional issue. Looks like building with gcc is not supported when using MacOS X' headers.

See the error here:
```
➜  open-zwave git:(master) ✗ make
CPPFLAGS= /Applications/Xcode.app/Contents/Developer/usr/bin/make -C /Volumes/mozilladev/foxbox/open-zwave/cpp/build/ -
Building OpenZWave Version 1.4-79-g0d716a3-dirty
Building tinystr.o
Building tinyxml.o
Building tinyxmlerror.o
Building tinyxmlparser.o
Building hid.o
In file included from /usr/include/dispatch/dispatch.h:51:0,
                 from /System/Library/Frameworks/CoreFoundation.framework/Headers/CFStream.h:15,
                 from /System/Library/Frameworks/CoreFoundation.framework/Headers/CFPropertyList.h:13,
                 from /System/Library/Frameworks/CoreFoundation.framework/Headers/CoreFoundation.h:55,
                 from /System/Library/Frameworks/IOKit.framework/Headers/hid/IOHIDLib.h:27,
                 from /System/Library/Frameworks/IOKit.framework/Headers/hid/IOHIDManager.h:29,
                 from /Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c:25:
/usr/include/dispatch/object.h:143:15: error: expected identifier or ‘(’ before ‘^’ token
 typedef void (^dispatch_block_t)(void);
               ^
/usr/include/dispatch/object.h:362:3: error: unknown type name ‘dispatch_block_t’
   dispatch_block_t notification_block);
   ^
/Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c: In function ‘get_string_property_utf8’:
/Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c:285:11: error: variable ‘chars_copied’ set but not used [-Werror=unused-but-set-variable]
   CFIndex chars_copied;
           ^
/Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c: In function ‘hid_enumerate’:
/Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c:450:11: error: variable ‘len’ set but not used [-Werror=unused-but-set-variable]
    size_t len;
           ^
/Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c: In function ‘hid_open_path’:
/Volumes/mozilladev/foxbox/open-zwave/cpp/hidapi/mac/hid.c:705:10: error: variable ‘len’ set but not used [-Werror=unused-but-set-variable]
   size_t len;
          ^
cc1: all warnings being treated as errors
make[1]: *** [/Volumes/mozilladev/foxbox/open-zwave/.lib/hid.o] Error 1
make: *** [all] Error 2
```

I think that always building with `cc` and `c++` instead of `gcc` and `g++` should do the trick on most systems. Do you know why we use cc/c++ on FreeBSD only ? (cf https://github.com/OpenZWave/open-zwave/blob/0d716a3526eea2c23691c90073c3c5895af3bd24/cpp/build/support.mk#L47-L56)

julienw:


julienw:
I tested this on a linux computer and a MacOS X computer that has only clang (and mapped to gcc). I'll try on the MacOS X computer where bug #816 happens soon.

julienw:
I checked this fixes issue #816 on that computer too.

Fishwaldo:
something is messed up on your "troublesome" PC. On All my mac's, cc and gcc are clang:
``
BUNSO29:open-zwave justinhammond$ which cc
/usr/bin/cc
BUNSO29:open-zwave justinhammond$ ls -lah /usr/bin/cc
lrwxr-xr-x  1 root  wheel     5B Oct  2  2015 /usr/bin/cc -> clang
BUNSO29:open-zwave justinhammond$ which gcc
/usr/bin/gcc
BUNSO29:open-zwave justinhammond$ ls -lah /usr/bin/gcc
-rwxr-xr-x  1 root  wheel    18K Mar 12 16:35 /usr/bin/gcc
BUNSO29:open-zwave justinhammond$ gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/c++/4.2.1
Apple LLVM version 7.3.0 (clang-703.0.29)
Target: x86_64-apple-darwin15.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
BUNSO29:open-zwave justinhammond$ cc -v
Apple LLVM version 7.3.0 (clang-703.0.29)
Target: x86_64-apple-darwin15.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
BUNSO29:open-zwave justinhammond$ uname -a
Darwin BUNSO29.dmz.dynam.ac 15.4.0 Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64 x86_64
``

julienw:
Yes, on the Mac I tried, it's like you. I believe that on my colleague's Mac gcc was installed to work on something else (likely Firefox OS). Which is OK as long as it doesn't use the platform's header (eg: when cross compiling).

Still that means building openzwave is broken as soon as the user installed gcc, even if it's not common. And because this is easy to fix, I think we should do it :)

Do you know why we were not using cc/c++ before? On my Linux, cc points to gcc, and c++ to g++. Is it because we want to support odd environments where the default C compiler is bad ?

Another solution (or rather an _additional_ solution to this patch) is to set CC/CXX/LD using `?=` instead of `:=` so that we can override them from the environment.

Last but not least, we could set _nothing_ related to the compilers in the Makefile. Defaults set by make looks quite good already (check with `make -p`): on my Mac it's `cc`/`c++`, on my Linux it's `cc`/`g++`. And `ld` as default for `LD` but there is also a LINK.cc we could use. (I admit I don't know why we don't use `ld`).

To cross compile, the default make rules use a `TARGET_ARCH` variable that could be used instead of CROSS_COMPILE. But that could be a backward incompatible change.

julienw:
I think this breaks cross compiling because some systems don't have the prefixed cc and c++. Likely the best is to allow passing CC and CXX by using `?=` instead of `:=`.

Fishwaldo:
that is most likely what I will do... sane defaults with a easy way to override.

Fishwaldo:
Hi,

Thanks

Issue #787, opened 2016-03-03
Fix Command_Class_Alarm everspringct/hsm02.xml


Nico0084:
Hi
 There is some errors in HSM02 file config,
Command class alarm is on version 1 and device isn't a multi-instance
See attached file hsm02.xml.txt to fix it
[hsm02.xml.txt](https://github.com/OpenZWave/open-zwave/files/156431/hsm02.xml.txt)

I propose another solution to fix that with a ValueTypeList for "Alarm Level"
But we should modify ozw code, and I not sure of other effect in it.
Command_class_alarm create byte values, and in this case we should remove one and replace it with type list.
On my proposal, this is doing during void Node::ReadValueFromXML.
If there is diffence type between  Value created with command_class and Value from xml config, replace process is do.
This must ensure that xml file have no error, otherwise I'm not sure of side effect !

But with this replacement we could have a better integration and much flexibility to handle device, not ?

see git diff of alarm.cpp and node.cpp  and new HSM02 xml : 
```
diff --git a/cpp/src/command_classes/Alarm.cpp b/cpp/src/command_classes/Alarm.cpp
index 3dad19c..c2934eb 100644
--- a/cpp/src/command_classes/Alarm.cpp
+++ b/cpp/src/command_classes/Alarm.cpp
@@ -229,12 +229,30 @@ bool Alarm::HandleMsg
                        value->OnValueRefreshed( _data[1] );
                        value->Release();
                }
-               if( (value = static_cast<ValueByte*>( GetValue( _instance, AlarmIndex_Level ) )) )
+               // For device on version 1 the level could have different value. This level value corres
+               if ( Value* value = GetValue( _instance, AlarmIndex_Level ) )
                {
-                       value->OnValueRefreshed( _data[2] );
+                       switch ( value->GetID().GetType() )
+                       {
+                               case ValueID::ValueType_Byte:
+                               {
+                                       ValueByte* valueByte = static_cast<ValueByte*>( value );
+                                       valueByte->OnValueRefreshed( _data[2] );
+                                       break;
+                               }
+                               case ValueID::ValueType_List:
+                               {
+                                       ValueList* valueList = static_cast<ValueList*>( value );
+                                       valueList->OnValueRefreshed( _data[2] );
+                                       break;
+                               }
+                               default:
+                               {
+                                       Log::Write( LogLevel_Info, GetNodeId(), "Invalid type (%d) for A
+                               }
+                       }
                        value->Release();
                }
-
[...]

PR #840, opened 2016-04-11
Modify for implement replace value type from xml, see #787

Nico0084:


Fishwaldo:
Thanks.

I've pulled this into the Dev Branch... 

@nechry This would be useful for other devices that document the Alarm Class...

PR #841, opened 2016-04-11
Corrected description for Aeotec Heavy Duty Smart Switch Gen5: LED Status

Mashi:
Corrected description for Aeotec Heavy Duty Smart Switch Gen5: Configuration value 20 is not just LED status after power on, but even the switch status setting after power on (power loss).

### DIFF
--- a/config/aeotec/hdss_gen5.xml
+++ b/config/aeotec/hdss_gen5.xml
@@ -16,7 +16,7 @@
       <Item label="Enable" value="1" />
     </Value>
     <Value type="list" index="20" genre="config" label="LED Mode" size="1" value="0">
-      <Help>LED status after power on</Help>
+      <Help>Switch status after power on</Help>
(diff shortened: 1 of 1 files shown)


nechry:
in official doc is only wrote Led
https://aeotec.freshdesk.com/helpdesk/attachments/6009584699

but if we change le help description we also need to change the label ?

Is more "Switch state recovery" or something better

I ask clarification with aeon-labs.

nechry:
the current label and help description is right. I got answer from Field Application Engineer and he confirm this behaviour:

> Param 20 is for the LED status only, so typically, when the HDSS is ON status the LED will remain solid on.
> But if this setting was set to 1, then the LED will always be on even if the status is OFF.
> If set to 2, the LED will always be off even if the status of the HDSS is ON.
> For the HDSS, after a power outrage, the last status will always return back to its status before power has shut down.

Please reject this PR.

PR #842, opened 2016-04-12
Update manufacturer_specific.xml for add GR-105N

narfight:
Add line for the GR-105N device. see http://www.pepper1.net/zwavedb/device/790


Fishwaldo:
Thanks!

PR #844, opened 2016-04-12
Fix rpm spec file

thomasjfox:
- Change to the directory name created by "make dist"
  Otherwise it tries to change the directory to "libopenzwave-%{version}" and fails

- Enable parallel make build

- Add unpackage ozw_config binary to the libopenzwave-devel package

- Set current version number: 1.4.167

### DIFF
--- a/openzwave.spec
+++ b/openzwave.spec
@@ -11,7 +11,7 @@ License: LGPL-2.0+
 %else 
 License: LGPLv2+
 %endif
-Version: 1.4.0
+Version: 1.4.167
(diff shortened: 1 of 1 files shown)


Fishwaldo:
I need to do some testing, as our RPM and DEB files are built on a private instance of OpenBuildService, but this should be ok. Will merge after I had a chance to test.

Fishwaldo:
I pushed in a slightly different version that works with our OBS. Thanks

PR #849, opened 2016-04-13
Update Makefile to support building with bitbake environment

binli71:
Update Makefile and support.mk to support building with bitbake environment. Such as Yocto Project.


binli71:
This pull request should go with https://github.com/OpenZWave/open-zwave/pull/850 together

Issue #847, opened 2016-04-13
segment fault when getting zero size user code packet


binli71:
When using openzwave controller in security mode connecting to door lock ( security device ), the device could send to controller " User Code Packet is 0". The " User Code Packet is 0 " will trigger following segment fault.
"
[39m[39m2016-04-12 20:00:25.815 Info, Node002, Sending (Query) message (Callback ID=0x01, Expected Reply=0x04) - Nonce_Report - 0x01, 0x11, 0x00, 0x13, 0x02, 0x0a, 0x98, 0x80, 0x2f, 0x16, 0x64, 0x0f, 0x3d, 0xdd, 0xa5, 0x3b, 0x05, 0x01, 0xc5:
[39m[34m2016-04-12 20:00:25.823 Detail, Node002,   Received: 0x01, 0x04, 0x01, 0x13, 0x01, 0xe8
[39m[34m2016-04-12 20:00:25.823 Detail, Node002,   ZW_SEND_DATA delivered to Z-Wave stack
[39m[34m2016-04-12 20:00:26.991 Detail, Node002,   Received: 0x01, 0x05, 0x00, 0x13, 0x01, 0x00, 0xe8
[39m[34m2016-04-12 20:00:26.991 Detail, Node002,   ZW_SEND_DATA Request with callback ID 0x01 received (expected 0x01)
[39m[39m2016-04-12 20:00:26.992 Info, Node002, Request RTT 3693 Average Request RTT 2958
[39m[34m2016-04-12 20:00:27.042 Detail, Node002,   Received: 0x01, 0x1e, 0x00, 0x04, 0x00, 0x02, 0x18, 0x98, 0x81, 0xbf, 0xee, 0xb3, 0xc0, 0x43, 0x72, 0x76, 0xb6, 0xa4, 0x71, 0x58, 0xe4, 0x36, 0x2f, 0x58, 0x76, 0xa4, 0xe0, 0xac, 0x7d, 0x1d, 0x1b, 0xf8
[39m[39m2016-04-12 20:00:27.043 Info, Raw: 0x98, 0x81, 0xbf, 0xee, 0xb3, 0xc0, 0x43, 0x72, 0x76, 0xb6, 0xa4, 0x71, 0x58, 0xe4, 0x36, 0x2f, 0x58, 0x76, 0xa4, 0xe0, 0xac, 0x7d, 0x1d, 0x1b, 0xf8
[...]

Issue #848, opened 2016-04-13
Makefile change to support building openzwave in bitbake environment ( Yocto Project )


binli71:
I would suggest following change in the 
cpp/build/support.mk
cpp/build/Makefile
in order to build openzwave in bitbake environment as a layer in yocto project.

The patch details:

diff --git a/cpp/build/support.mk b/cpp/build/support.mk
index 3f74c7e..66dc42f 100644
--- a/cpp/build/support.mk
+++ b/cpp/build/support.mk
@@ -43,6 +43,15 @@ endif
 # version number to use on the shared library
 VERSION := $(VERSION_MAJ).$(VERSION_MIN)
 
+# using setting from bitbake
+ifeq ($(BITBAKE_ENV),1)
+CC     := $(CC)
+CXX    := $(CXX)
+LD     := $(CXX)
+AR     := $(AR)
+RANLIB := $(RANLIB)
+else
+
 # support Cross Compiling options
 ifeq ($(UNAME),FreeBSD)
 # Actually hide behind c++ which works for both clang based 10.0 and earlier(?)
@@ -61,6 +70,8 @@ else
 AR     := $(CROSS_COMPILE)ar rc
 RANLIB := $(CROSS_COMPILE)ranlib
 endif
+
+endif
 SED    := sed
 
 
@@ -96,9 +107,15 @@ else
 pkgconfigdir ?= $(shell pkg-config --variable pc_path pkg-config | awk '{split(
 endif
 
+ifeq ($(BITBAKE_ENV),1)
+sysconfdir := $(PREFIX)/etc/openzwave/
+includedir := $(PREFIX)/include/openzwave/
+docdir := $(PREFIX)/share/doc/openzwave-$(VERSION).$(VERSION_REV)
+else
 sysconfdir ?= $(PREFIX)/etc/openzwave/
 includedir ?= $(PREFIX)/include/openzwave/
 docdir ?= $(PREFIX)/share/doc/openzwave-$(VERSION).$(VERSION_REV)
+endif
 
 top_builddir ?= $(CURDIR)
 export top_builddir
diff --git a/cpp/build/Makefile b/cpp/build/Makefile
index a632f79..31af00a 100644
--- a/cpp/build/Makefile
+++ b/cpp/build/Makefile
@@ -93,6 +93,10 @@ else
 hidapi := $(notdir $(wildcard $(top_srcdir)/cpp/hidapi/linux/*.c)) # we do not 
 endif
 
[...]

PR #852, opened 2016-04-13
fix segment fault when getting zero size user code packet in openzwave library

binli71:
The patch fixes segment fault when getting zero size user code packet in openzwave library in order to address the following issue.

https://github.com/OpenZWave/open-zwave/issues/847


binli71:
Also added support to build with bitbake for Yocto Project in order to address following issue:
https://github.com/OpenZWave/open-zwave/issues/848

PR #855, opened 2016-04-18
Print the names for nodes and values that have beed detected in the network.

alan-martinovic:
I've found the basic example somewhat unclear when starting to experiment with the library.
The suggested changes hopefully make it more clear on how to access the nodes available in the network.


Fishwaldo:
Seems fine. I love PR's that help other users understand OZW!

Many Thanks

PR #858, opened 2016-04-19
fix for issue 847

binli71:
Signed-off-by: Bin Li <bin.li@windriver.com>

This patch fixes the issue 847 "segment fault when getting zero size user code packet".

https://github.com/OpenZWave/open-zwave/issues/847

### DIFF
--- a/cpp/src/command_classes/UserCode.cpp
+++ b/cpp/src/command_classes/UserCode.cpp
@@ -262,10 +262,11 @@ bool UserCode::HandleMsg
 			m_userCodesStatus[i] = _data[2];
 			if (size > 0) {
 				memcpy( data, &_data[3], size );
-				value->OnValueRefreshed( data, size );
 			} else {
-				value->OnValueRefreshed( 0x00, 1);
+				size = 1;
+				data[0] = 0;
 			}
+			value->OnValueRefreshed( data, size );
 			value->Release();
 		}
 		Log::Write( LogLevel_Info, GetNodeId(), "Received User Code Report from node %d for User Code %d (%s)", GetNodeId(), i, CodeStatus( _data[2] ).c_str() );


cpp/src/command_classes/UserCode.cpp
```diff
@@ -264,7 +264,8 @@ bool UserCode::HandleMsg
 				memcpy( data, &_data[3], size );
 				value->OnValueRefreshed( data, size );
 			} else {
-				value->OnValueRefreshed( 0x00, 1);
+				size = 0;
+				value->OnValueRefreshed( (uint8 const*)&size, 1);
 			}
```
dhylands: I'm not familiar enough with the coding styles of the openzwave maintainers, but personally, I would be inclined to code this using something like the following (I try to avoid casts if I can).
```
uint8 zero = 0;
value->OnValueRefreshed(&zero, 1);
```
but otherwise, your PR looks good. It might be worth mentioning that the bug being fixed was introduced by commit 0d716a3526eea2c23691c90073c3c5895af3bd24

binli71: Thanks, Dave,

To follow your CI comment, should the procedure be:

- close the PR858
- edit the code
- create a new PR

Regards
Bin

dhylands: github uses branch based PRs, so you don't need to close the PR and create a new one. If I were going to make the above change, then I would do the following:
```
git checkout branch-which-contains-the-changes
edit the files
git add the-files
git commit --amend
git push -f my-remote branch-name
```
This will replace the old commit with the new one, however, no email notification will be sent out, so I normally go and add a comment to the PR to let reviewers know that a change has happened.

If you're dealing with multi-commit PRs, then I would add a new commit (i.e. don't use --amend) and use rebase -i to merge the new commit into the appropriate existing commit.

binli71: Thanks, Dave,

Re-worked the fix and taking the casting out.

https://github.com/OpenZWave/open-zwave/pull/858/commits/4314523d320f96fd67a8692c3310bd7fedd7bc94

Regards
Bin

Fishwaldo:
Thanks

PR #861, opened 2016-04-19
fix for issue 848

binli71:
Signed-off-by: Bin Li <bin.li@windriver.com>

Makefile change to support building openzwave in bitbake environment ( Yocto Project )

https://github.com/OpenZWave/open-zwave/issues/848

### DIFF
--- a/cpp/build/support.mk
+++ b/cpp/build/support.mk
@@ -43,6 +43,15 @@ endif
 # version number to use on the shared library
 VERSION := $(VERSION_MAJ).$(VERSION_MIN)
 
+# using seting from bitbake
+ifeq ($(BITBAKE_ENV),1)
+CC     := $(CC)
+CXX    := $(CXX)
+LD     := $(CXX)
+AR     := $(AR)
+RANLIB := $(RANLIB)
+else
+
 # support Cross Compiling options
 ifeq ($(UNAME),FreeBSD)
 # Actually hide behind c++ which works for both clang based 10.0 and earlier(?)
@@ -60,6 +69,8 @@ RANLIB := ranlib
 else
 AR     := $(CROSS_COMPILE)ar rc
 RANLIB := $(CROSS_COMPILE)ranlib
+endif
+
 endif
 SED    := sed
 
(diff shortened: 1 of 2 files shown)


cpp/build/support.mk
```diff
@@ -96,9 +107,15 @@ else
 pkgconfigdir ?= $(shell pkg-config --variable pc_path pkg-config | awk '{split($$0,a,":"); print a[1]}')
 endif
 
+ifeq ($(BITBAKE_ENV),1)
+sysconfdir := $(PREFIX)/etc/openzwave/
+includedir := $(PREFIX)/include/openzwave/
+docdir := $(PREFIX)/share/doc/openzwave-$(VERSION).$(VERSION_REV)
+else
 sysconfdir ?= $(PREFIX)/etc/openzwave/
```
Fishwaldo: why this? they are the same as the defaults....

binli71: In the yocto bitbake env, the support.mk being involved multiple times.
When the first time being involved, the "PREFIX" is empty and
"

sysconfdir ?= $(PREFIX)/etc/openzwave

"
will keep the value //etc/openzwave.

With the ":=", the build will be able to pick up the up to date PREFIX 
setting as the build infrastructure specified, and doing right install.

Regards
Bin

Fishwaldo: why can't PREFIX be specified during the first run?

You do realise that PREFIX is used as part of the source compilation, so if its not consistent, then things could break?

(for instance, we use PREFIX when compiling to specify the default location to search for the device database...)

binli71: On 16-04-27 02:08 AM, Justin Hammond wrote:
>
> In cpp/build/support.mk 
> <https://github.com/OpenZWave/open-zwave/pull/861#discussion_r61208238>:
>
> > @@ -96,9 +107,15 @@ else
> >  pkgconfigdir ?= $(shell pkg-config --variable pc_path pkg-config | awk '{split($$0,a,":"); print a[1]}')
> >  endif
> >
> > +ifeq ($(BITBAKE_ENV),1)
> > +sysconfdir := $(PREFIX)/etc/openzwave/
> > +includedir := $(PREFIX)/include/openzwave/
> > +docdir := $(PREFIX)/share/doc/openzwave-$(VERSION).$(VERSION_REV)
> > +else
> >  sysconfdir ?= $(PREFIX)/etc/openzwave/
>
> why can't PREFIX be specified during the first run?
>
Bin:
Bitbake used by yocto project is to build embedded linux for small 
devices. It typically does both host build first then doing cross 
compile for the target CPU after, then packaging the final image for the 
target device.
The bitbake used in Yocoto project doesn't allow user to specify the 
PREFIX freely. It is automatically invoking a lot of python / shell 
scripts doing host native / target cross compile build prcesses.
The recursive make happens all the time.
There is no way to enforce the final PREFIX value in the very beginning 
run. Because the PREFIX value for the final target path would be unknown 
during the very first run.
>
> You do realise that PREFIX is used as part of the source compilation, 
> so if its not consistent, then things could break?
>
> (for instance, we use PREFIX when compiling to specify the default 
> location to search for the device database...
>

Bin:
Fully understand the concern. This is why I introduced BITBAKE_ENV to 
make sure it is for bitbake build system only to search for default 
location, using compile tools and options from bitbake... all accordingly.

Regards
Bin

PR #864, opened 2016-04-21
Add 2gig CT50e thermostat with USNAP module RTZW-01

neudorf:
Using same code as ct30.xml and has worked well for CT50e so far in testing.


neudorf:
Apologies for the clumsy 3 commit. Second commit just reverts the first.

Fishwaldo:
Thanks!

PR #866, opened 2016-04-22
Fix CPPFLAGS in Makefiles

ffontaine:
Put $(CPPFLAGS) between double quote otherwise compilation will fail if CPPFLAGS
are defined by the buildsystem with more than one variable such as "-DX -DY"


Fishwaldo:
Already in Dev Branch... 
https://github.com/OpenZWave/open-zwave/tree/Dev

PR #870, opened 2016-04-24
Added Popp KFOB to Popp besides Z-Wave.Me

RaymondNg2:
My KFOB is only recognized when the KFOB line is added to the manufacturer Popp


nechry:
hello @RaymondNg2 do you got success with play sound form 0x26 CC ? I got this device but no way to the sound effect.

PR #871, opened 2016-04-25
Fix warning on wcsdup implicit declaration

ffontaine:
Define __GNU_SOURCE in cpp/hidapi/linux/hid.c (like already done in
cpp/hidapi/libusb/hid.c) to fix implicit declaration of wcsdup function


Fishwaldo:
Thanks!

PR #873, opened 2016-04-27
Fix libopenzwave.pc when using custom values

ffontaine:
sysconfdir and includedir can be overriden by the build system however those values weren't used when building libopenzwave.pc

Signed-off-by: Fabrice Fontaine <fabrice.fontaine@orange.com>


Fishwaldo:
Thanks!

Issue #883, opened 2016-05-14
_instance is not logged when reporting node messages


travellingkiwi:
When logging information about a node (Received traffic, meter updates etc) the _instance is not reported. This makes it extremely difficult to diagnose issues on multi instance devices

travellingkiwi:
Two choices... Do we want something quick & dirty (e.g. edit Meter.cpp to add _instance at various places, OR do we want to do something a little more involved?

A little more involved could be an extra parameter to Log::Write to add the instance. This would then be passed through all the way to LogImpl::GetNodeString and be formatted as "Nodexxx:xxx" in the log files.

I favor Nodexxx:xxx but not sure what this may break along the way (e.g. The logAnalyzer on the website). Happy to do the code... In fact I'll probably start on option 2 now in the hopes someone else agrees with me.

Fishwaldo:
Closing this to discuss in the PR

PR #884, opened 2016-05-15
Log _instance as well as node for switches and meters (Fix for issue #883)

travellingkiwi:
Add the _instance to the node when logging message regarding switches and meters. This ensures
that we can track which instance of a multi-instance device the message is for


Fishwaldo:
Hi,
So I see two issues here:
1) Yes - It will break the existing Log Analyzer. I can fix that, but it just takes time..
2) Technically, almost all CC's support MultiInstance, so this need to be universal change... If you can fix this up, I'll try to pull it into the Dev Branch once I got time to also get the Log Analyzer updated. (days/weeks? hopefully not months!)

travellingkiwi:
Sorry, what do you mean by universal change?

Fishwaldo:
All Command Classes, not just the ones you have done so far.... 

If you read though the source of each, you will see those documented that dont' support MulitInstance and can ignore them...

travellingkiwi:
Done. Plus Values so they log correctly as well. e.g.

```
016-05-20 08:01:08.335 Detail, Node010:001, Refreshed Value: old value=false, new value=false, type=bool
2016-05-20 08:01:08.335 Detail, Node010:001, Changes to this value are not verified
2016-05-20 08:01:08.335 Info, Node010:001, Received Meter report from node 10: Power=0.7W
2016-05-20 08:01:08.335 Detail, Node010:001, Refreshed Value: old value=0.8, new value=0.7, type=decimal
2016-05-20 08:01:08.335 Detail, Node010:001, Changes to this value are not verified

```

Fishwaldo:
Thanks... will merge soon once i update the Log Analyzer (which should be simple)

Fishwaldo:
Hi,

Thanks

PR #889, opened 2016-05-19
Remove endpoint mapping from the Qubino ZMNHBDx.xml defiition for the

travellingkiwi:
Qubino 2x relay switch
This change enables the qubino to be seen as 3 devices. The first instance
is the device itself (Master switch, power and energy). The relays are numbered
by OZW as instance 2 and instance 3 with associated switches, power & energy reporting

### DIFF
--- a/config/qubino/ZMNHBDx.xml
+++ b/config/qubino/ZMNHBDx.xml
@@ -109,7 +109,4 @@
 	<!-- Remove COMMAND_CLASS_BASIC -->
 	<CommandClass id="32" action="remove" />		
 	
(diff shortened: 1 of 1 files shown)


rofrantz:
Can we apply this fix also for the older version ZMNHBA2 ?

nechry:
yes is work for  ZMNHBA2 to. I already patch and validate with some jeedom users. there changes all include in next PR. the jeedom patch also include the refresh policy, otherwise the binary state don't report all time as we expect.

nechry:
Close with https://github.com/OpenZWave/open-zwave/pull/901

PR #890, opened 2016-05-23
Enabled support for Aeotec Illuminator G2 and Micro Switch 2nd edition

WoLpH:
The devices are not 100% identical to the `ses.xml` and `sei2.xml` but similar enough to function. Essentially the only difference is metering


nechry:
I recently completed all Aeotec. I fully rename it also with real product name and all parameters. There config will be merged very soon.

you can take a look her: https://github.com/jeedom/plugin-openzwave/commit/ff6849a88e2144fe44530400a2c74228e9251b58

WoLpH:
Ah, brilliant :)

In that case I'll close this pull request. Your patch looks much better

PR #893, opened 2016-05-25
Build project vs2013

PascalPiche:
Conversion to vs2013 project


Fishwaldo:
is there any problem with just letting VS upgrade the 2008 project to 2013?

it works for me in our CI enviroment.

Fishwaldo:
ping... closing if no response...

PR #897, opened 2016-05-29
Added support for Qubino flush shutter DC

WoLpH:


WoLpH:
Became a bit messy with automatic merges :)

Fishwaldo:
fyi we can squash commits when merging now, so not that much of a issue anymore. :)

WoLpH:
True, but it actually had a few other fixes in there as well which should not be merged yet :)

PR #909, opened 2016-06-11
Add devices Leviton DZPD3-1LW and DZPA1-1LW

benweb:


nechry:
There devices don't have any parameters or special rules for association groups? 

The added value for devices is theres attributes fully defined.

benweb:
These are the most basic on/off plugin switches provided by Leviton. AFAIK they shouldn't need any parameters.

PR #916, opened 2016-06-21
Use errno variable instead of _set_errno function, fixes MINGW build …

zgmnkv:
Use errno variable instead of _set_errno function, fixes MINGW build with old msvcr lib

### DIFF
--- a/cpp/src/platform/windows/LogImpl.cpp
+++ b/cpp/src/platform/windows/LogImpl.cpp
@@ -41,7 +41,11 @@ errno_t fopen_s(FILE** pFile, const char *filename, const char *mode)
 {
     if (!pFile)
     {
-        _set_errno(EINVAL);
+#ifdef __MINGW64__
+	_set_errno(EINVAL);
+#else
+        errno = EINVAL;
+#endif
         return EINVAL;
     }
 


cpp/src/platform/windows/LogImpl.cpp
```diff
@@ -41,7 +41,7 @@ errno_t fopen_s(FILE** pFile, const char *filename, const char *mode)
 {
     if (!pFile)
     {
-        _set_errno(EINVAL);
```
Fishwaldo: I'd prefer to use something like this http://lists.fossil-scm.org:8080/pipermail/fossil-users/2013-January/011654.html if you can figure out the correct defines for MINGW

zgmnkv: Thanks, I'll try this approach. By the way, what's the point of using _set_errno ?

zgmnkv: Seems that MINGW doesn't define msvcrt version

Fishwaldo: set_errno was introduced for platforms that need to do "something special" with the errno assignment (local threat variable handling etc). I'd just prefer to use that even though the posix standard says you can assign to directly to the variable.

> Seems that MINGW doesn't define msvcrt version

Hence why I said:

> if you can figure out the correct defines for MINGW

zgmnkv: I realized that _set_errno doesn't work only with classic mingw32, it works with mingw w64. I've updated code.
By the way, why do you use custom MINGW define instead of built in __MINGW32__ and __MINGW64__ ?

Fishwaldo: The Mingw updates were contributed by another user. So its what he used. :) Feel free to open another PR if you don't mind fixing that up.

Thanks

Fishwaldo:
btw, that fixes your mingw, but breaks it for standard visual C compiles... I'll fix it up though.

zgmnkv:
Thanks! I don't think this can affect MS Visual C compiler as it all is under #ifdef MINGW

PR #806, opened 2016-03-20
Device Support: Multisensor 6 config update

shanemadden:
Updated config parameters for Multisensor 6, using the [most recent reference material I've been able to find.](https://github.com/robertvandervoort/SmartThings/blob/master/Aeon%20Multisensor%206/Engineering%20Spec%20-%20Aeon%20Labs%20MultiSensor%206.pdf)

 - Param 3: 
   - Make it clear that this is for motion specifically, not any other kinds of wake-up.
   - Clarify the round-up behavior a little, hopefully.
 - Param 40:
   - Causes reporting only when thresholds are hit, as opposed to enabling threshold based reporting. Made that a little clearer.
 - Param 41-44:
   - Update units, sizes, and defaults.
 - Param 101-103:
   - Update help on the bitmask to make the numbers clearer.
 - Param 111-113:
   - Updated to accurately reflect the two different behaviors for these params based on the device's power source (and added a comment on how to tell from the application side which power source the device is on)

PR #920, opened 2016-06-30
Multisensor 6 Help Text Update

shanemadden:
Updates to the Aeotec Multisensor 6 configuration class text to correct and clarify - updated changes from https://github.com/OpenZWave/open-zwave/pull/806.


shanemadden:
(oh, and also a couple bits of whitespace cleanup in the file)

nechry:
Nice! thank's @shanemadden 
for param 42 and 44 I think the note 1 about unity is useless, the unity is already present in value attribute, but why not.

PR #927, opened 2016-07-08
Update dsb29.xml

ben423423n32j14e:
Resolved issue in Domoticz in which these sensors would become unresponsive after a period of time (usually about 12 hours). Tested with 3 sensors.

changed setasreport="true" to setasreport="false"

### DIFF
--- a/config/aeotec/dsb29.xml
+++ b/config/aeotec/dsb29.xml
@@ -35,7 +35,7 @@ https://aeotec.freshdesk.com/helpdesk/attachments/6009584513
 	</CommandClass>
 
 	<!-- Basic set as report -->
-	<CommandClass id="32" setasreport="true"/>
+	<CommandClass id="32" setasreport="false"/>
 
 	<!-- Association Groups -->
 	<CommandClass id="133">		


nechry:
can you give details about your issues, I have a old first version DSB04100 and the 2 lasted ZW120 and ZW112

ben423423n32j14e:
Mine is the "Door/Window Sensor (2nd Edition)" DSB29-ZWAU.

It took me a very long time to work out why the devices would work fine after including in the network and then stop after about 6 - 12 hours. The only way to get them working again was to either press the Z-Wave button on the back of the device which would make them work for a bit longer then stop again or exclude and re-include them in the network.

My setup is Razberry Z-Wave Sigma 500 series, Domoticz latest Beta with OpenZWave.

I am certain that this change is what fixed the issue in my system.

Cheers

ben423423n32j14e:
I discovered the solution researching here:
https://www.domoticz.com/forum/viewtopic.php?t=2463
The same fix for this sensor also worked for the DSB29-ZWAU.

nechry:
the other topic is about a AEON recessed door sensor (DSB54) and the setasreport still set to true in config file.
do you set the param 121 to 17 ?

ben423423n32j14e:
This is a screenshot from Domoticz, are you suggesting the 255 should be changed to 17? The screenshot is all default values:

screen shot 2016-07-08 at 6 17 46 pm

nechry:
yes the 255 fire only a basic set, with 17 you will got a binary sensor report + battery report.
if you don't want battery report just set to 16.

ben423423n32j14e:
ok I've set 17 and changed setasreport="true" in both the zwave network config and the DSB29.xml and restarted Domoticz. Let's leave it 48hrs and I'll report back with the result if the instability returns. Then we will know for sure.

Cheers

Fishwaldo:
It's pretty unlikely this is the fix, as it doesn't change anything of the device, it just changes the way OZW responds to Basic Set messages. 

@nechry's comment about changing from Basic Set messages to Binary Report messages is the correct "config" 

For background read about the Basic class In the wiki here and why it's not optimal.

ben423423n32j14e:
For some reason in my setup these modules stop working when setasreport="true".

I've changed it back to false on my system, if it doesn't cause any issues globally I'd suggest setting it this way in the master copy.

Fishwaldo:
setting that change will break even basic functionality of "unconfigured" devices. 

all setasreport effectively does is trigger OZW to query the status of the Binary CC. If thats hanging the device, then even manual refreshes etc will hang it.

PR #928, opened 2016-07-08
Small update and verify with version 2.5

pepeEL:


nechry:
@pepeEL he try to simplify configuration label. I think is better to use the help element to explain the the meaning of a parameter.

pepeEL:
What do you mean ?

Fishwaldo:
The label is too long. Try to use a short description, and put more information into the help text.

piokuc01:
I use it the same from manual fibaro.

pepeEL:
I use it as the same in manual fibaro

nechry:
yes but sometime the "official" label are too long and can be more precise with few words.

Fishwaldo:
Think about it from a GUI perspective. A long label can make GUI's look really bad. :)

Fishwaldo:
Hi,

Thanks

PR #930, opened 2016-07-08

saintman23:
Corrected XML config filename for ' type="0400" id="0109" '


nechry:
I'm not sur about this change, why you change de target definition but not the device name. they are 2 revisions of this device the new one 221 contains extra parameters.

saintman23:
The file I changed was fgs221.xml, which is defenitely for Double Swich FGS221 (2x1,5kW). Not for Single Switch FGS211 (1x3kW)

nechry:
we need also to rename the device name for a FGS-212

saintman23:
Renamed to FGS211

nechry:
is a old version?

saintman23:
type="0400" is an old one.

PR #938, opened 2016-07-22
Log::SetLoggingClass doesn't work if used before Manager::Create

tomq42:
With the previous code, this was the way the flow went.

I call Log::SetLoggingClass. Pointer to my class is stored in the static variable m_pImpl.
I call Options::Create
Options::Create calls Log::Create
Log::Create checks the singleton pointer, it's null. so it creates a new Log instance.
Log constructor notices that m_pImpl is non-null and uses what it finds there.

Manager::Create calls Log::Create.
Log::Create deletes the existing singleton log instance.
Log::~Log deletes the object pointed to by the static m_pImpl, then sets it to null.
Log::Create then creates a new Log object.
Log constructor notices that the static m_pImpl is null and hence creates a new implementation using the default class. My log implementation class is no longer used.

Clearly m_pImpl used to be a member variable of the Log class, and has been changed. All references to it use s_instance->m_pImpl for example, which is what you would have to do if it WAS a member variable. However, SetLoggingClass was introduced, it was made static, but issues remained.

This change attempts to fix it.
To avoid the m_pImpl object leaking at the end of the process a small deleter object is used. I did consider using std::auto_ptr, but it caused too many other changes.


tomq42:
Is there something else I need to do for this?
I realise I haven't made a changelog update.. Interestingly though I'm not sure I understand how that's supposed to work, in that the change log seems to generally have the form "Pull Request #xxx - Comment". Since I don't know the PR number until I've done my commit and created the PR, I clearly can't include such a comment in my original commit. So I commit, make a PR, commit amend my commit, then force push to update the PR?

Fishwaldo:
I'll do some testing. In my App its working ok, so let me see whats going on.

I also need to look into the locking now as well, You have moved from the s_instance->p_impl to direct p_impl, but the locking is still around the s_instance.

tomq42:
s_instance->m_pImpl was incorrect previously, because m_pimpl was a static member, so using s_instance was unnecessary. I've moved the location of the static, but otherwise done nothing to affect how it was used.
Clearly m_pimpl used to be a member, and at some point a "static" was added, but all usages remained the same, dereferencing s_instance. Maybe that change introduced some locking issues, don't know.

Fishwaldo:
Hi,

Thanks

tomq42:
That's somewhat inconvenient. It'll take me a while to figure out how to get all my changes across to a new fork.

tomq42:
Do you really mean that I need to create a new fork? I've already got one, so github won't let me create another. Do you just mean I need to rebase my branch?

Fishwaldo:
You can try a rebase... not sure if that will work... see this post I just sent to the list:
https://groups.google.com/forum/#!topic/openzwave/PsQYq6JmOaE

Sorry for the troubles... it really wasn't intended for this to happen.

PR #939, opened 2016-07-22
OPENZWAVE_DISABLE_EXCEPTIONS seems confused

tomq42:
The comment says:

Exceptions : define OPENZWAVE_DISABLE_EXCEPTIONS in compiler flags to enable exceptions instead of exit function

Define "DISable exceptions" to ENable exceptions.

And the implementation doesn't match anyway.
Exceptions are disabled by default, by virtue of a
#define OPENZWAVE_DISABLE_EXCEPTIONS

This change leaves the default behaviour such that exit() is used.
It adds a working switch to allow you to enable exceptions if you want.
You specify OPENZWAVE_ENABLE_EXCEPTIONS on the command line (ENABLE and not DISABLE).

### DIFF
--- a/cpp/src/Defs.h
+++ b/cpp/src/Defs.h
@@ -146,14 +146,13 @@ static inline int version_cmp(struct ozwversion a, struct ozwversion b)
 }
 
 #include "OZWException.h"
-#define OPENZWAVE_DISABLE_EXCEPTIONS
 #if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Already fixed in the Dev Branch.

tomq42:
But it IS a problem on master though. Can the fix not be applied to master?
As it is, I have to apply this fix manually.
If the response is "oh well, just use dev", then I'm not sure what the point of master is.

PR #943, opened 2016-07-25
Stop button for SwitchMultilevel

LeroyK:
Add a stop button to SwitchMultilevel. Useful for roller shutters to stop movement in any case (e.g. when changing its level in percentage).


.gitignore
```diff
@@ -37,4 +37,6 @@ OZWForm.vshost.exe.manifest
 .gitignore
```
Fishwaldo: Please drop this from the PR

cpp/src/value_classes/ValueButton.cpp
```diff
@@ -96,6 +96,17 @@ bool ValueButton::PressButton
 }
 
 //-----------------------------------------------------------------------------
+// <ValueButton::ResetButton>
```
Fishwaldo: hrm. This should also generate a ValueChanged Notification so a application can update its state.

Fishwaldo:
Please see my comments, also can you rebase this against the Dev Branch rather than master. (master is only for bug fixes).

Thanks

Fishwaldo:
Hi,

Thanks

PR #945, opened 2016-07-26
Add start/stop level change report event handling to SwitchMultilevel

ericyeargan:
This change allows reception of dim up/down events from dimmer controller paddles. Unfortunately, I also changed the message type handling from an if/else ladder to a switch which makes the diff difficult to parse. Let me know if I can help clarify.


cpp/src/command_classes/SwitchMultilevel.cpp
```diff
@@ -166,70 +168,162 @@ bool SwitchMultilevel::HandleMsg
...
-	}
 
-	if( SwitchMultilevelCmd_SupportedReport == (SwitchMultilevelCmd)_data[0] )
-	{
-		uint8 switchType1 = _data[1] & 0x1f;
-		uint8 switchType2 = _data[2] & 0x1f;
-		uint8 switchtype1label = switchType1;
-		uint8 switchtype2label = switchType2;
-		if (switchtype1label > 7) /* size of c_switchLabelsPos, c_switchLabelsNeg */
+		case SwitchMultilevelCmd_Set:
```
Fishwaldo: Why did you implement Set? OZW is not a Device. We dont expose the SwitchMultiLevel CC in our NIF so this code will never be executed/Called.

Please Delete this portion.

Fishwaldo:
Please see my comments, and drop the Set Handling. Thats "dead" code.

cpp/src/command_classes/SwitchMultilevel.cpp
ericyeargan: I believe I added that while trying to get the bright/dim received events to work. Will remove...

Fishwaldo:
Hi,

Thanks

ericyeargan:
@Fishwaldo, no problem - new PR submitted...

PR #957, opened 2016-08-09
Adding support for Trane XR524, A.k.a. TZEMT524AA21MA

jusovsky:
Adding support for new thermostat.  This has only been lightly tested, and I don't have all values mapped, but the important ones are good to go.


nechry:
we also need the manufacturer_specific.xml ids definition
Please ref to https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices

Fishwaldo:
Hi,

Thanks

PR #958, opened 2016-08-11
CommandClass registration was hard to read and error prone

tomq42:
This change uses the X macro technique to ensure that the command class registration
doesn't have any typos or copy and paste errors.
It also makes it simpler to read.


Fishwaldo:
Hi,

Thanks

PR #971, opened 2016-08-19
Add start/stop level change report event handling to SwitchMultilevel

ericyeargan:
(Re-submission of previous pull-request)

This change allows reception of dim up/down events from dimmer controller paddles. Unfortunately, I also changed the message type handling from an if/else ladder to a switch which makes the diff difficult to parse. Let me know if I can help clarify.

This request also removes the dead set handler code from the previous request.


Fishwaldo:
Closing

PR #992, opened 2016-09-27
Devolo Metering Plug and Permundo Plugs

tobiasgraf:
Added Devolo Metering Plug .
Also added Permundo PSC132ZW Smart Mini Plug and PSC234ZW Smart Plug to manufacturer_specific.xml

I have some more devolo devices i'm going to add, would you prefere to have one PR for each device or one for all?


nechry:
I will check very soon your PR. Sometimes is better to have one by device but many in a same PR still good

nechry:
@tobiasgraf just a small thing, next time do not include the distfiles.mk file. we prefer manually add this entry with the change log by herself.

thank 's again for the PR

PR #1005, opened 2016-10-11
Fixed wrong sign on the Value_System_Config_Mode comparasion

minorai:


minorai:
Sorry about no description. First try at pull request :) This commit will fix the bug that does not allow DoorLock command class to send System Config Mode messages with ID over 0x02

Fishwaldo:
Good Catch. Thanks

PR #1006, opened 2016-10-11
Unknown: id=0176 Unknown: type=0005, id=0001

ic3cool:
modified:   config/manufacturer_specific.xml
	new file:   config/telldus/tzdw100.xml


ic3cool:
something like that?

PR #1009, opened 2016-10-13
Make the ValueID default constructor public

dhozac:
This is required by python-openzwave when it creates a Map with
ValueIDs as the value.


Fishwaldo:
This is not correct. Python-openzwave is doing it wrong if it needs to construct a ValueID

PR #1014, opened 2016-10-25
Fix for Security handling

tobiasgraf:
The ZME Wall Controller Wall C-S sends the SecurityCmd_MessageEncapNonceGet (0xC1). When this can not be decrypted (for example the `MAC Authentication of Packet Failed. Dropping` Error occured), the expected nonce-report does not get sent to the device. 
So, this now also sends the nonce-report on decryption fail.
Also fixes a bug that prevent the SecurityCmd_MessageEncapNonceGet to get recognized after the decryption.
It checked for _data[6], but _data[6] was already overwitten by the decrypted package.
This caused the Wall Controller to run into its timeout and wait for 20 seconds until another frame was sent.


Fishwaldo:
Looks Good. I'll do a quick test tonight and merge if no issues..

Fishwaldo:
Thanks!

Issue #533, opened 2015-04-26
Additonal type for Remotec ZXT-120


rayslinky:
manufacturer_specific.xml lists the Remotec ZXT-120 as type="0101" id="8377". I recently received one that is reporting type="0100" id="8377". I manually added it to my manufacturer_specific.xml to mimic the 0101 and it works as expected.

killercode:
My remotec ZXT-120 also has different bitmask configuration, took me a long to understand how to configure...

rayslinky:
You're right -- they are a bit different. :/

Fishwaldo:
Then we would need a new config file for the updated device in that case.

killercode:
Yeah.... but the problem is that it seems that is a fake chinese remotec ZXT-120... Which means that it has the same manufacturer id and device id with different settings.... I don't know how this can be solved...

Fishwaldo:
Then I'm not going to update for "Fakes"... :)

antoniolobefaro:
Hi guys, 
I'm new about zwave and remotec but I need to use these in a project thtat is about to end . I'm able to see the remotec in zwave network using DomotiGa (I see the temperature releved) but how can I set the THERMOSTAT_MODE or THERMOSTAT_SETPOINT or THERMOSTAT_FAN_MODE? 
Do you have examples?

PR #1023, opened 2016-10-28
Add product type for ZXT-120US

daradib:


daradib:
fixes #533. I think the fakes are a different issue - I didn't have any bitmask problems.

PR #1034, opened 2016-11-07
Fix FreeBSD soname

stromnet:
Hi,
a previous FreeBSD related change lost the -soname linker flag. This patch re-introduces it.

An unrelated question: What is the current version number / release plan?
The changelog has a V1.5 title which seems to get continually updated. But there is a V1.5 git tag pointing to 2 commits after V1.4 (a lot of commits ago).

I'm working on the FreeBSD package, it would be good to know what's to expect :)

Thanks!

### DIFF
--- a/cpp/build/Makefile
+++ b/cpp/build/Makefile
@@ -39,7 +39,7 @@ TARCH	+= -arch i386 -arch x86_64
 LDFLAGS += -dynamiclib -install_name "$(DESTDIR)/$(instlibdir)/$(SHARED_LIB_NAME)"
 LIBS	+= -framework IOKit -framework CoreFoundation -arch i386 -arch x86_64
 else ifeq ($(UNAME),FreeBSD)
-LDFLAGS+= -shared -lusb
+LDFLAGS+= -shared -lusb -Wl,-soname,libopenzwave.so.$(VERSION)
 
 # Pre FreeBSD 10.2 we have no native, or "old" native iconv.h (non-posix compliant
 # const modifiers)


Fishwaldo:
master branch is like a rolling release of stable 1.4 code. Current version there is 1.4.2237.
Dev branch is unstable and version 1.5. 

If your packaging, you should do it off the master branch. (bugfixes/config file updates into master branch - enhancements/breaking changes etc into Dev branch)

stromnet:
Ah, I see. The V1.5 tag is nothing to rely on then.

From looking the code (git describe --long --tags), I guess the 1.4.2237 version number comes from the number of git revs?

Fishwaldo:
yeah the revision is roughly the number of commits since the last tag on that branch... so when we push 1.6 (1.5 is a dev number) and tag it, master should start from revision 1.6.0 again.

stromnet:
Ok, perfect! I've submitted 1.4.2238 to the FreeBSD ports: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=206526#c21

There where some ideas/input from the port maintainers about some things:
* -Werror should not be in release mode
* CFLAGS/LDFLAGS should honour user overrides (prepended to, not appended to)
* To simplify debugging builds, don't use @ in makefiles, at least not statically. Perhaps instead `AT:="@" ... $(AT)something`: http://stackoverflow.com/questions/3861911/ignoring-at-symbol-in-makefiles

I didn't make any efforts on those for now though.

PR #1050, opened 2016-11-25
Fix null pointer

tobiasgraf:
It was possible for `m_currentControllerCommand` to be NULL, thus accessing it causes a segmentation fault. Moving the log output inside the if where this is checked.


Fishwaldo:
I don't see how it could be NULL, but accept your change anyway.

Issue #997, opened 2016-10-04
Update UserCode command class to Support ASCII and deleting UserCodes


dicarlo2:
`SetValue` only supports settings user codes or refreshing user codes[1]. I'd be happy to try adding the functionality, but I'd need some pointers on the "right" way to do it. Ideally we'd be able to set a given index with any of the `UserCodeStatus` enum values.

[1] https://github.com/OpenZWave/open-zwave/blob/master/cpp/src/command_classes/UserCode.cpp#L303

nechry:
I do this task directly in my gateway by setting only 0x00 in value. I'm not sure this feature must be present in the library.

Fishwaldo:
the UserCode isn't 100% compliant with the Specs. For clearing a usercode, we should be setting the status to available, and sending 4 NULL bytes. So this does need work.....

But

We know also know that usercode is ASCII based, so this should be converted from a RAW to string ValueID. Thats going to break a few applications, so I'd prefer to postpone that to after 1.6 is released. 

(if you want to start refactoring the UserCode CC with the above points, thats fine, please base on the Dev branch).

techgaun:
@dicarlo2 have you pursued this at all? I was wondering if I could jump in and try contributing on fixing this.
@nechry could you explain little bit what you mean by setting only 0x00 in value? Do you mean sending the raw packets to the serial device?
@Fishwaldo If @dicarlo2 is not pursuing this, I could look into it. I looked at the Dev branch and didn't see this issue done.

nechry:
something like this:
`
2017-06-08 20:44:49.704 Info, Node200, Value::Set - COMMAND_CLASS_USER_CODE - Code 1: - 1 - 1 - 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00
2017-06-08 20:44:49.704 Detail, 
2017-06-08 20:44:49.704 Detail, Node200, Queuing (WakeUp) UserCodeCmd_Set (Node=200): 0x01, 0x15, 0x00, 0x13, 0xc8, 0x0e, 0x63, 0x01, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x25, 0x5e, 0x26
`
it's work for benext rfid keypad, the thing is the keypad have only 1,2,3 and 4 digit keys. if your device have the 0 digit ,10 zero will unlock so it's not a good way... 

[...]

PR #1059, opened 2016-12-03
allow clearing of user codes when user passes [0,0,0,0]

drytoastman:
I think most zwave locks allow removing/clearing codes from their memory.  According to the zwave spec file "SDS12652-13 - Z-Wave Command Class Specification N-Z.pdf" on page 333, you can set the status to Available and pass 4 bytes of 0 data.

I've tested this with my old Schlage BE369 deadbolt and it appears to work as intended.


drytoastman:
Note, this is raw zero data, not the **ascii** characters '0000' or [0x30, 0x30, 0x30, 0x30] for setting the door code of '0000'.

Fishwaldo:
Hrm. Let me think about this. using '0000' implies that the user is knowledgeable about Z-Wave. Id like something intuitive like a button etc.

drytoastman:
The primary reason I picked the '\0\0\0\0' was because I'm coming through the python API Value object so my only real options are setting the value or something through the node object but that doesn't seem right.  Can you assign a zero length string to a value?  That might be more understandable for the average user.

Fishwaldo:
Based on this posting: http://forum.z-wavepublic.com/t/3-51-user-code-command-class-version-1/80 we now know that the RFID'd devices were breaking the specs. The usercode should be a ASCII string. 

So I'm inclined to make the following changes:
1) Add a ValueString containing the ASCII representation of the usercode and some code to see if the actual usercode falls in the ASCII range. If not, then dont expose it. 
2) Keep the Existing ValueRaw for applications to continue to work with RFID's or existing ASCII based devices.
3) Expose a new ValueInt that is write only. Takes a index of the UserCode to clear. When submitted the usercode class will do the necessary
4) adopt your "0000" code to clear as well.

The only constraint, we only have 256 indexes to work with on ValueID's. I think its time to expand that to 65536 :)

nechry:
I agree with your proposition @Fishwaldo

nalipaz:
This issue is now going on a year standing, is it possible there will be any movement on this change in the foreseeable future?

zmcandee:
@Fishwaldo Any progress on incorporating this? The PR works as described and seems to be an important update for people that want to clear codes.  Could we merge this as is and create a separate issue for adding ValueInt and ValueString?

nalipaz:
Looking further around the threads it seems that 0000 only clears on certain devices and on others it is a working code when set. That said I don't think this should necessarily be incorporated as it has unintended consequences for those who have a device that doesn't clear upon setting 0000.

drytoastman:
I was the original poster.  That patch was really a kludge to work within the current API.  It probably makes more sense to add a ClearValue or such to the API but I don't know enough about the official zwave command classes to know if that is correct or not.  Unfortunately, it seems like lock manufacturers are doing different things for clearing so it may end up being some piece that needs custom handling based on the lock type.

Fishwaldo:
I'm Closing this as I'm refactoring the UserCode CC right now. See #997

PR #1064, opened 2016-12-08
fix to Nonce generation

PriMachVisSys:
Nonce handling has a bug.  If one is generated with the same first byte as an earlier in the cache (ie. m_nonces[i][0] == m_nonces[j][0] where i<j) and is used to encrypt a packet, then GetNonceKey() will pick the first (i) for decryption.  This causes the decryption to generate garbage, the MAC comparison to fail, and the packet to be dropped.

The fix is to require the first byte of each nonce be unique during generation.  This is actually required by the security protocol.

There are two other tweaks in this patch.  First, it also requires the first byte be non-zero.  This requirement is coming from a long-time ZWave developer; I haven't been able to find it documented anywhere.  (To not require this use the equation for the other bytes for the first, ie. 256*rand()/(RAND_MAX+1.0).)  Second, this uses the high-order bits from rand(), which are supposed to better.

This change has been stress-tested locally.

### DIFF
--- a/cpp/src/Node.cpp
+++ b/cpp/src/Node.cpp
@@ -3427,9 +3427,31 @@ Node::DeviceClass* Node::GenericDeviceClass::GetSpecificDeviceClass
 //-----------------------------------------------------------------------------
 uint8 *Node::GenerateNonceKey() {
 	uint8 idx = this->m_lastnonce;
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Thanks

PR #1066, opened 2016-12-15
Update ValueList.cpp

jotakar:
Lists doesn't work if value is different of index, so if mode[1] = 1 it works, but if mode[1] = 3 then it doesn't work. It confuses index with value.
In this function we see
 tempValue->m_valueIdx = GetItemIdxByValue(_value); 
but _value (argument of the function) is not a value, it's an index, see setByLabel function where this function is called with an index as param, not a value.
Perhaps should the name be changed?


Fishwaldo:
I reverted the offending commit 7b7c10e

PR #1068, opened 2016-12-15
Add configuration for HomeSeer HS-WD100+ Wall Dimmer

sterling:


nechry:
You sure about size of parameter 10?

nechry:
A param 7 and 8 are also supposed to exists.
The 10 is a byte and accept 1 to 255, not a list.

sterling:
I referenced http://products.z-wavealliance.org/products/1730/configs for the parameters. Using this configuration appears to work locally.

nechry:
Ok if you have validate a change with the parameters 10 and the device accept it, is the answer.

nechry:
Thanks @sterling

PR #1074, opened 2016-12-22
replace fixed value initialization vector (for encryption key) 

PriMachVisSys:
Code filled the sender portion of the intialization vector with 0xaa (with a random seed commented out).  It needs to be random bytes.  Because the IV is used twice and overwritten in-between by the aes_mode_reset() call, we need to create a copy for generating the MAC.

There's also a spelling typo in an unrelated comment that got fixed.


Fishwaldo:
looks fine. I was using 0xAA during debugging. I guess its stable enough to move back to Random NONCE.

PR #1078, opened 2016-12-31
Fix ThermostatSetpoint Command Class detecting wrong setpoint types

atis380:
In response to a ThermostatSetpointCmd_SupportedGet, the ThermostatSetpoint command class considers each bit number is a setpoint type Id. However, the ThermostatSetpoint specification makes a distinction between both. This makes OpenZwave reporting wrong setpoint types (eg "Moist Air" instead of "Away heating"), so I added a mapping between bits and their corresponding setpoint types.

I also removed the ThermostatMode command class in the Horstmann/scsc17.xml config file because the thermostat does not handle it (it doesn't report it in its NIF, and doesn't mention it in its documentation either). This led to a unusable Mode value being shown in applications. However, the thermostat can _control_ a device using this command class.

Also renamed "Heating 1" and "Cooling 1" setpoint types into "Heating" and "Cooling" to get closer to the command class specification and because I don't see the reason of this "1". Of course, if I missed a point, let me know.

### DIFF
--- a/config/horstmann/scsc17.xml
+++ b/config/horstmann/scsc17.xml
@@ -1,14 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- SCSC17 (Secure / Horstmann) -->
 <Product xmlns='http://code.google.com/p/open-zwave/'>
-  <!-- The thermostat does not properly report its operating modes -->
-  <CommandClass id="64" getsupported="false">
-    <SupportedModes>
-      <Mode index="0" label="Off"/>
-      <Mode index="1" label="Heat"/>
-    </SupportedModes>
-  </CommandClass>
-  <!-- This thermostat's setpoint descriptions are 0 based, not 1 -->
+
   <CommandClass id="67" base="0" override_precision="0"/>
 
   <!-- Configuration Parameters -->
@@ -26,7 +19,7 @@
     </Value>
     <Value type="byte" instance="1" index="4" genre="config" label="Delta T: Temperature change" min="1" max="50" value="5">
       <Help>Change between temperatures to trigger a temperature report (step 0.1)
-        1 to 50, 1 or 0.1 (Default 5 = 0,5C) -> (1C = 10)
+        1 to 50 (Default 5 = 0,5C)
       </Help>
     </Value>
   </CommandClass>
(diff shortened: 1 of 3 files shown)


atis380:
Considering the specifications and this code change, the `base="0"` setting should apply to all ThermostatSetpoint command classes. I now doubt that some thermostat really need a `base="1"` setting in their ThermostatSetpoint command class configuration directive.

From the existing thermostat config files, only Danfoss living defines its base at 1 explicitly (other Danfoss thermostat don't), and the Qubino LDX and IDxS1 which let it to its default value (here again other Qubino modules defines it as 0). I don't own these devices so I can't check but if someone could send me a log of them, I would be pleased to check them and fix if needed.

I can test the Qubino ZMNHID1 H1S2P1.

atis380:
Tested the Qubino ZMNHID1 H1S2P1.

Added the param 59 (set operating mode to "heat" or "cool") to the ZMNHIDx H1S2P1 (it may also apply to H1S1P1, by the way but don't have access to it to check it).

It then appears that the thermostat reports different supported setpoints (and different supported thermostat modes) based on its configuration :

- if param 59 is set to "Heat", the thermostat reports the "Off" and "Heat" modes, and the "Heating" setpoint
- if param 59 is set to "Cool", the thermostat reports the "Off" and "Cool" modes and the "Cooling" setpoint

As a consequence, when param 59 is changed, the module has to be rescanned for supported thermostat modes and for supported thermostat setpoints.

Except this tricky situation, I found that the thermostat correctly reports its supported setpoints and thermostat modes, so I updated the config file.

atis380:
Hi guys,
Do you have any question regarding this PR ? A comment maybe ?
When do you expect to be able to review it ?
Regards

Fishwaldo:
Hi,
* Please seperate out config file change from code changes....
* We should leave the Heating 1 and Cooling 1 as they are - Many applications use string matching etc and this would break applications that do.

As noted on the mailing list, we need to support both implementations, so a config file option for specifying what implementation the device is using needs to be made - Or some way to disover this during the initial inclusion.

Fishwaldo:
ping....

atis380:
Hi,
Let me close this PR so I can create 2 separating config files from code changes.
Will also add a config file option.

Issue #1058, opened 2016-12-03
configurations HELP text gets wrapped to one line


gizmocuz:
If the original configuration file includes a help text that is spanned across multiple lines, and a node is included, the resulted configuration file from the user (zwcfg_xxxx.xml) will have a help text that is one line.

For instance, the NodeOn msp31xxMicroSmartPlug.xml

Original:
```
        <Value type="byte" index="3" genre="config" label="Follow State" size="1" min="0" max="3" value="3">
            <Help>
                This parameter allows Basic Set Commands to be sent when the channel output state changes (ON or OFF). Only the two least significant bits can be addressed.
                Bit 0: Enable or disable Basic Set Commands sent to associated nodes of Group 2, when Channel Output State changes.
                Bit 1: Enable or disable Basic Set Commands sent to associated nodes of Group 3, when Channel Output State changes.
            </Help>
        </Value>
```

After inclusion and the configuration file is written to disk:

```
               <Value type="byte" genre="config" instance="1" index="3" label="Follow State" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="3" value="3">
                    <Help>This parameter allows Basic Set Commands to be sent when the channel output state changes (ON or OFF). Only the two least significant bits can be addressed. Bit 0: Enable or disable Basic Set Commands sent to associated nodes of Group 2, when Channel Output State changes. Bit 1: Enable or disable Basic Set Commands sent to associated nodes of Group 3, when Channel Output State changes.</Help>
                </Value>
```

This results in a not so clean help text.

possible solutions:

**# 1 (not tested)**
place it in CDATA tags:
```
<![CDATA[ 
first line
second line
last line
]]>
```

**# 2 Add "\n" behind each line**
first line\n
second line\n
last line\n

Maybe in addition to this, do we need the help text in the final zwcfg_xxxx.xml file, or can we somehow retrieve it via the library out of the original file ? (Or maybe this help text can be created internally at startup ?)

Fishwaldo:
yep. Known issue. CDATA is the prefered method, as using a \n in the lines would "break" HTML widgets etc. Its on my plans for when I fix up the help text for builtin values. #327

PR #1079, opened 2017-01-02
Fix typo's in Fibaro FGS-213 & FGS-223 config files

roytomeij:
As the title says, this fixes typo's and faulty search/replaces in the FGS-213 and FGS-223 config files.


nechry:
Usually description with only one line I don't add the dot at the end. Thanks for others catch

roytomeij:
I added theme here for consistency @nechry because most lines had them already. Will keep this in mind for next time!

nechry:
Yes it's not a problem. And it's not a official rule. We must found a way to display help description nicely, in future, a issue is open in regard. See:https://github.com/OpenZWave/open-zwave/issues/1058

PR #1097, opened 2017-01-14
Adding Zooz ZSE40 4-in-1 sensor

djschaap:


nechry:
please use list type for enumerable items

djschaap:
Updated to use list type for 1, 6, & 7.

Issue #986, opened 2016-09-04
OpenZWave Mutex Lock


Ryszard-Trojnacki:
I have a problem with hardware:
https://groups.google.com/forum/#!topic/openzwave/tgtXfSude5I
and for workaround I'm reloading driver (Manager::RemoveDriver, Manager::AddDriver) but unfortunately sometimes library (OpenZWave) hangs on a mutex.

I'm doing Remove, Sleep, Add, Sleep:

``` java
                        synchronized(lock) {
                            try {
                                Manager.removeDriver(zwaveDevice);
                            }catch(Exception e) {
                                return;
                            }
                            try {
                                Thread.sleep(1000);
                            } catch (InterruptedException e) { return; }
                            try {
                                Manager.addDriver(zwaveDevice);
                            }catch(Exception e) {
                                return;
                            }
                            log.info("Driver added");
                            try {
                                Thread.sleep(3000);
                            } catch (InterruptedException e) { return; }
                        }

```

Synchronized block is for blocking calls to Manager.
I'm sleeping 3000 ms to wait for reload and after that my polling thread is calling Manager::RefreshValue and it is locked.

My Stacktrace (from Java - jstack) is:

```
----------------- 8917 -----------------
0x00007f7bf8c76b5d      __lll_lock_wait + 0x1d
0x00007f7ba6acdb9f      _ZN9OpenZWave5Mutex4LockEb + 0x27
0x00007f7ba6aa64d7      _ZN9OpenZWave9LockGuardC1EPNS_5MutexE + 0x2f
0x00007f7ba6aa0fc2      _ZN9OpenZWave7Manager12RefreshValueERKNS_7ValueIDE + 0x68
0x00007f7ba6a746cd      Java_pl_rtprog_paigo_zwave_Manager_refreshValue + 0xb4
0x00007f7be9f3ad7f      <Unknown compiled code>
```

And GDB backtrace for this thread is:

```
Thread 36 (Thread 0x7f7ba72b5700 (LWP 8917)):
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007f7bf8c70d82 in __GI___pthread_mutex_lock (mutex=0x7f7bb8007f18) at ../nptl/pthread_mutex_lock.c:115
[...]

PR #1105, opened 2017-01-23
possible fix (or workaround) for a deadlock

tobiasgraf:
My OpenZwave stopped responding multiple times a day. And after a certain time i was seeing alot of `2016-12-21 06:21:20.520 Error, ERROR: Not enough space in stream buffer`.  
 Looking into this I found out that the driverThread was hanging while trying to get a lock for m_nodeMutex.
```
2017-01-20 10:54:57.871 Warning, CheckCompletedNodeQueries m_allNodesQueried=0 m_awakeNodesQueried=1
2017-01-20 10:54:57.874 Always,       Locking 827319692 from void OpenZWave::Driver::CheckCompletedNodeQueries()
2017-01-20 10:54:57.876 Always,       Locking 818914372 from bool OpenZWave::Manager::IsNodeFailed(uint32, uint8)
2017-01-20 10:54:57.878 Always,       Locking 827319692 from void OpenZWave::Driver::CheckCompletedNodeQueries() ... done with ret: true
2017-01-20 10:54:57.882 Always,       Unlocking 827319692
2017-01-20 10:54:57.884 Always,       Locking 818914372 from bool OpenZWave::Manager::IsNodeFailed(uint32, uint8) ... done with ret: true
2017-01-20 10:54:57.891 Always,       LockGuard destructor 818914372, isSignaled? false
2017-01-20 10:54:57.898 Always,       Locking 818914264 from bool OpenZWave::Manager::RefreshValue(const OpenZWave::ValueID&)
2017-01-20 10:54:57.900 Always,       Locking 818914264 from bool OpenZWave::Manager::RefreshValue(const OpenZWave::ValueID&) ... done with ret: true
2017-01-20 10:54:57.904 Warning, CheckCompletedNodeQueries all=0, deadFound=1 sleepingOnly=1
2017-01-20 10:54:57.906 Always,       LockGuard destructor 827319692, isSignaled? false
2017-01-20 10:54:57.909 Always,       LockGuard destructor 818914264, isSignaled? false
2017-01-20 10:54:57.925 Error, MutexImpl::UnLock failed with error: 2 (1)
```
In this log you see that 827319692 is trying to unlock twice, once with a call to the LockGuard::Unlock() and once via the destructor. But in the meantime another thread (818914264) got the lock and also tries to unlock the mutex. So now 1 unlock fails and as a result the m_lockcount is now -1. The next lock increases this again to 0, but as isSignaled() returns true when m_lockcount is 0 the destructor of the LockGuard does not unlock this, and thus the mutex is locked forever.
To work around this I removed all manual calls to the LockGuard::Unlock() method and instead made a scope around them, now no LockGuard should try to unlock twice.
I did not look at the MutexImplementations for windows and winRT.
Does this make any sense?

What i am not sure, in case the unlocking fails, would it help to increase the m_lockCount again?

perhaps #986 and https://groups.google.com/forum/#!topic/openzwave/tgtXfSude5I could be related to this?

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -1526,27 +1526,28 @@ void Driver::CheckCompletedNodeQueries
 		bool sleepingOnly = true;
 		bool deadFound = false;
 
-		LockGuard LG(m_nodeMutex);
-		for( int i=0; i<256; ++i )
 		{
-			if( m_nodes[i] )
+			LockGuard LG(m_nodeMutex);
+			for( int i=0; i<256; ++i )
 			{
-				if( m_nodes[i]->GetCurrentQueryStage() != Node::QueryStage_Complete )
+				if( m_nodes[i] )
 				{
-					if ( !m_nodes[i]->IsNodeAlive() )
+					if ( m_nodes[i]->GetCurrentQueryStage() != Node::QueryStage_Complete )
 					{
-						deadFound = true;
-						continue;
-					}
-					all = false;
-					if( m_nodes[i]->IsListeningDevice() )
-					{
-						sleepingOnly = false;
+						if( !m_nodes[i]->IsNodeAlive() )
+						{
+							deadFound = true;
+							continue;
+						}
+						all = false;
+						if( m_nodes[i]->IsListeningDevice() )
+						{
+							sleepingOnly = false;
+						}
 					}
 				}
 			}
 		}
-		LG.Unlock();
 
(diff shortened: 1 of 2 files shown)


nechry:
I also have this issue from time to time (once a month, around) in my network, we (Justin and me), previously, suspected a threading issue with background worker. But I remove all of them,  and I still got this issue.

Fishwaldo:
Ok. Looks reasonable. I will probably add some additional code to check for locks that are going negative and alert...

Issue #1108, opened 2017-01-25
Allow to discover Controller Command that triggered Notification


gdombiak:
When a Notification of type Notification::Type_ControllerCommand is received, it would be helpful to be able to obtain the command that triggered the Notification.

See: https://groups.google.com/forum/#!topic/openzwave/HG-DZsBKLIk

Fishwaldo:
Please track this in the #1117 PR.

PR #1117, opened 2017-01-30
Add getCommand to Notification Class

tobiasgraf:
This is usefull in the ControllerCommand notification to the command that triggered the Notification.
see https://github.com/OpenZWave/open-zwave/issues/1108

also added the ControllerCommand to the Notification::GetAsString()

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -5531,6 +5531,7 @@ void Driver::UpdateControllerState( ControllerState const _state, ControllerErro
 		}
 		Notification* notification = new Notification( Notification::Type_ControllerCommand );
 		notification->SetHomeAndNodeIds(m_homeId, 0);
+		notification->SetCommand(m_currentControllerCommand->m_controllerCommand);
 		notification->SetEvent(_state);
 
 		if( _error != ControllerError_None )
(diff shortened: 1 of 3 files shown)


gdombiak:
Hi Tobias,

Thanks for creating the PR for this enhancement. Will the change be enough so that listeners can distinguish, for instance, if the event relates to "add node (insecure)" vs "add node (secure)"? I'm not very familiarized with the code of this library, from a quick look it seems to be using the same command class (number) so wonder if we can distinguish between both modes of inclusion.

Thanks,
Gaston

tobiasgraf:
Hi Gaston,

no there is only one `ControllerCommand_AddDevice` for both cases. I'm not sure if this can easily be done. 

regards

Tobias

gdombiak:
Hi Tobias,

Agreed. Refactoring commands is way too much for this. It seems like 'm_controllerCommandArg' is being used for indicating if the inclusion mode is secure or not. Do you think that including 'm_controllerCommandArg' in the event is a bad move? I do not know what kind of info may appear there and if that is exposing too much the internals.

Thanks,
Gaston

Fishwaldo:
Could you rebase this against the Dev branch instead? I like to only introduce changes that don't change API's on master branch.

tobiasgraf:
Rebasing brings in other stuff too.
I'll open a new PR on dev branch.

Fishwaldo:
Yeah. I need to Sync the branches. On my TODO for tomorrow

PR #993, opened 2016-09-27
improve support for CentralScene Command Class

tobiasgraf:
Hi,

I implemented the CentralScene according to the new releases Comand Class spec.
It's not complete but works for the devices I have (nodon octan remote).

The spec tells us 
"A receiving node MUST interpret the absence of a Key Held Down event for 400ms as the indication of a
Key Released event." but this i have done in my application. Would it be better to have it somewehe in openZwave?

I used the scene number as instance because i highly doubt someone would ever use multichannel in combination with centralScene.

PR #1124, opened 2017-02-09
Continuation of Improve support for CentralScene Command Class

robbiet480:
Continuing PR #993


nechry:
@robbiet480 do you already take a look this jeedom implementation in beta branch?
https://github.com/jeedom/plugin-openzwave/blob/beta/resources/python-openzwave/openzwave/cpp/src/command_classes/CentralScene.cpp

robbiet480:
@nechry Yup, but hadn't taken a close look at the differences between the jeedom branch and @tobiasgraf's code. Do you think yours is a better base to work from?

nechry:
We expose more possibilities, like "Scene Number" see comment of @stromnet 
https://github.com/OpenZWave/open-zwave/pull/993#issuecomment-274277201

I don't want said the @tobiasgraf's work is not good, but we must consider to keep all good parts (implementation) for both side.

robbiet480:
@nechry I'll go ahead and re-stage this PR with your code as a base.

PR #1126, opened 2017-02-09
Makefile: use SHARED_LIB_NAME variable in LDFLAGS

stintel:
When not on Darwin, SHARED_LIB_NAME is libopenzwave.so.$(VERSION). Use
this variable in LDFLAGS on FreeBSD and Linux, instead of duplicating
its value.

Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>


stintel:
Thanks!

PR #1127, opened 2017-02-09
unix: fix compilation against musl libc

stintel:
POSIX.1-2001 requires sys/select.h for select() and friends.
Compile-tested on glibc and musl, runtime tested on musl.

Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>


Fishwaldo:
Thanks

stintel:
Thanks!

PR #1130, opened 2017-02-12
Fix corruption hidden by bad cast. Destination is 8 bits, but 32 bits…

ash7777:
… are changed on Windows.


Fishwaldo:
Seems reasonable, but I see another bug in the code. Pushing a fix for that now.

PR #1131, opened 2017-02-13
Added support for POPP 009402

geirra:
Support for Popp: 10 Year Smoke Detector and Siren

Requires PCB 1.5 and firmware 1.14 (or higher)


nechry:
@geirra please always do a make xmltest before submit your PR. your file is bad formated

Issue #1143, opened 2017-02-19
ForceInstances removal not implemented


martian:
MultiInstanceAssociation::Remove() does not look at ForceInstances, while ::Set() does.

This causes a problem, when dealing with a ForceInstances device,  when a node is added to an association, which is rewritten to a .0 instance in MultiInstanceAssociation::Set().

This node can not be removed anymore (as tested on Fibaro FGS-223 with either domoticz or owz-control-panel), because  MultiInstanceAssociation::Remove() does not rewrite the node to an instance.

Fishwaldo:
please see comments on #1144

PR #1144, opened 2017-02-19
ForceInstances, implement remove()

martian:
Fix for #1143.

Implement ForceInstances in MultiInstanceAssociation::Remove().

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+- Apply ForceInstances patch to association removal too.
 Version 1.5
  - Pull Request #743 - Updated Fibaro Devices from Navstev0 (Justin)
  - Pull Request #740 - Updated Fibaro FGMS-001 ID's from Technosf (Justin)
--- a/cpp/src/command_classes/MultiInstanceAssociation.cpp
+++ b/cpp/src/command_classes/MultiInstanceAssociation.cpp
@@ -442,7 +442,8 @@ void MultiInstanceAssociation::Remove
 	Log::Write( LogLevel_Info, GetNodeId(), "MultiInstanceAssociation::Remove - Removing instance %d on node %d from group %d of node %d",
 	           _instance, _targetNodeId, _groupIdx, GetNodeId());
 
-	if ( _instance == 0x00 )
+	if ( ( _instance == 0x00 )
+			&& (m_alwaysSetInstance == false) )
 	{
 		Msg* msg = new Msg( "MultiInstanceAssociationCmd_Remove", GetNodeId(), REQUEST, FUNC_ID_ZW_SEND_DATA, true );
 		msg->Append( GetNodeId() );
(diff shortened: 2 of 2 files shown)


Fishwaldo:
I'm not sure i want to implement this. The ForceInstances is a bit of a hack, and I don't want to rely on it to much. 
New Code should be using the `Manager::GetAssociations( uint32 const _homeId, uint8 const _nodeId, uint8 const _groupIdx, InstanceAssociation** o_associations );` calls, not the old `Manager::GetAssociations( uint32 const _homeId, uint8 const _nodeId, uint8 const _groupIdx, uint8** o_associations );` (which I should mark as depreciated)

and thus passing the correct instance to the Manager::RemoveAssociation call. Hence, this shouldnt be required.

martian:
I agree ForceInstances is a bit of a hack, but this fixes a bug it introduces by implementing it at least symmetrically.  When I add an association to, say, node 27, which becomes node 27-instance-0, with the current code I cannot remove that anymore. Do you see another way to solve that?

Fishwaldo:
Instance 0 is essentially the Root Node of a device - So I couldn't see why it couldn't be removed. Either by the Association CC or the MultiChannelAssociation (need to rename in OZW) commands. 
We only need to call the MultiChannelAssocation Remove command when the instance > 0

Can you clarify your statement:

> When I add an association to, say, node 27, which becomes node 27-instance-0, with the current code I cannot remove that anymore

ehhh... that doesn't make sense. your creating a association on Node 27 to Node 27? do you mean adding the controller node (1?) to Node 27. If you didn't specify a instance, it would then be 1.0 (node.instance). why can't you remove that?

martian:
I would also expect instance 0 to be practically equivalent to the node, but my fibaro modules do not treat them the same.

We do need to call MultiChannelAssociation Remove when instance=0 in cases where MultiChannelAssociation Add was called when instance=0. This happens (happened) when ForceInstances is in effect while associating with a non-controller module.

Now that pull request #1152 has been accepted, this change is not relevant anymore. Still I'd like to explain it.

I tried to control a fibaro wall plug (FGWPF-102 v2.5) by pressing a button connected to the second channel of a fibaro dual switch (FGS-223, multi-instance). So this concerns a direct association between nodes, and is not related to the controller. 

My 'receiving' wall plug is node 3, the 'sending' dual-switch is node 2.

I add node 3 to association group 4 of node 2. Because of the ForceInstances workaround in MultiInstanceAssociation::Set(), this is processed as if it was an association with instance 3.0 instead.

The command that is sent is:
MultiInstanceAssociationCmd_Set (Node=2): 0x01, 0x0d, 0x00, 0x13, 0x02, 0x06, 0x8e, 0x01, 0x04, 0x00, 0x03, 0x00, 0x25, 0x46, 0x0e
CC_MULTI_CHANNEL_ASSOCIATION/SET/group=4/[] (no nodes)/MARKER/[instance-3.0]

And, indeed, 3.0 has been added:
Received Multi Instance Association report from node 2, group 4
The group contains: Node 3 instance 0

First issue (now fixed): the wall plug does not respond when I press the button, although I know an extra packet is sent (monitored using an SDR).

Then I try to remove node 3 from association group 4 of node 2.

The command that is sent is:
MultiInstanceAssociationCmd_Remove (Node=2): 0x01, 0x0b, 0x00, 0x13, 0x02, 0x04, 0x8e, 0x04, 0x04, 0x03, 0x25, 0xa0, 0xe9
CC_MULTI_CHANNEL_ASSOCIATION/REMOVE/group=4/[node-3]

Second issue (now not relevant anymore): although the module does not complain, instance 3.0 is still in the association group.

Received Multi Instance Association report from node 2, group 4
The group contains: Node 3 instance 0

That was the reason I proposed MultiInstanceAssociation::Set() and MultiInstanceAssociation::Remove() to be made symmetrical. Because of the way the first issue is now fixed, the second is less relevant.

Modules that got such .0 associations from running the previous code can be cleaned by resetting or excluding and re-including.

martian:
Now that pull request #1152 has been accepted, this change is not relevant anymore. Closing pull request.

Issue #1142, opened 2017-02-19
ForceInstances breaks direct association (from Fibaro FGS-223 to FGWPG-102)


martian:
I tried to create a direct association from a fibaro FGS-223, group 4 (S2 on/off) to a fibaro FGWPF-102  (v2.5) wall plug. The association shows up in domoticz and open-zwave-control-panel. I see the association/set command in the log. But in practice, the wall plug does not switch when the FGS button is operated.

After getting this working using a controller using another zwave stack, and some debugging on open-zwave, I found out the ForceInstances workaround not only changes the association to the controller, but also the associations to other nodes. In my case, that causes the association to become non-functional.

Fishwaldo:
please see comments on #1145

PR #1145, opened 2017-02-19
apply ForceInstances more selectively

martian:
Fix for #1142 (and #1143)

Allow a per-device-model choice between ForceInstances for all associations or just associations to the controller. Fix fibaro FGS-2x3 to do this for associations to the controller only. Includes a modified fix for Remove().

A simpler code change can be found in my [controller-only](https://github.com/martian/open-zwave/tree/ForceInstances-controller-only) branch. That variant does affect all device models, including Qubino.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,4 @@
+- Add config option for ForceInstances limited to the controller node.
 Version 1.5
  - Pull Request #743 - Updated Fibaro Devices from Navstev0 (Justin)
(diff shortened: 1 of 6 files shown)


Fishwaldo:
hrm. Thinking about this, I think Id prefer to apply ForceInstances to the controller node only, and not have a option. If your trying to associate a device to a MultiInstanceAssociation target, then you should be using the newer Manager::AddAssociation (with a instance argument) calls anyway. 

and as I mentioned in the other thread, not in favour of the Remove Change. Can you open a PR for your controller-only PR instead?

PR #1151, opened 2017-02-21
Support for the BeNext TagReader500

hvbommel:
The new BeNext TagReader500 is using a variable Usercode length ranging form 4 up to 10 
digits and fails to work with the current OpenZWave implementation. The previous models where always using 10 digit, sometimes zero filled, usercodes. I have added the device to the xml files and updated UserCode.h & Usercode.cpp to support the new device.

### DIFF
--- a/config/BeNext/TagReader500.xml
+++ b/config/BeNext/TagReader500.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--http://www.pepper1.net/zwavedb/device/302-->
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+
+  <!-- Configuration -->
+  <CommandClass id="112">
+
+    <Value type="byte" genre="config" instance="1" index="1" label="Set to Default" value="-86">
+      <Help>Set all configuration values to default values (factory settings).</Help>
+    </Value>
+
+    <Value type="byte" genre="config" instance="1" index="2" label="Feedback Time" value="15">
+      <Help>To configure the time the beep is automatically turned off in seconds.</Help>
+    </Value>
+
+    <Value type="byte" genre="config" instance="1" index="3" label="Feedback Timeout" value="0">
+      <Help>To configure the timeout to wait for a WAKEUP_NO_MORE_INFORMATION before the error beep is automatically sound. The error beeps are fixed 8 beeps shortly after each other.</Help>
+    </Value>
+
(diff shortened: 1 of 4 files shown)


Fishwaldo:
what exactly is broken with the current UserCode CC? It works fine on other devices that are different lengths?

Anyway, the UserCode CC will be re-factored soon (see #1059), so could you split this PR into config changes and hang back on the UserCode changes till #1059 is done. I'll try to address anything thats broken into the #1059 changes.

hvbommel:
In the previous code, when the user entered let say "1234" as the the code it was extended to 10 digits like "1234000000". So the reply to the device would be "1234000000" instead of "1234". The new BeNext TagReader500 does fail on this and therefore it does not set the code. The fix I made will do the following logical operations: 1. When 10 digits are received (key tags) all works as before. 2. When less then 10 digits are received, return the same less then 10 digits. 3. When less then 4 digits received zero fill to 4 digits and return value. Can you send me a note when I should do a new pull request ? The config changes only make sense after Usercode is updated.

hvbommel:
Once 1059 is available I will retest & debug to see if my changes are still required. Looking at the code I think they still do as a Hardcoded user code length of 10 is used in Usercode.cpp

hvbommel:
Hello Justin,

 

Do you have any idea if and when PR #1175 will be considered ? 

Or is there anything I can do to improve it so that it is easier to get it in ?

 

PS: I did the request on the master branch. Is it better when I resubmit the request on the Dev branch ?

 

Thanks

 

Henri

 

Van: Justin Hammond [mailto:notifications@github.com] 
Verzonden: woensdag 22 februari 2017 06:26
Aan: OpenZWave/open-zwave <open-zwave@noreply.github.com>
CC: hvbommel <henri@vanbml.nl>; Author <author@noreply.github.com>
Onderwerp: Re: [OpenZWave/open-zwave] Support for the BeNext TagReader500 (#1151)

 

what exactly is broken with the current UserCode CC? It works fine on other devices that are different lengths?

Anyway, the UserCode CC will be re-factored soon (see #1059 <https://github.com/OpenZWave/open-zwave/pull/1059> ), so could you split this PR into config changes and hang back on the UserCode changes till #1059 <https://github.com/OpenZWave/open-zwave/pull/1059>  is done. I'll try to address anything thats broken into the #1059 <https://github.com/OpenZWave/open-zwave/pull/1059>  changes.

Issue #1065, opened 2016-12-10
Qubino ZMNHID1 cannot set mode


jotakar:
With the last OZwave update the thermostat ON/OFF ZMNHID1 does not accept the mode change. It remains in Off and can not change to Heat.
I've found that if I set this modification in ozwcfg_0xxxxxxxxx
```
            <CommandClass id="64" name="COMMAND_CLASS_THERMOSTAT_MODE" version="1" request_flags="4" innif="true">
                <Instance index="1" />
                <Value type="list" genre="user" instance="1" index="0" label="Mode" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" vindex="0" size="1">
                    <Item label="Off" value="0" />
                    <Item label="No1" value="1" />
                    <Item label="No2" value="2" />
                    <Item label="Heat" value="3" />
                </Value>
                <SupportedModes>
                    <Mode index="0" label="Off" />
                    <Mode index="3" label="Heat" />
                </SupportedModes>
</CommandClass>
```
I see that the value is 3 for Heat, but if I don't use the Items labeles No1 and No2 OpenZwave gives an error.
It looks like OZW assigns the index in the list as the value to be sent to device. So if in OZW control panel I select Heat the device receive the right value, but if I select No1 or No2 it remains in its actual value (Off or Auto).

jotakar:
I've seen in valueList.cpp this code
```
//-----------------------------------------------------------------------------
// <ValueList::SetByValue>
// Set a new value in the device, selected by item index
//-----------------------------------------------------------------------------
bool ValueList::SetByValue
(
    int32 const _value
)
{
    // create a temporary copy of this value to be submitted to the Set() call and set its value to the function param
    ValueList* tempValue = new ValueList( *this );
    tempValue->m_valueIdx = GetItemIdxByValue(_value);

    // Set the value in the device.
    bool ret = ((Value*)tempValue)->Set();

    // clean up the temporary value
    delete tempValue;

    return ret;
}
```
I see that with
 ```
 ValueList* tempValue = new ValueList( *this );
    tempValue->m_valueIdx = GetItemIdxByValue(_value);
```
We send to the device the index of value in the List not the value of the item, because tempValue contains the index not the value. Could be this the problem?
I can't find the definition of Set for ValueList :(

jotakar:
And more:
In the file cpp/src/value_classes/ValueList.cpp  I've replaced
[...]

PR #1154, opened 2017-02-24
Fix ValueList::SetByValue and add ValueList::SetByIndex()

tobiasgraf:
continuing from
https://github.com/OpenZWave/open-zwave/pull/1035
and
https://github.com/OpenZWave/open-zwave/issues/1065

When using the
```
bool Manager::SetConfigParam
(
		uint32 const _homeId,
		uint8 const _nodeId,
		uint8 const _param,
		int32 _value,
		uint8 const _size
)
```
function, this clearly takes a value as a argument.
Coming down this path, the ValueList::SetByValue need to get a index for the value. That's why I made the first PR.
Now that this got reverted i again got crashes.
So this commit adds a SetByIndex() function. Now both paths should work.

image

The ValueList::SetByLabel and Node::SetConfigParam were the only callers to ValueList::SetByValue(). So I changed the ValueList::SetByLabel to use the new SetByIndex().

@jotakar or @madarwin could you test this?

### DIFF
--- a/cpp/src/value_classes/ValueList.cpp
+++ b/cpp/src/value_classes/ValueList.cpp
@@ -216,7 +216,7 @@ void ValueList::WriteXML
 
 //-----------------------------------------------------------------------------
 // <ValueList::SetByValue>
-// Set a new value in the device, selected by item index
+// Set a new value in the device, selected by item value
 //-----------------------------------------------------------------------------
 bool ValueList::SetByValue
 (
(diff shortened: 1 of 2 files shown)


chronofanz:
Has there been any progress on this?  I also hit this bug where I was trying to set a valuelist by value; it would be great if this fix was checked in, thanks!

cpp/src/value_classes/ValueList.cpp
```diff
@@ -225,7 +225,29 @@ bool ValueList::SetByValue
 {
 	// create a temporary copy of this value to be submitted to the Set() call and set its value to the function param
   	ValueList* tempValue = new ValueList( *this );
-	tempValue->m_valueIdx = _value;
+	tempValue->m_valueIdx = GetItemIdxByValue(_value);
```
chronofanz: It might be cleaner to also call have ValueList::SetByValue to call SetByIndex below

chronofanz:
Understood, thanks for the update!

PR #1125, opened 2017-02-09
CentralScene support

robbiet480:
Continuing work from #1124, #993, but this time using the code base that @nechry suggested. I have personally tested this and it works with my HomeSeer HS-WD100+.

I'm looking for guidance on how to complete this. Right now, the values are being updated, but should I implement a new notification as well so that applications can just listen for that notification instead of having to collect possibly 9 values together?

Also, does anyone know which devices support Central Scene version 1 or Central Scene version 2? My switch seems to only support version 1.

PR #1156, opened 2017-02-26
CentralScene, Create Scene values when specified in configuration file

gizmocuz:
- CentralScene, Create Scene values when specified in configuration file
- Remotec zrc-90, included scene count

The changes in the CentralScene.cpp/HandleMsg file look a bit weird in this compare, best to display the whole file 

Basically the value creation is now in a new function 'CreateSceneValues' and is called from HandleMsg

With this patch, hardware that does not report the amount of scenes supported, and having the 'scenecount' set in the configuration file, will now work.
Before, only the first scene button worked, and when pressing the other buttons gave the "Can't find ValueID for SceneCount" warning message


Fishwaldo:
See #1125

PR #1161, opened 2017-03-02
Adding Trane TZEMT524AA21MA thermostat

djschaap:
No thermostat documentation available; copied verbatim from existing
TZEMT400AB32MAA.xml.


nechry:
please use list instead of byte for selection parameters
remove all "Unknown" parameter ?
the Association Groups element is right ?

djschaap:
I've updated this definition, removing the unknown parameters and using lists wherever possible.  I'm not sure how to determine/verify the association groups.

PR #1162, opened 2017-03-03
Added additional config entry for 2GIG CT-30 Radio Thermostat

pcragone:
The values for the CT30 thermostat that I have were not present in the manufacturer specific list, so I added them in.  Let me know if anything is missing.


pcragone:
So, for what it's worth, I put this PR in because I figured it would solve the problem of not being able to configure my thermostat.  I've recompiled and run ozwcp, deleted the node and re-added it, and it is still unrecognized, even though the ID and type are correctly specified in my local install.  Any thoughts on this?

pcragone:
Not sure what I did, but in regards to my last comment, everything seems to be working correctly now!  OpenZWave correctly recognizes the CT30 with the type and ID specified in this PR as the CT30

PR #1163, opened 2017-03-04
Configuration files for BeNext TagReader500

hvbommel:
This step 1 for the BeNext TagRead500 support. Step 2 will be to update Usercode.cpp and UserCode.h to make the device work. This will be done later, if still needed, because the Usercode class is currently under construction.

### DIFF
--- a/config/BeNext/TagReader500.xml
+++ b/config/BeNext/TagReader500.xml
@@ -5,41 +5,58 @@
   <!-- Configuration -->
   <CommandClass id="112">
 
-    <Value type="byte" genre="config" instance="1" index="1" label="Set to Default" value="-86">
-      <Help>Set all configuration values to default values (factory settings).</Help>
-    </Value>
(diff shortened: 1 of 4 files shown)


nechry:
please don't mix library update and device config.
the param 1 with value -86 is not possible,
parameters 2 and 3, with time is good to add units,
param 5 and 7 use a list with possible value setting,
remove the 6, is not used...
the 8 is also a list?

hvbommel:
I have updated TagReader500.xml based on feedback. Yes, 8 is also a list. PS: I read somewhere that negative byte values (AA = 170 = -86) needed to be pout in the xml as negative values. That is why I put it that way in my 1st version. I have changed it now to 170 and it seems to work ok.

nechry:
I just do some small typo, but I will not merge the CC part. 
But please provide 2 distinct PR, and the one on CC  must be rebase on dev branch

hvbommel:
Leave a comment

hvbommel:
Created new pull request for BeNext TagReader500

PR #1166, opened 2017-03-05
add ZWN-RSM1 plus

kentcalero:
Enerwave ZWN-RSM1 PLUS—Smart Single Relay Switch Module 
http://enerwaveautomation.com/products/zwnrsm1s/ 
http://enerwaveautomation.com/wp-content/uploads/products/zwnrsm1s/ZWN-RSM1-Plus-0208160043-02.pdf


nechry:
According to the pdf you must add a param to your xml file @kentcalero

PR #1167, opened 2017-03-05
Support user code length 4..10 (instead of always zero filled to 10).

hvbommel:
When the BeNext TagReader500 sends a USER_CODE_REPORT, with a value of for example "1234", it does not allow a 10 digit zero filled USER_CODE_SET tag code return. It needs a USER_CODE_SET tag code of "1234" exactly as was send in the USER_CODE_REPORT, and not "1234000000". Only then a keypad user code can be entered. Note that most KEY_TAGS will work because they have a 10 digit key tag value. So only when the user uses the keypad, to enter a user code, the current code fails to work. This is fixed after this update.

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -7095,7 +7095,7 @@ void Driver::processConfigRevision
 			} else if (result->NodeID == 0) {
 				/* manufacturer_specific */
 				m_mfs->setLatestRevision(atol(result->result.c_str()));
-				if (m_mfs->getRevision() < atol(result->result.c_str())) {
+				if ((uint)m_mfs->getRevision() < (uint)atol(result->result.c_str())) {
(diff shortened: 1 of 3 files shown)


hvbommel:
The driver fix, compile error on Raspberry Pi, is not related to the Usercode pull request but because of this issue I could not verify the Usercode change on RaspberryPi. That is why I am submitting it here.

hvbommel:
Questions: 
1. Is there still Usercode work, related to #1059, in progress ?
2. Are there plans to review #1169 ?

Background:
The note below is from my previous pull request. I closed it and did the config changes separate. Thaey are now in but to make the device work #1169 changes are needed.

Old Note for reference: (#1151):
"What exactly is broken with the current UserCode CC? It works fine on other devices that are different lengths? Anyway, the UserCode CC will be re-factored soon (see #1059), so could you split this PR into config changes and hang back on the UserCode changes till #1059 is done. I'll try to address anything thats broken into the #1059 changes."

PR #1175, opened 2017-03-13
UserCode support BeNext TagReader500

hvbommel:
When the BeNext TagReader500 sends a USER_CODE_REPORT, with a value of for example "1234", it does not allow a 10 digit zero filled USER_CODE_SET tag code return. It needs a USER_CODE_SET tag code of "1234" exactly as was send in the USER_CODE_REPORT, and not "1234000000". Only then a keypad user code can be entered. Note that most KEY_TAGS will work because they have a 10 digit key tag value. So only when the user uses the keypad on BeNext TagReader500, to enter a user code, the current code fails to work. Note: See #1151, this was the old pull request that I canceled while #1059 was being worked.

### DIFF
--- a/cpp/src/command_classes/UserCode.cpp
+++ b/cpp/src/command_classes/UserCode.cpp
@@ -255,15 +255,15 @@ bool UserCode::HandleMsg
 
 		if (i == 0)
 		{
-                        //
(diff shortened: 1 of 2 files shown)


DieterKoblenz:
Same issue here. Tags work with a 10 digit key value however with fewer keys it does not work.

PR #1193, opened 2017-03-29
Add BSDmakefile

yokattana:
The default BSD make (at least on FreeBSD) is not GNU make, nor is it compatible with it. This new BSDmakefile will case BSD make to invoke GNU make.


Fishwaldo:
Pushing to Dev Branch

PR #1196, opened 2017-03-29
Add GE 26933 Smart Motion Dimmer

drkp:
Still trying to work out all the details of this device (especially the motion sensor), but this adds support for the configuration parameters.


nechry:
The group 3 name is right?

drkp:
Oops, thought I fixed that already. Fixed.
(no documentation yet on what the difference between groups 2 and 3 is, though!)

nechry:
@drkp :the group_num is also wrong, I fixed now.
try to run 
make xmltest 
before submit a PR

niemyjski:
I'm also seeing the double device issue and am curious what can be done

DigitalBites:
I am seeing these issues as well and would really like to have this device working.  Seeing duplicate switches and sensors and on/off updates only partially being received (during off, light will turn off but leaves switches at 30% or so as it dims off.

PR #1198, opened 2017-04-01
Updated UWP SDK target version to be compiled against 14393 instead of 10240

dotMorten:
Visual Studio doesn't install v10240 SDK by default any longer, so most people will experience problems with opening this project. 
The MinVersion is still 10240 and is still supported to run on, so there's no breaking changes from this.
The only difference is you will now have to have the 14393 SDK installed on your PC, instead of 10240 (which has now been standard for VS for a long time)


dotMorten:
This PR could be closed if #1200 is accepted instead

Fishwaldo:
Thanks

PR #1200, opened 2017-04-01
Add vs2017 support to UWP

dotMorten:
Added support for compiling UWP with VS2017 tooling.
Moved shared identical parts of VS2015 and VS2017 projects into a shared project for ease of maintenance so that files included in the build is separate from build settings (this could later be expanded into being used by Win32 builds as well, if VS2015+ support is desired there)


dotMorten:
Came up with a simpler less intrusive way. Inbound PR...

PR #1201, opened 2017-04-01
Vs2017 support

dotMorten:
Builds on top of #1198.

Use the default platform toolset instead of a hardcoded value, which allows compiling will multiple different versions of Visual Studio (VS2017 and futures etc)


Fishwaldo:
Thanks

PR #1210, opened 2017-04-16
Fixed crash when data_length equals 0

gizmocuz:
I experienced a crash on both Windows and Linux systems when the data_length was 0.
This causes 'OnValueRefreshed' to be called with an empty string, resulting in a crash.

Tested the 'Dev' branch with the 'config' folder from the master.
In the setup i disabled downloading/checking for the manufacturer file update

I suspect this has something to do with 'GreenWave PowerNode 6' node, or right after this node is queried

With this patch the library was able to query all nodes and continue

### DIFF
--- a/cpp/src/command_classes/ManufacturerSpecific.cpp
+++ b/cpp/src/command_classes/ManufacturerSpecific.cpp
@@ -257,6 +257,8 @@ bool ManufacturerSpecific::HandleMsg
         uint8 deviceIDType = (_data[1] & 0x07);
         uint8 dataFormat = (_data[2] & 0xe0)>>0x05;
         uint8 data_length = (_data[2] & 0x1f);
+        if (data_length==0)
+			return false;
         uint8 const* deviceIDData = &_data[3];
         string deviceID = "";
         for (int i=0; i<data_length; i++) {


Fishwaldo:
There is something seriously wrong here. Can you send logs of the messages that cause the crash....

gizmocuz:
Yes, but only in a few weeks
But you can easily simulate thus by setting the size to zero in the code
Bug or no bug, we should make sure this can never happen and check against this
The length is taken out of an received buffer and in my case with one of my hardware types the value is zero

gizmocuz:
Or we should make sure it can accept a zero length but then it should not crash in the following code

Fishwaldo:
The problem here is that the message its crashing on is the deviceID and Product ID. If we don't get that we can never load config files or whatnot. I'd like to see the actual message to see if we can work around a 0 length message and still extract that info.

Fishwaldo:
ping.... any Logs you can share that show this issue?

gizmocuz:
I'm just back... will do so very soon

gizmocuz:
- Checked out a fresh copy of the dev branch
- I needed to correct the visual studio solution (include a few files, fixed compile errors (plz see the comment i made in the dev merge)

Experiencing the same crash, same location, and it seems related to the greenwave powernode 6, but i am attaching the log

Some visual studio screenshots during the crash:

image

image

image

As you see, and empty deviceID is parsed to default_value->OnValueRefreshed
This pull request will check this first.

[OZW_Log_21052017.txt](https://github.com/OpenZWave/open-zwave/files/1016773/OZW_Log_21052017.txt)

The timeout on Node-1 (the Aeotec controller (old one)) worries me a bit as well

gizmocuz:
Here another log file (with the config from the dev branch) same crash, but this one seems to have an additional log line
[OZW_Log.txt](https://github.com/OpenZWave/open-zwave/files/1016784/OZW_Log.txt)

Fishwaldo:
The error wasn't anything todo with the lenght - hence why I was skeptical... 
it was because we had a duplicate Index for the ValueID's. we were trying to set a string on a INT valueID... :)

That will teach me to not use ENUM's for indexes.... thanks for the report..

gizmocuz:
Glad you got it solved!

PR #1214, opened 2017-04-25
Updated MinGW compilation

Ryszard-Trojnacki:
Updated MinGW Makefile - added AES files to compilation.
Fixed inclusion on windows header file to lower case to be able to compile this under Linux (cross compilation).


Fishwaldo:
Thanks

PR #1216, opened 2017-04-29
Add support for Horstmann/Secure SSR302

cyclops1982:
First time i add support for a device. So i have no clue if this is all correct or not, but it seems to work better for me as initially the device showed 3 switches and it surely only has 2 which operate correctly as well.
The heating mode also works, but only after you ask it to report. I believe there might be an improvement there, but without much specification on the device config XML, it's quite difficult to find out how to add more stuff :)

Manual (with zwave details) can be found here:
http://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/1619/SSR%20302%20User%20Manual.pdf


nechry:
All empty cc definition are not required.
The association group have to be defined only with cc 133.

nechry:
Did you perform a make xmltest on it?

cyclops1982:
Sorry, i thought i ran make xmltest but i clearly didn't!

I removed the non-ascii characters now, but still get this:
Warning: <CommandClass> element has non-unique value in 'id' key attribute: 32 at cpp/build/testconfig.pl line 56.
Warning: <CommandClass> element has non-unique value in 'id' key attribute: 32 at cpp/build/testconfig.pl line 81.

Errors: (Please Correct before Submitting to OZW)


Now, the line 56 and/or 81 i can comment or even remove, but it keeps on giving me an error message on that. So, the output of make xmltest isn't very clear for me.

cyclops1982:
Seems that that error message came from another file. Updating to the latest on master works.

nechry:
The 2 Warning line are not error but only "Warning"

the last log entrty: Errors: (Please Correct before Submitting to OZW) are always display. but if issues are presents, they are display. The testconfig.pl can be modified to remove the Error entry in case off not error

cyclops1982:
Thanks! Don't fully understand why those CC's are not needed while they are in the spec. Is that just something openzwave does automatically?

nechry:
take a look to the zwcfgxxx.xml generated you will see yes ozw perform a CC discovering. but some behaviours must be tweaked with specific device configuration file. The basic information are Configuration parameters and association group name. with a thermostat most time the mode is also important. but you do fine your config, just with extra definition not required.

cyclops1982:
Many thanks!

PR #1218, opened 2017-05-10
Update rsroom.xml to support decimal values

techdude200:
Added decimal value descriptors for command class thermostat setpoint. This makes it possible to set the setpoint with decimal values and not only integers.


nechry:
Did you run:
`make xmltest` ?

techdude200:
Now the xmltest validates also after my last update

PR #1219, opened 2017-05-12
Fixes for Qubino Smart-Meter ZMNHTDx

Netbulae:


nechry:
Qubino did a firmware update and change again association definition.
we must keep device compatibility with older version
we will assume this version for S4, S5 et S6 is the default and I will keep the current for s1 to S3 version but commented. Soon we can handle many config for same device.

Fishwaldo:
These guys suck! Always changing things that break backwards compatibility (or getting it wrong from the start!)

nechry:
yes and is not the first time

PR #1224, opened 2017-05-16
Added Philio PSP05-D Motionsensor Outdoor

superjunky:
Tested.

I'm new to github, so bare with me if I make any mistakes. Advice me if needed.

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -784,6 +784,7 @@
 		<Product type="0001" id="0012" name="PAN04-1 Double Relay Switch 2x1.5kW with Power Measurement" config="philio/pan04.xml"/>
 		<Product type="0001" id="0015" name="PAN08-1 In Wall Roller Shutter Controller" config="philio/pan08.xml"/>
 		<Product type="0002" id="0002" name="PSM02-1 Slim Multi-Sensor" config="philio/psm02.xml"/>
+		<Product type="0002" id="0050" name="PSP05-D PIR/Motion Outdoor" config="philio/psp05-d.xml"/>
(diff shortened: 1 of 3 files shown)


nechry:
please check my last commit
https://github.com/OpenZWave/open-zwave/commit/cc0c4ee09f96fc5219ae85c5768987c862a22561.
I think you reuse a other config file as master?

superjunky:
Yes, thanks @nechry .

I'm not sure how to link to a line, but I suggest to not remove line 49. For the same reason it was added to other Philio motion sensors. It helped me out configuring the device.

nechry:
ok but the default value is not 22, only recommended value?

nechry:
with default value no binary report?

superjunky:
Yes, the default value is 4

nechry:
yes the 00010110 (22) seem good configuration

superjunky:
The original setting of the device is 4 according to the documentation. But it made no sense in ozw..
I prefer 22 as default, but other Philio devices have it set at 4. It's up to you ;)

nechry:
ok good I restore the ozw recommended setting. 
I have also rename the config, this config are shared with 3 version B C and D.

nechry:
We must set the real default config value in xml file, is the internal value store in device. If the documentation said 4 we must wrote 4. but yes I use the comment to provide help for ozw usage

nechry:
I jeedom gateway we have recommended setting during the post inclusion process. and I set all recommended values for users

nechry:
If you set a wrong default value (not the one defined) your device will report this value as the current. but internal the real default value still set. with your config (default set to 22) users thing it's the real value and they will not change to get the device working and they will not understand.

superjunky:
Thanks for the advice. Will use it for future reference.

nechry:
you welcome, and happy to get a new contributor

PR #1225, opened 2017-05-17
Add New Notification Type to Get the Hex Values of a Completed Msg Transaction

jonathanhockman:
Adds a `NotificationRaw` that is Queued every time a msg transaction is completed regardless of whether or not `m_notifytransaction` is flagged. Added `m_raw` to notification to store the raw byte data from the completed message. Added `GetRawString()` to Notification to pull the string representation of the bytes. This was all written with python-openzwave in mind.

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -2345,6 +2345,11 @@ void Driver::ProcessMsg
 				Log::Write( LogLevel_Detail, _data[3], "  Message transaction complete" );
 				Log::Write( LogLevel_Detail, "" );
 
+				Notification* notification2 = new Notification( Notification::Type_NotificationRaw );
+				notification2->SetHomeAndNodeIds( m_homeId, _data[3] );
+				notification2->SetRaw(_data);
+				QueueNotification( notification2 );
+				
 				if( m_notifytransactions )
 				{
 					Notification* notification = new Notification( Notification::Type_Notification );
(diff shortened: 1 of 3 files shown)


jonathanhockman:
For the life of me I could not figure out how to return the actual raw byte array and use it in Python properly which is why I resorted the the string. Every time I tried returning the raw byte array I would receive decode errors on the Python side, and would see that the received array was the incorrect length. If anyone has any suggestions that would be great.

jonathanhockman:
@Fishwaldo We figured out the vector to python issue to make it that much simpler to use.

To answer your question, though, the use case is that we are using this on a device that will be using mobile data and are trying to reduce our payload sizes going over the mobile network as much as possible. So, we have two options: we can take the abstracted data and use that to form a reduced payload size which requires extra processing or simply take the small byte array from the message and pass that along to our webservice. We would like to just simply take the raw information in many cases and let our webservice handle the real logic, but OZW is still very useful for many other things (not least of which is handling the actual msg construction coming in on the serial port). 

For our use it didn't make sense to reinvent all the other underlying functionality of OZW just to get the raw msg when OZW already does it.

Also worthy of noting is we are using this with python, so having all of the python bindings in the python-ozw module already in place is certainly an asset. To get our intended result without using OZW would require quite a bit of work that has already been accomplished.

I hope that answered your question. If not I will try to clarify our intentions further.

jonathanhockman:
@Fishwaldo We are going to add the ability to set an option to receive raw notification in the options.xml file, similar to NotifyTransactions option, so it's not always queued unless they have specified they want that.

Fishwaldo:
I really don't see this as useful to anybody but you though, so I'm pretty hesitant to take it in...

jonathanhockman:
That's fair. Figured we'd see see what happens. If it's not something you're comfortable with then we completely understand.

PR #1228, opened 2017-05-19
Fix compile warnings

mienkofax:
During the compilation in linux with the -pedantic fleg I came across the minor warnings that I have fixed.


Fishwaldo:
Thanks

PR #1229, opened 2017-05-24
Fix logging for byte values

### LINKED ISSUE #466: Invalid value ÿ in log message when setting value for COMMAND_CLASS_SWITCH_MULTILEVEL
```
I noticed when setting a value for a COMMAND_CLASS_SWITCH_MULTILEVEL that there 
appears in the log an invalid character.

This might also happen for other classes

Log output:
2015-02-27 10:39:57.053 Info, Node011, Value::Set - 
COMMAND_CLASS_SWITCH_MULTILEVEL - Level - 0 - 1 - ÿ

notice the ÿ in the message

With kind regards,
Rob
```

Original issue reported on code.google.com by `gizmo...@gmail.com` on 27 Feb 2015 at 9:48

armills:
Writing a byte to the string stream ends up converting it as a character. It's cast to an integer here since it's the numeric value we want to see.

Fixes #466

### DIFF
--- a/cpp/src/value_classes/ValueByte.cpp
+++ b/cpp/src/value_classes/ValueByte.cpp
@@ -81,7 +81,7 @@ string const ValueByte::GetAsString
 ) const
 {
 	stringstream ss;
-	ss << GetValue();
+	ss << (uint32)GetValue();
 	return ss.str();
 }
 


Fishwaldo:
Thanks

PR #1238, opened 2017-06-01
Add bitfield.cpp to vcxproj files

dotMorten:
Bitfield.cpp was missing from Windows build, causing linking to fail


Fishwaldo:
There is lots more still broken on the Dev Branch for Windows (Particularly 2013/2010) which I would eventually get to as well... :)

dotMorten:
As long as we can build and most things work :-)

dotMorten:
Is 2010 and 2013 really important to maintain still? Might be simpler to just unify under 2015 (which would support 2017 too)

Fishwaldo:
There were a few projects I know of still using 2013 earlier this year. Not sure about 2010.

Fishwaldo:
Ok. The Dev Branch builds for all versions again. I'll try to keep it working.

dotMorten:
> There were a few projects I know of still using 2013 earlier this year. Not sure about 2010.

I wonder if they were willing to move. It could reduce the amount of maintenance needed for you.

PR #1248, opened 2017-06-12
Support for Leviton ZWave-Plus Dimmers

bmenchaca:


bmenchaca:
Glad I could help - one note, though.  The Fade-On and Fade-Off time for the Leviton dimmers appears to ONLY apply if the dimmer is turned on/turned off at the switch.  If I turn it on over Z-Wave, it's still fading on/fading off.  Not a huge deal, but I am guessing there is an undocumented config value that controls it, or it's a bug in the FW.  Will research.

bmenchaca:
This looks like a "feature" of the implementation of the firmware.  I believe the XML is correct, and it will be up to Leviton to fix (if they view this as an issue at all).

PR #1256, opened 2017-06-18
Added support for POPP Z-Rain

Momotica:
I've added the POPP Z-Rain


nechry:
Many issues /mistakes. Parameter with size 2 cannot be a byte but a short type. Max and defaults values are not good. If you aren't sure how to fix, I will do it latter

Momotica:
I have made some changes, hope this is right.
It's the first time I did add anything.

nechry:
It's better but the 4 from my point of view must a list with item. And the default and max value at 99 (others index)for basic seem not good, most time is 255, do you have the related documentation?

nechry:
http://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/2418/POPE700168_Manual.pdf
Ok I found it and your boundaries seem good.
So just a list for number 4.

Momotica:
Thanks, I added the list to number 4.

PR #1257, opened 2017-06-19
add leviton dzpa1

techgaun:
This is the device: http://products.z-wavealliance.org/products/1959. The config params and association groups match with https://github.com/OpenZWave/open-zwave/blob/master/config/leviton/dz15s.xml so just using that


nechry:
exlude the distfiles.mk file and please add your leviton/dz15s.xml file. we can close the PR and create a new one insted

techgaun:
@nechry hi thanks for reviewing it. I updated this PR to include that change only. The dz15s.xml file is already there: https://github.com/OpenZWave/open-zwave/blob/master/config/leviton/dz15s.xml and it uses the same config parameters.. Do I need to copy that xml file and create a separate one?

nechry:
ok sorry I did't realise

techgaun:
thanks!

PR #1260, opened 2017-06-25
Fix Visual Studio 2008 project

GreenLightning:
As the title says.


Fishwaldo:
Thanks. (I'll be dropping 2008 from the 1.6 release though)

PR #1269, opened 2017-07-10
Flip exception define to work as it is documented

mooninite:
Before this change you would always have exceptions disabled
and always use exit(). Now you can configure this behavior at
compile-time as the documentation states.

This should not disrupt any current projects as the default behavior is still the same.

### DIFF
--- a/cpp/src/Defs.h
+++ b/cpp/src/Defs.h
@@ -146,14 +146,13 @@ static inline int version_cmp(struct ozwversion a, struct ozwversion b)
 }
 
 #include "OZWException.h"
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Already fixed in Dev Tree.

mooninite:
OK. I was unaware the "Dev" branch took priority over master. I did not see any documentation stating that.

To me it looks like master has more recent work committed to it. Typically most projects use "master" as the master branch.

Would you be committing this to master as well? Or updating documentation to state development should be done on "Dev"?

PR #1279, opened 2017-07-21
BeNext/P1dongle.xml

ewgast:
add config file


nechry:
can you provide the BeNext/P1dongle.xml file ?

PR #1282, opened 2017-07-23
Updated GetValueAsDecimal to also work on systems not using . (period…

larslindnilsson:
…) as decimal point


Fishwaldo:
The dotNet wrapper is being depreciated. You should send this fix to the @OpenZWave/openzwave-dotnet-uwp repo instead.

PR #1283, opened 2017-07-24
Added Zooz ZEN20/23/24 configs

mcd1992:
ZEN23 - http://a.co/38I7fJ1
ZEN24 - http://a.co/2VGVCDs

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -1183,9 +1183,9 @@
 		<Product type="0003" id="0088" name="ZSE33 Smart Chime" config="zooz/zse33.xml" />
 	</Manufacturer>
 	<Manufacturer id="015d" name="Zooz">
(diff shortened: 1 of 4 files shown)


nechry:
please add specific devices configuration files
ZEN23 
http://products.z-wavealliance.org/products/2058
ZEN24 
http://products.z-wavealliance.org/products/2060
With parameter and association group

mcd1992:
Alright, this should be correct. Let me know if I missed anything or did something wrong.

Thanks

nechry:
I think it's more easier to read if you use a list instead of a boolean. But if you think not, I will merge, nothing is wrong.

mcd1992:
I'll swap it to a list.

mcd1992:
@nechry should be good now.

Thanks.

Issue #1272, opened 2017-07-17
Silently dropped frames


tobiasgraf:
In networks with heavy load, especially with lots of security frames (NONCE_GET) Controller Command  could hang and never return.
The reason for this:
Receiving a NONCE_GET OZW handles this quite low level, below all queues. When a NONCE_GET is received, a NONCE_Report is immediately sent. If this NONCE_Report fails because of e.g. a CAN, then it is re-transmitted on queue level. But instead of a extra queue for this, the re-transmission gets done in Driver::WriteMsg(). The way this is done (as far as i understand it): If there is a nonce that needs to be re-transmitted, send it instead of the actual frame. The actual frame is then lost. While this is not that big of a deal for a SWITCH_BINARY_SET or something like this, if the "eaten" frame is e.g. a ControllerCommand_HasNodeFailed, then the Queued Controller Command (in the Controller Queue) will have the state 'starting and never recover from this.

You can see this in the attached log at around 13:27:30.971
[OZW_Log.txt](https://github.com/OpenZWave/open-zwave/files/1151883/OZW_Log.txt)
Instead of something like this
```
2017-06-22 13:21:17.256 Detail, contrlr, Queuing (Command) ControllerCommand_HasNodeFailed: 0x01, 0x04, 0x00, 0x62, 0x02, 0x9b
2017-06-22 13:21:17.257 Detail, Notification: HasNodeFailed ControllerCommand - Starting
2017-06-22 13:21:17.257 Detail, 
2017-06-22 13:21:17.258 Info, contrlr, Sending (Command) message (Callback ID=0x00, Expected Reply=0x62) - ControllerCommand_HasNodeFailed: 0x01, 0x04, 0x00, 0x62, 0x02, 0x9b
2017-06-22 13:21:17.266 Detail, contrlr,   Received: 0x01, 0x04, 0x01, 0x62, 0x00, 0x98
2017-06-22 13:21:17.275 Detail, 
2017-06-22 13:21:17.277 Warning, Node002, Received reply to FUNC_ID_ZW_IS_FAILED_NODE_ID - node 2 has not failed
```
you'll see
```
2017-06-22 13:27:30.971 Detail, contrlr, Queuing (Command) ControllerCommand_HasNodeFailed: 0x01, 0x04, 0x00, 0x62, 0x15, 0x8c
2017-06-22 13:27:30.972 Detail, Notification: HasNodeFailed ControllerCommand - Starting
[...]

PR #1290, opened 2017-07-28
Workaround for silently dropped frames

tobiasgraf:
As requested by @Fishwaldo  #1272 as PR

In networks with heavy load, especially with lots of security frames (`NONCE_GET`) Controller Command could hang and never return.
The reason for this:
Receiving a `NONCE_GET` OZW handles this quite low level, below all queues. When a `NONCE_GET` is received, a `NONCE_Report` is immediately sent. If this `NONCE_Report` fails because of e.g. a `CAN`, then it is re-transmitted on queue level. But instead of a extra queue for this, the re-transmission gets done in `Driver::WriteMsg()`. The way this is done (as far as i understand it): If there is a nonce that needs to be re-transmitted, send it instead of the actual frame. The actual frame is then lost. While this is not that big of a deal for a `SWITCH_BINARY_SET` or something like this, if the "eaten" frame is e.g. a `ControllerCommand_HasNodeFailed`, then the Queued Controller Command (in the Controller Queue) will have the state `starting` and never recover from this.

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -1046,6 +1046,15 @@ bool Driver::WriteNextMsg
 		{
 			m_queueEvent[_queue]->Reset();
 		}
+		if (m_nonceReportSent > 0) {
+			MsgQueueItem item_new;
+			item_new.m_command = MsgQueueCmd_SendMsg;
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Thanks!

PR #1291, opened 2017-07-28
Fix high CPU load when adding nodes

tobiasgraf:
The `m_currentControllerCommand->m_controllerStateChanged` was only reset when there is a callback, but all the ControllerCommands called by the manager don't supply a callback method (except the one for custom Controller Commands, where a Callback could be given).
This caused the DriverThreadProc loop to always read that command (e.g. AddNode) from the loop, go into the `WriteNextMsg` and run into the if( `m_currentControllerCommand->m_controllerStateChanged` ), doing nothing and return again and again and again... -> high cpu load because of endless loop until canceled or finished.


Fishwaldo:
Thanks!

PR #1292, opened 2017-07-29
Add Coolcam 1CH Wall Switch

ultrara1n:


nechry:
Please define association group. The current definition is wrong. If you have the device check how Openzwave discovery them by itself.

LouiS22:
Will look into this with the help of the 2 channel users manual.

Levynger:
The Single channel is not identified correctly using this config., 
I have 11 Single, and 2 Dual neo Coolcams, although the type/id seems to be correct, the node identified as "unknown" without any configuration.

for some reason, the controller doesnt write to log in debug mode, still trying to tackle this one.. 
however the screenshot below shows the type and ids which seems to be correct, so no clue why it is not identified correctly.

image

PR #1294, opened 2017-07-31
Adding Inovelli Smart Plug modules

djschaap:


nechry:
Can you also add specific configuration file for Associations groups definitions (Lifeline) http://products.z-wavealliance.org/Search/DoAdvancedSearch?regionId=2&category=-1&manufacturer=609

djschaap:
Done!  Is it acceptable to use the same XML config file for all three entries?  These seem quite simple - a single association group and no known configuration parameters.

PR #1297, opened 2017-08-04
Fix MultiInstance handling for CapabilityReport

tobiasgraf:
It happened sometimes that the MultiChannelCapabilityReport for the same endpoint came in more than once. In this case a additional Instance for the same endpoint gets created. By checking if there is a instance for the given endpoint and CommandClass, this can be prevented.

I'm not sure if this also needs to go into the "all endpoints are the same" part?

zwcfg_0xXXXXXXX.xml with this error
```
<CommandClass id="38" name="COMMAND_CLASS_SWITCH_MULTILEVEL" version="3" issecured="true">
	<Instance index="1" endpoint="1" />
	<Instance index="2" endpoint="1" />
	<Instance index="3" endpoint="2" />
```
Log of 2 MultiChannelCapabilityReport for endpoint 1
[multiinstance_for_endpoint.txt](https://github.com/OpenZWave/open-zwave/files/1200414/multiinstance_for_endpoint.txt)

### DIFF
--- a/cpp/src/command_classes/MultiInstance.cpp
+++ b/cpp/src/command_classes/MultiInstance.cpp
@@ -533,6 +533,15 @@ void MultiInstance::HandleMultiChannelCapabilityReport
 				CommandClass* cc = node->GetCommandClass( commandClassId );
 				if( cc )
 				{
+					// get instance gets an instance for an endpoint
+					// but i'm only interested if there is a related instance for an endpoint and not in the actual result
+					// soo if the result is != 0, the endpoint is already handled
+					bool endpointAlreadyHandled = cc->GetInstance( endPoint ) != 0 ; 
+					if ( endpointAlreadyHandled )
+					{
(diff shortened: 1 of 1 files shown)


AlCalzone:
Just out of curiosity: In ioBroker (using the node-zwave-shared library), one of my Fibaro Dimmers has 4 of each state under SWITCH_MULTILEVEL, while another one has only 1 (as it should be).
Could this bug be the reason?

tobiasgraf:
Possible, if you still have the addNode logs of this you can search for `Received MultiChannelCapabilityReport from node 136 for endpoint`, if one endpoint shows up more than once this could be the issue. 
Btw my log is also from a Fibaro Dimmer, and as it has 2 Inputs (S1 and S2) it should have 2 `SWITCH_MULTILEVEL`

AlCalzone:
Sadly, its too long since I included that device. Maybe the 2nd one has 2 instances, not sure right now. In ioBroker the states are ordered like SWITCH_MULTILEVEL.Level_1, SWITCH_MULTILEVEL.Level_2, SWITCH_MULTILEVEL.Dim_1 and so on. 

But the other definitely has up to 4 occurences of some of those states.

Gonna take a look at the logs when I include my new Roller Shutter controllers over the weekend.

Fishwaldo:
Thanks!

PR #1299, opened 2017-08-12
Added Forest Shuttle L curtains

bsmeding:


nechry:
this is not a valid configuration

Issue #1083, opened 2017-01-05
Zipato Keypad RFID - Sending wake up no more info too early


Mavyre:
Hello,

It seems that the WakeUp_NoMoreInformation is sent too early to the device. HA software doesn't have the time to send the Binary Switch to change the notification, and is then sent to the next WakeUp, meaning the next time that a badge/code is validated. Notification sound are thus completely messed up.

Here are a part of the log that could help you:
```
2017-01-05 16:37:05.819 Info, Node010, Sending (WakeUp) message (Callback ID=0x23, Expected Reply=0x13) - WakeUpCmd_NoMoreInformation (Node=10): 0x01, 0x09, 0x00, 0x13, 0x0a, 0x02, 0x84, 0x08, 0x25, 0x23, 0x67
2017-01-05 16:37:05.828 Detail, Node010,   Received: 0x01, 0x04, 0x01, 0x13, 0x01, 0xe8
2017-01-05 16:37:05.828 Detail, Node010,   ZW_SEND_DATA delivered to Z-Wave stack
2017-01-05 16:37:05.844 Detail, Node010,   Received: 0x01, 0x07, 0x00, 0x13, 0x23, 0x00, 0x00, 0x02, 0xca
2017-01-05 16:37:05.844 Detail, Node010,   ZW_SEND_DATA Request with callback ID 0x23 received (expected 0x23)
2017-01-05 16:37:05.844 Info, Node010, Request RTT 24 Average Request RTT 24
2017-01-05 16:37:05.844 Info, Node010,   Node 10 has been marked as asleep
2017-01-05 16:37:05.844 Detail,   Expected callbackId was received
2017-01-05 16:37:05.844 Detail,   Expected reply was received
2017-01-05 16:37:05.844 Detail,   Message transaction complete
2017-01-05 16:37:05.844 Detail,
2017-01-05 16:37:05.844 Detail, Node010, Removing current message
2017-01-05 16:37:05.845 Detail, Node010, Notification: Notification - Node Asleep
2017-01-05 16:37:05.850 Detail,
2017-01-05 16:37:05.850 Info, Node010, Sending (Send) message (Callback ID=0x24, Expected Reply=0x13) - SwitchBinaryCmd_Set (Node=10): 0x01, 0x0a, 0x00, 0x13, 0x0a, 0x03, 0x25, 0x01, 0xff, 0x25, 0x24, 0x35
[...]

PR #1090, opened 2017-01-08
Set a time interval between commands for Remotec ZXT120, and other slow devices

jotakar:
Remotez ZXT120 (v1.8)  result in timeout error when we send a commandSet fllowed by a commandGet
It looks like it answer an ACK before it process the command and then it doesn't act and the error comes.
Then I create a new param for nodes, named toSleep, that sets a time interval after a command is sent to the node (just before the command message is removed from queue). 
The toSleep parameter is stored in the config file zwcfg_0xxxxxx, in the <Node id="19" ...> area.
I know its working for me and another network working with Domoticz.

PR #1310, opened 2017-09-05
Add 'wakeupdelayms' attribute to <CommandClass id=132> to delay handling of wake up.

h3ctrl:
A workaround for issue: https://github.com/OpenZWave/open-zwave/issues/1083
Also discussed here: https://groups.google.com/forum/#!topic/openzwave/CfeyxHFOf1s

Creates a 'wakeupdelayms' attribute which combines with a timer thread to delay the handling of wakes up. This workaround gives an application more chance to respond to certain events.

### DIFF
--- a/config/device_configuration.xsd
+++ b/config/device_configuration.xsd
@@ -132,6 +132,7 @@
      </xs:restriction>
     </xs:simpleType>
    </xs:attribute>
+   <xs:attribute name='wakeupdelayms' use='optional'/>
    <xs:attribute name='version' type='xs:string' use='optional'/>
    <xs:attribute name='request_flags' type='xs:string' use='optional'/>
    <xs:attribute name='endpoints' type='xs:string' use='optional'/>
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -173,6 +173,9 @@ m_pollThread( new Thread( "poll" ) ),
 m_pollMutex( new Mutex() ),
 m_pollInterval( 0 ),
 m_bIntervalBetweenPolls( false ),				// if set to true (via SetPollInterval), the pollInterval will be interspersed between each poll (so a much smaller m_pollInterval like 100, 500, or 1,000 may be appropriate)
+m_timerThread( new Thread( "timer" ) ),
+m_timerEvent( new Event() ),
+m_timerMutex( new Mutex() ),
 m_currentControllerCommand( NULL ),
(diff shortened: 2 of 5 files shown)


Fishwaldo:
Ok - Two things going on here:
1) A Timer Function.
2) Delay the WakeUpNoMoreInformation Packet.

Overall Observations: Please do this against the Dev Branch, not master. 

Lets start with 1 first:

- I'd like to see this made a bit more Generic - in that any code can specify a a timeout and a method to call back. (std::bind type callback). 
- Move the Timer Code into its own class, rather than the Driver Class. Its getting really crowded in there. (we should start the thread from the driver class though)
This allows the Timer Class to be used for other things as well. (like #1090 which I find current a real nasty hack, hence my reluctance to do that for one single device)


On the 2nd point:
1) I think we shouldn't put the Delay on SetAwake as some applications may check for awake/asleep before issuing commands Instead, I'd put the Delay on Sending the NoMoreInformation packet in the WakeUp::SendPending method instead. something like create a new Method called "SendNoMoreInformation" and register that Method with the TimerClass and the timeout specified in the xml. When the Timer fires, we jump to that function, double check the queued messages and flush if necessary, and then send the actual NoMoreInformation packet. 

I think that should achive what you are after, and keep the state of OZW in sync with the device. 

There are some other generic comments on the code, but lets see if thats still valid after you do this.

jotakar:
The #1090 request  is not for one single device, other devices with the same problem can use this method for entering a delay among send and get commands.  And now it's working fine.
But I like generic solutions so I'm trying to understand this new approach to apply to #1090, but I see it's too specific. Isn't it?

h3ctrl:
Thanks for the feedback. I will try to implement a more generic timer against dev as soon as I get a chance. I think it could then also be used for a non-blocking solution to #1090.

h3ctrl:
I am mostly ready with a new version, but was hoping for some guidance on what language features I can use.

The easiest is to assume that std::function and std::bind are available, but these are C++11 only. The code currently relies on std::tr1::function and std::tr1::bind, as this is what I have on my system (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4)) without having to specify -std=c++11.

Is it safe to require that some form of these exist (whether in std or std::tr1), or would it be better to avoid them completely?

h3ctrl:
More generic version in #1315

PR #1315, opened 2017-09-12
Timer thread and 'delay_no_more_information' option for wake up.

h3ctrl:
A more generic version of https://github.com/OpenZWave/open-zwave/pull/1310

New pull request as this now targets the Dev branch. The timer currently uses std::tr1::function and std::tr1::bind with an ifdef for std::function/bind. This doesn't seem to fit the current platform abstraction as it is more a compiler feature issue, so advice on the preferred way to handle this welcome.


cpp/src/TimerThread.cpp
```diff
@@ -0,0 +1,158 @@
...
+TimerThread::~TimerThread
+(
+)
+{
+	m_timerMutex->Lock();
+	for ( list<TimerEventEntry *>::iterator it = m_timerEventList.begin(); it != m_timerEventList.end(); ++it ) {
+		delete (*it);
+	}
+	m_timerMutex->Unlock();
+	m_timerMutex->Release();
```
Fishwaldo: Could you use the Scoped Locking Macro's we use elsewhere for this. (LockGuard - defined in Util.h)

cpp/src/TimerThread.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+			// Go through all waiting actions, and see if any need to be performed.
+			m_timerMutex->Lock();
+			list<TimerEventEntry *>::iterator it = m_timerEventList.begin();
+			while( it != m_timerEventList.end() ) {
+				int32 tr = (*it)->timestamp.TimeRemaining();
+				if (tr <= 0) {
+					// Expired so perform action and remove from list.
+					Log::Write( LogLevel_Info, "Timer: delayed event" );
+					(*it)->callback();
```
Fishwaldo: I think you should also Lock the nodeMutex here as this will be called from a different thread right?
`LockGuard LG(driver->m_nodeMutex)`

cpp/src/TimerThread.h
```diff
@@ -0,0 +1,112 @@
...
+#ifdef USE_TR1
+#include <tr1/functional>
+#define TT_STDBIND std::tr1::bind
+#define TT_STDFUNCTION std::tr1::function
+#else
+#include <functional>
+#define TT_STDBIND std::bind
+#define TT_STDFUNCTION std::function
+#endif
+
```
Fishwaldo: I'm thinking you can automate the selection of headers etc:
`#if __cplusplus >= 201103L
#include <memory>
using std::shared_ptr;
#else
#include <tr1/memory>
using std::tr1::shared_ptr;
#endif`

should work

cpp/src/TimerThread.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+//-----------------------------------------------------------------------------
+// <TimerThread::TimerSetEvent>
+//-----------------------------------------------------------------------------
+void TimerThread::TimerSetEvent
+(
+		int32 _milliseconds,
+		TimerCallback _callback
+)
```
Fishwaldo: may also want to pass a context void * pointer in, so for callbacks we can pass arbitrary struct etc (not needed for the delayed sending, but for future support)

cpp/src/command_classes/WakeUp.cpp
```diff
@@ -457,23 +500,28 @@ void WakeUp::SendPending
...
+			msg->Append( 2 );
+			msg->Append( GetCommandClassId() );
+			msg->Append( WakeUpCmd_NoMoreInformation );
+			msg->Append( GetDriver()->GetTransmitOptions() );
+			GetDriver()->SendMsg( msg, Driver::MsgQueue_WakeUp );
+		} else {
+			Log::Write( LogLevel_Info, GetNodeId(), "  Node %d has delayed sleep of %dms", GetNodeId(), m_delayNoMoreInfo );
+			TimerThread::TimerCallback callback = TT_STDBIND(&WakeUp::SendPending, this, false);
+			GetDriver()->GetTimer()->TimerSetEvent(m_delayNoMoreInfo, callback);
+		}
```
Fishwaldo: hrm. For clarity, I'd split this out into a new method that just sends the NoMoreInfo:
Something like:
WakeUp::SendNoMoreInfo() {
Msg *msg = new Msg(...)
...
GetDriver()->SendMsg(...)
}

then here just do a:
if ( !allowDelay || m_delayNoMoreInfo ==0 ) {
SendNoMoreInfo()
} else {
TimerThread::TimerCallback callback = TT_STDBIND(&WakeUp::SendNoMoreInfo...);
...
}

Otherwise I'm scared if a application keeps queuing up messages, we could end up in a infinite loop in SendPending

Fishwaldo:
I added some quick comments on the TR1 vs Std namespace in the review.... and a few other things. If you can have a stab at fixing them up, I'll merge. 

Thanks heaps for your help on this. Very much appreciated.

jotakar:
I would like to apply this enhancement to replace the #1090 fix, so I beg for a more commented code, or an explanation about how this modification  works.  Because I see it's applied to wake_up  but the problem in #1090 is "get after set" (the actual solution in #1090 fix this problem but it looks like it is not good enough) 
Thanks.

cpp/src/TimerThread.cpp
h3ctrl: My assumption was that any callbacks which the timer thread runs (where the event was likely set using TimerSetEvent from the driver thread) would have to be responsible for their own locking, to avoid potential deadlocks.

I am not sure the nodeMutex should be locked here, as no node data is directly accessed. For the delayed no more information message, it calls into WakeUp::SendNoMoreInfo which then calls Driver::SendMsg, and Driver::SendMsg then locks the nodeMutex when it looks up the node.

cpp/src/TimerThread.cpp
h3ctrl: Just wanted to double check what was wanted here. The callback function can have its arguments set through std::bind to pass information in that way. Do you mean an extra pointer to store in a struct TimerEventEntry?

h3ctrl:
Hi Jotakar, the timer thread basically allows you to request a function call to be made (including all the function call arguments) at some time in the future. I did try and look if a timer call might be able to solve the "get after set" issue, but I think I need to spend more time in understanding the message queue system first. Someone who is more familiar with it can perhaps comment on if it would work or not.

jotakar:
Thanks h3ctl.   
@Fishwaldo said that your request could solve the "get after set" issue, then he must know how to do.

cpp/src/TimerThread.cpp
Fishwaldo: ehhh. I missed that. 👍

cpp/src/TimerThread.cpp
Fishwaldo: I'm inclined to err on the side of caution. In the WakeUp Case - as you mentioned, its handled lower levels, but other cases might not.

In fact, I can also see a potential race here - If we put something on the callback, and then delete the instance (while the callback is pending) we could end up with a crash...

(eg, in this case, if the Node that had the WakeUp CC was deleted while a timer was pending, when it fires we would be accessing something use after free).

Maybe we need to save a reference to each callback in the instance, and in the destructor, cleanup any pending timers for that instance...

Maybe a subclass? then make WakeUp Class inherit it?

Fishwaldo:
@jotakar you would register a timer for doing the Get Request instead of the Delay you are doing now. But wait till we sort this out before trying :)

Fishwaldo:
I just pushed to your tree with the subclass and changes I suggested. Do you want to give it a test and confirm working as intended from your side?

h3ctrl:
My thought for stopping callbacks on a deleted object was to store the event under a key in the TimerThread (perhaps using a map for faster lookups). The object requesting a callback could then pass its 'this' pointer as the key, and then in the destructor have a call to something like TimerThread::TimerDelEvents(key = this) to remove all events under the key. Anyway, you beat me to implementing it :).

It is working for me. I added in a couple of iterator increments and I removed a call to TimerThread::TimerDelEvent after a Timer::TimerFireEvent as there is already the chain: Timer::TimerFireEvent -> Timer::TimerDeleteEvent -> TimerThread::TimerDelEvent

h3ctrl:
Sorry, I've had quite a hectic last week and have lost track of things a bit. Just wanted to check if there are still any outstanding changes to be made?

jurgenweber:
Ping, I would love to see this merged.

jurgenweber:
ping, this has been a long while now..

jotakar:
when will it included in libs?

jotakar:
Ping... is this to be included in library? What's the problem? It's aproved

janpascal:
Another ping... Could this PR be merged please? Or is there some problem with it still?

jotakar:
It depends on Fishwaldo... so sit down and wait, perhaps when hell is frozen ... :)

PR #1317, opened 2017-09-13
Neo Coolcam Light Switch (1 channel)

LouiS22:
This should fix #1292


Levynger:
The Single channel is not identified correctly using this config.,
I have 11 Single, and 2 Dual neo Coolcams, although the type/id seems to be correct, the node identified as "unknown" without any configuration.

for some reason, the controller doesnt write to log in debug mode, still trying to tackle this one..
however the screenshot below shows the type and ids which seems to be correct, so no clue why it is not identified correctly.

image

sophof:
I have a similar issue. In addition associating a switch with a plug on group 2/3 doesn't work, but I'm unsure where that fault lies. According to the documentation it should work.

PR #1334, opened 2017-10-01
add new device config: remotec zts-500

rayslinky:
This change adds the Remotec ZTS-500 Z-wave thermostat.


nechry:
I think it would be better to use the list type for parameters 4, 10, 14 and 15 in order to simplify the selection by end users.
Also try to simplify the names of groups of associations. 1) Lifeline, 2) Heating Mode group, 3 Cooling Mode group

PR #1338, opened 2017-10-14
Update LS02ch config

LouiS22:
Seems one of the config parameters was wrong


nechry:
@LouiS22 I think is better to use a list type with the 4 possibilities ?

LouiS22:
If you say so, but I have no clue how to do that :)

nechry:
JUst the 4 possibilities like the description, or I did not fully understand

PR #1349, opened 2017-10-22
Add zWave node instance to logs

travellingkiwi:
Add the zWave node instance to logs - this makes it easy to see which instances in a multi-instance node a message is from. e.g. the Fibaro double switch. Or the Qubino relay x2


Fishwaldo:
Closing

PR #1351, opened 2017-10-22
Fix handling Indicator report data

achakirov:
Some devices (like multi button scene controllers Cooper RFWC5/RFWDC) utilize more than 1 bit to report status of their indicators. The fix makes the received data to be treated properly as a value in 0-255 range.


Fishwaldo:
Hi,
In the future, please do not combine updates like this (a bug fix to code and a config file update). Please use separate PR's. 

Thanks

achakirov:
Ok, I didn't know. Just drop the config file update changes -- I'll do a
separate PR.

PR #1355, opened 2017-10-26
Updated Alarm/Door index labels

gazoscalvertos:
Fixed the labels to allow openzwvae to set the values as this caused an exception


nechry:
I'm not sure if this change is really useful, named values. If it's the name of the sound it would be more interesting. Also many software will display the list of values in ascending order, the result will not be pleasant to read.

PR #1341, opened 2017-10-17
Eurotronic spirit z + thermostat setpoint v3

tobiasgraf:
The PR for #1340

The Thermostate_Setpoint has 2 interpretations. OZW seems to follow Interpretation B per default. 

>It has been found that early implementations of this Command Class specification apply two non-
interoperable interpretations of the bit mask advertising the support for specific Setpoint Types.
As a consequence, one may find thermostat products and controller products in the marketplace which
implement either of the two bit mask interpretations found in Table 106. The notation x.y indicates Bit
Mask byte x, bit y.

>Implementations of Thermostat Setpoint Command Class, version 3 MUST comply with Interpretation A.

The EUROtronic Spirit Z Thermostate (and comet and stella z too)  uses version 3 and thus follows interpretation A.

image

This PR adds a config Parameter `<CommandClass id="67" base="0" typeInterpretation="A" />` .
Together with this i added the Capabilities Get and Report Command support (its now possible to get the min and max from the devices).
And while i'm at it, the Thermostate_Mode got new values as well.

Also adds a config file for the Spirit Z Thermostat

PR #1385, opened 2017-12-05
Thermostat setpoint v3

tobiasgraf:
Separate PR for Thermostat Setpoint V3

Adds new setpoint types and `Capabilites` Command to discover min / max of the setpoint types.
Also adds new configuration parameter `typeInterpretation` because there is a interpretations Issue with this command class.

>It has been found that early implementations of this Command Class specification apply two non-
interoperable interpretations of the bit mask advertising the support for specific Setpoint Types.
As a consequence, one may find thermostat products and controller products in the marketplace which
implement either of the two bit mask interpretations found in Table 106. The notation x.y indicates Bit
Mask byte x, bit y.

see #1341 for more details


Fishwaldo:
Thanks

PR #1390, opened 2017-12-09
avoid leaking a driver if a second driver got ready for the same home id

gms1:


Fishwaldo:
Can you describe why this change is needed, and why  it avoids the race condition in destructing drivers?

gms1:
the race condition happened because the driver was still in the "ready map" while the dtor was already called

In Manager::SetDriverReady the driver gets currently removed from the pending list and added to the ready map, which has the home id as key. So if two drivers are loaded which are using the same home id, the second one wins, but the first one is not even removable anymore, because all references to this driver are lost 

In this commit the pending list has been replaced by an 'all driver' map, having the controller path as key
with the added advantage for easier lookup, if a driver for a controller path is already loaded ( see Manager::AddDriver) or to iterate over all available drivers (see Manager::SetPollInterval)...

Guenter

gms1:
this solution seems to me the cleanest; on the other hand this issue is anything but urgent and could have been solved with fewer code changes (keeping the driver reference in an additional list / map, deleting one of the two competing drivers, ...)

cpp/src/Manager.cpp
```diff
@@ -315,44 +301,25 @@ bool Manager::RemoveDriver
...
-			return true;
-		}
-	}
-
 	// Search the ready map
 	for( map<uint32,Driver*>::iterator rit = m_readyDrivers.begin(); rit != m_readyDrivers.end(); ++rit )
 	{
 		if( _controllerPath == rit->second->GetControllerPath() )
 		{
-			/* data race right here:
```
Fishwaldo: This work doesn't fix the race condition described here. Please restore this comment.

gms1:
> I think a more simple solution would be as soon as we get the HomeID in the Driver 
> Class, do a lookup if it exists, and bail out if it doesn't (so the second driver never > gets fully loaded).

I believe we are lost, because fixing this issue will not be worth the effort.

The drivers are receiving their homeid/nodeId from the z-wave interface, after the m_controller->PlayInitSequence has been called, so at this time also the second driver can be considered (more or less) as fully loaded
and as soon as we want to delete this driver, the driver will start to send notifications within the dtor and no one will be able to distinguish to which driver those notifications belong
Do you agree with this opinion? If so, let us close this PR, if not, I would like to help, but then I would need more information on what is desired

gms1:
regarding the race condition:
doing this notifications within the destructor may trigger the attempt to access a partially destroyed driver, but keeping the driver in the ready map while its dtor has already been called, opens the possibility to access this partially destructed driver. both should be avoided, or like the comment in the destructor of the driver says, is currently asking for trouble

further quote from this comment:
> We might need a flag around the Manager::GetDriver() class that stops applications
>	 * from getting a half destructed Driver Reference, but still retain a Internal GetDriver() method
>	 * that can return half destructed Driver references for internal classes
with the proposed changes, we would have stopped the GetDriver method from getting this half destructed Driver reference and using the proposed m_drivers map, we would have the possibility to retrieve this half destructed Driver for internal use

PR #1393, opened 2017-12-11
Request to add support for Philio PAN06

allan-gam:
Adding support for Philio PAN06

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -878,7 +878,7 @@
 	<Manufacturer id="013c" name="Philio Technology Corporation">
 		<Product type="0001" id="0001" name="PAN11-4 Smart Energy Plug In Switch"/>
 		<Product type="0001" id="0003" name="PAN04-1 Double Relay Switch 2x1.5kW with Power Measurement" config="philio/pan04.xml"/>
-		<Product type="0001" id="0004" name="PAN06-1 In Wall Dual Relay (1 Way) Switch Module"/>
+		<Product type="0001" id="0004" name="PAN06-1 In Wall Dual Relay (1 Way) Switch Module" config="philio/pan06.xml"/>
 		<Product type="0001" id="0005" name="PAN07-1 TWO SPDT Switch module with meter" config="philio/pan07.xml"/>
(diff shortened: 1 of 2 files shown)


rrozema:
Option 4 says in the help "0 disables", however min value is 1. So the user can not disable auto off using this settings file?

allan-gam:
According to the [PAN06-1A manual](https://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/1181/PAN06-1A%20manual-20141118.pdf) valid values for parameter 4 are 0 to 0x7FFF. Where does it say that min value is 1?

rrozema:
line 27: <Value type="short" genre="config" instance="1" index="4" label="4. Auto off time: 0=Disable,  1s ~ 32767s delay" **min="1"** max="32767" value="0">

nechry:
Please remove numbers in labels and also the label must be simplified, use the help element for explanation

rrozema:
I agree that there is to much text in the labels that should be moved into a help element. However i -as a beginner user- personally like having the index number in the label as this is often also how it"s presented in the device's manual. Having the index number onscreen makes it a lot easier to match the screen with the instructions in the manual.

allan-gam:
Having a closer look at the text, I found it necessary to make a total overhaul of the labels and help texts except for the last item "Existence of endpoint 3 Action" (I'm unable to grasp what it's about) The original text seems to be fetched from the user manual and it kind of "cheesy" mildly put. It took me a while to actually understand what they meant.

Doing this, makes it look a bit different from the users manual but hopefully it's more understandable. What do you guys think?

nechry:
@rrozema recall the parameter index number must be perform by your gateway UI. 
Like this for exemple:
paramters

nechry:
@allan-gam auto="false" is only for group 1, auto="true" is accepted for other groups if required.
most of time group 1 is for Lineline and we must add the primary controller, others groups are in general for directs associations

nechry:
@allan-gam if you run a ```make xmltest``` all errors will be reported, is a good habit before submit a PR. I fixed the PR

allan-gam:
Thank you @nechry

PR #1399, opened 2017-12-16
Add additional Z-Wave devices

rotx:
Fortezz WWA-01AA Water and Freeze Alarm
GE 45601 Advanced Remote
Leviton VRCZ4-MRZ 4-Button Zone Controller with Switch


nechry:
thank's @rotx

PR #1400, opened 2017-12-17
Removing references to pepper1

### LINKED ISSUE #1397: Documentation links pointing to Pepper1 DB are broken
It looks like the Pepper1 Z-Wave product DB is no more.  Digging through some links on their site, there is the [Z-Wave Alliance Product Catalog](https://products.z-wavealliance.org/) that appears to have plenty of info.  I've never used Pepper1 so I can't say if it's the same sort of stuff.

I'd be willing to go through and update the documentation if this seems like a reasonable location to point people to.

https://products.z-wavealliance.org/

veleek:
Resolves #1397 

The Pepper1 DB no longer exists.  This removes a bunch of broken links from various configuration files.  And removes support for the Pepper1 metadata item (which was used by one product only).


veleek:
Ping!

Fishwaldo:
Thanks!

PR #1413, opened 2018-01-05
Add August Smartlock Gen 3

shaun040:
Add August Smartlock Gen 3


nechry:
Can you add a config file with associations definitions 
https://products.z-wavealliance.org/products/2624

Issue #1047, opened 2016-11-21
Please add the Popp 012501 Strike Lock Control device


catweazle0172:
Hello,

Please add the Popp 012501 Strike Lock Control 
http://www.popp.eu/wp-content/uploads/2016/11/Manual_ElectricStrikeLockControl_POPP_en.pdf
This code is working for me:
Add the following line in the manufacturer_specific.xml under 	<Manufacturer id="0154" name="Popp">
Product type="0005" id="0001" name="012501 Strike Lock Control" config="popp/012501.xml" /

[012501.xml.zip](https://github.com/OpenZWave/open-zwave/files/604677/012501.xml.zip)
Thanks!

nechry:
close with https://github.com/OpenZWave/open-zwave/commit/d0e42ec0c30f5066b6bac243d99f126d11638687

Issue #801, opened 2016-03-15
Enerwave zwn-rsm2 device shares device id with existing device, would like to add parameter


instantchow:
The Enerwave ZWN-RSM2 Smart Dual Relay Switch Module ([pdf](http://enerwaveautomation.com/wp-content/uploads/products/zwnrsm2/zwn-rsm2_ac_092514.pdf)) carries the same product ID and device ID as the ZW-20R. There does not seem to be any parameter collision between these two devices as the ZW-20R has one parameter at index 1 while the RSM2 has one parameter at index 3.

I have added the ZWN-RSM2 singular config parameter to the existing zw20r.xml config file. Perhaps they can coexist?

https://github.com/instantchow/open-zwave/commit/bee97ad45a2ad432bc1cce58b7127ec97d4a8abb

Since I don't own the 20R Dual outlet device I cannot tell you definitively if this will be a problem. Perhaps @chadparry (original contributor for this device) can help.

also somewhat related #141

chadparry:
I can test a config with the ZW20R if that will help. I agree that there 
isn't any reason they should have a problem coexisting. But it will be 
confusing for users, because inactive config entries will be published. 
I can't think of any solution to avoid that.

Fishwaldo:
Unfortunately having non-existent config values will cause timeouts when OZW starts up, and both config values will be exposed in OZW. Best to bitch to the manufacturer about using duplicate ID's

Fishwaldo:
When #141 is done, we can add this.

kentcalero:
this seems to still be a problem, I'm experiencing it with zw20r (cannot find in zwave alliance db) and zwn-rsm1. How do we handle different devices sharing the same ID?

nechry:
We add a commented entry in database pointing on a specific xml file. You must manually comment and uncomment the two entries. It's a manuel action. There files are only useful at the inclusion, if you dont remove your zwcfg file they dont used anymore. It's a workaround but at the moment it's the only solution we have

kentcalero:
That's what I figured. A solution for the future could be an update to manually set the config file for each node. This will also allow for easy updating after the node is added. This should not be a complicated add as it's just a button that performs the load procedure and the specific file name entered by the user.

Kent L. Calero

Sent from the field through my iPhone.

On Feb 25, 2017, at 11:42 AM, Jean-François Auger <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #1415, opened 2018-01-08
Added support for Enerwave device ZWN-RSM2 and ZWN-RSM2-PLUS Smart Du…

austinmroczek:
…al Relay Switch Module

Note:  these devices use the same device ID as the ZWN-RMS1-PLUS device so I created a commented out entry in manufacturer_specific.xml that will have to be manually changed to enabled this device.

run "make xmltest" -- complete
run "make dist-update" -- complete

Tested on:
Linux rpi2 4.9.59-v7+ #1047 SMP Sun Oct 29 12:19:23 GMT 2017 armv7l GNU/Linux

This is related to issue #801 .

### DIFF
--- a/config/enerwave/zwnrsm2plus.xml
+++ b/config/enerwave/zwnrsm2plus.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--    
+        Enerwave ZWN-RSM2 Smart Dual Relay Switch Module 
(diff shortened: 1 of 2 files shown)


trentjw:
thought I would give an update on this device the Enerwave device ZWN2-plus that I installed has its own ID number `enerwave unknown type 0111 id 0606 switch`  where the single relay version has id 605
I had to uncomment and change the id number for this to be recognized incase somebody else runs into this issue

PR #1420, opened 2018-01-14
Add Neo Coolcam NAS-WS01Z flood sensor

drkp:
This appears to have exactly the same parameters as the NAS-WS02Z sensor.


nechry:
Thanks

PR #1427, opened 2018-01-25
Updated ZHC5010 configuration

philipsoeberg:
I have four of these nodes working through v1.4 and v1.5 for node-red and domoticz and phyton.
To assist first-time users with these units, it is important that:
1) ZHC5010 Configuration which can enable/disable direct control over relay can be controlled and
2) ZHC5010 responds fast by default.

To do this, XML config file must have
1) MultiInstance CC set to mapping="all" (Yes, redundant but description enables users to change if instance0 is not needed (No direct Relay control)
2) Disable CC SWITCH_ALL. It seem not supported in v2.03 firmware (and earlier) and it cause OZW to timeout on startup.
3) Remove default Controller associations to any other CC than BASIC_REPORT, which is needed by Domoticz/Node-Red to learn of button-pres (which on/offs the internal switch).

If Controller is associating with all possible association groups, RTT may go as high as 4500ms if quick-pressing different buttons on the ZHC5010. I suspect this is due to OZW polling the switch value after a BASIC_REPORT CC which seem to clog up ZHC5010s tx pipe, but thats guesswork. This XML configuration makes ZHC5010 respond quick and fast out-of-the-box (on new network joins.. for already joined devices you still need to manually change associations or leave/join again).

Also updated some descriptions

### DIFF
--- a/config/logicsoft/ZHC5010.xml
+++ b/config/logicsoft/ZHC5010.xml
@@ -98,7 +98,7 @@
 			<Item label="Relay follows the state of button #2; when button is down the relay is on and when button is released the relay is off" value="10" />
 			<Item label="Relay follows the state of button #3; when button is down the relay is on and when button is released the relay is off" value="11" />
 			<Item label="Relay follows the state of button #4; when button is down the relay is on and when button is released the relay is off" value="12" />
-			<Item label="Relay is only controlled by commands sent to the root device. Commands to the root device will not be forwarded to device 1" value="13" />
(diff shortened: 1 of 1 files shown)


nechry:
please remove all auto="false" is the default behaviour for group other then first one.

philipsoeberg:
Removed auto="false" as suggested.

philipsoeberg:
Changed wording in a XML comment field

PR #1432, opened 2018-01-29
add new ct80 product type and id

chilicheech:


nechry:
Winter is coming... every day a new thermostat 
Thanks @chilicheech

PR #1438, opened 2018-02-06
WIP: Add Linear/GoControl WD500Z5-1

marthoc:
Added entry to manufacturer_specific.xml and added device specific file linear/wd500z5-1.xml, including configuration parameters and central scene parameters for this device.

### DIFF
--- a/config/linear/WD500Z5-1.xml
+++ b/config/linear/WD500Z5-1.xml
@@ -35,12 +35,5 @@
 			<Help>The Dim Timer (how fast the dim steps happen). It can be set to a value of 1 to 255. This value indicates in 10 millisecond resolution, how often the dim level will change.</Help>
 		</Value>
   </CommandClass>
-	<!-- Set the correct parameters for COMMAND_CLASS_CENTRAL_SCENE -->
-	<CommandClass id="91" name="COMMAND_CLASS_CENTRAL_SCENE" version="1" request_flags="4" innif="true" scenecount="2">
-    <Instance index="1" />
-    <Value type="int" genre="system" instance="1" index="0" label="Scene Count" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="-2147483648" max="2147483647" value="2" />
-    <Value type="int" genre="user" instance="1" index="1" label="Top Paddle Scene" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="-2147483648" max="2147483647" value="0" />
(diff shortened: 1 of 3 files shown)


marthoc:
I added a commit that removed the Central Scene parameters I had added in an earlier commit as some attributes do not appear to be supported - `make xmltest` was erring out on the attributes as follows:
```
/open-zwave/config/linear/WD500Z5-1.xml:39: element CommandClass: Schemas validity error : Element '{http://code.google.com/p/open-zwave/}CommandClass', attribute 'innif': The attribute 'innif' is not allowed.
/open-zwave/config/linear/WD500Z5-1.xml:41: element Value: Schemas validity error : Element '{http://code.google.com/p/open-zwave/}Value', attribute 'poll_intensity': The attribute 'poll_intensity' is not allowed.
/open-zwave/config/linear/WD500Z5-1.xml:42: element Value: Schemas validity error : Element '{http://code.google.com/p/open-zwave/}Value', attribute 'poll_intensity': The attribute 'poll_intensity' is not allowed.
/open-zwave/config/linear/WD500Z5-1.xml:43: element Value: Schemas validity error : Element '{http://code.google.com/p/open-zwave/}Value', attribute 'poll_intensity': The attribute 'poll_intensity' is not allowed.
/open-zwave/config/linear/WD500Z5-1.xml fails to validate
```
This is unfortunate as when these items are present in the device-specific file Home Assistant pulls in the correct parameters and recognizes the Central Scene features.

Also ran `make dist-update` which added linear/WD500Z5-1.xml as well as several other files (likely caused by the authors of those PRs not running `make diet-update` before PRing).

marthoc:
Control of this device is erratic in Home Assistant (dimmer does not respond consistently to commands - the turn_on command may set brightness to 90, or 9, or 3, etc, and brightness commands may not result in that brightness being set, e.g. if 90 is requested, the result may be a different value). I suspect this is related to the COMMAND_CLASS_SWITCH_MULTILEVEL. Changing the "Level" value to 100 (from 255) as with the Linear WD500 does not improve the behaviour.

marthoc:
@chilicheech Thanks for the suggestion! Adding refresh_value and delay seems to have fixed the erratic behaviour - now after a command the frontend takes a few moments to catch up (as is expected) but it’s working accurately. Do you think this is down to the OZW config or is this an HA quirk?

chilicheech:
I believe that happens because that dimmer doesn't report its level to the controller. So HA needs to be set up to poll the dimmer for its level shortly after setting the level.

PR #1443, opened 2018-02-13
Fix Fibaro Dimmer 2 FGD212 Parameters

cydine:
Parameters 53 and 59 are mixed up.

https://manuals.fibaro.com/content/manuals/en/FGD-212/FGD-212-EN-T-v1.3.pdf

This was fixed in f342283 but seems to have been (accidently) reverted in 89cee9a.

### DIFF
--- a/config/fibaro/fgd212.xml
+++ b/config/fibaro/fgd212.xml
@@ -249,7 +249,7 @@ Default setting: 10 (10%)</Help>
 1-32767 (1-32767 seconds);
 Default setting: 3600 (3600s)</Help>
     </Value>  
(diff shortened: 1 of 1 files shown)


nechry:
the param 53 size is suppose to be 2.
https://manuals.fibaro.com/content/manuals/en/FGD-212/FGD-212-EN-T-v1.3.pdf
@cydine, can you patch it also?

cydine:
A lot of those 2 byte parameters were change to 1 byte in 3a20b0b and ae2803b even though they are listed as 2 bytes in the manual?

What do you think?

cydine:
OK went through all the parameters and set them as per correct sizes in https://manuals.fibaro.com/content/manuals/en/FGD-212/FGD-212-EN-T-v1.3.pdf

PR #1444, opened 2018-02-13
Add Visual Studio configs for 2015 and 2017

bibi21000:
This filles will be needed for installing python-openzwave on Windows


bibi21000:
Will propose an update soon

PR #1445, opened 2018-02-13
Add Visual Studio configs for 2015 and 2017

bibi21000:
This filles will be needed for installing python-openzwave on Windows

### DIFF
--- a/cpp/build/windows/vs2015/OpenZWave.sln
+++ b/cpp/build/windows/vs2015/OpenZWave.sln
@@ -1,0 +1,40 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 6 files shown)


Fishwaldo:
Could you explain what the actual changes are? It’s hard to see from the PR!

bibi21000:
Essentially number version (msbuid) and win64 target.
There are others update made by VS2017 and VS2015 when converting VS2010
project.
The good news is that these projects can be built from command line using
MSBuild.
That's what I use to install python_openzwave with pip.


2018-02-14 8:26 GMT+01:00 Justin Hammond <notifications@github.com>:

Fishwaldo:
Ok - But I can currently build with VS2010, VS2013, and VS2015 by just letting msbuild automatically upgrade the VS2010 files. (I could add a VS2017 as well I guess)
My only concern is adding individual project files for each VS version is going to be a maintenance nightmare (meaning I have to open up 3-4 different versions of VS everytime a build change is required)

So unless something is happening where the automatic upgrade isn't working, I'm hesitant not to apply this

bibi21000:
OK
A contributor seems to have a solution to fix automatic build without these updates.
You can close this PR or let it open without merging. 
I will give more infos or opening a new one if he fails/succeeds.

PR #1448, opened 2018-02-20
Make all local import statements use correct relative paths

DvdGiessen:
Import statements around the code base do not use relative paths but all use absolute paths. That currently happens to works fine when compiling OpenZWave itself, since the `src` directory is added to the system library header import path and thus the compiler (after some tries) happens to resolve to the correct file.

However, after OpenZWave is installed as a system library and one tries to use it, one might encounter errors like this one:
```
In file included from /usr/local/include/openzwave/Driver.h:37:0,
                 from /workspace/src/main.cpp:9:
/usr/local/include/openzwave/value_classes/ValueID.h:33:18: fatal error: Defs.h: No such file or directory
 #include "Defs.h"
                  ^
compilation terminated.
```

In this example, `ValueID.h` is including `Defs.h` without a proper path, which fails once OpenZWave is installed as a system library since now the `Defs.h` is not directly in the search path. Sure, modifying the include path resolves that as well, but it's not so clean a solution.

Another solution is to use proper relative paths in all `#import` statements. This pull request refactors all source files to properly use relative paths, as well as modify the Makefile so that the system library header path is no longer mangled which enabled the old scenario. As this only modifies the relativity of import paths, builds are unaffected.


Broekman:
What was the reason for closing this? Would like to see this happen as well.

Legomaniac:
I have the same question as Broekman. On my NUC running Fedora 27, this caused the pip install to fail until I made some symlinks, as suggested in the Home Assistant forums. https://community.home-assistant.io/t/zwave-failed-to-initalise-after-upgrading-to-0-63-and-how-i-solved-it/42953/10

I have a feeling those symlinks shouldn't be necessary.

DvdGiessen:
Oh, I closed it since only after opening this pull I realized the package was set up for use with `pkg-config`, which addresses the include path problem when having OWZ installed as a system library. With the original problem dealt with, I didn't want to leave a pull request open just for the sake of making changes.

However, if there is interest for other use cases and the OWZ people see value in merging these changes, I'm all for it.

Broekman:
> I realized the package was set up for use with pkg-config

And you've just made me realize it. In that case it's better to leave it as is since it can possibly break existing programs using OZW. All major Linux distro's have OZW in their package managers anyways and for custom projects it's pretty much a one liner to fix.

Dexter86RE:
Hi there,

i'm pretty new to programming under Linux and I was trying to get the OZW libraries working on Arch-Linux and am getting the error described above.

Could someone help me with that? How do I use pkg-config to fix this?

Thanks in advance!

PR #1465, opened 2018-03-16
Add support for the Homeseer HS-WS200+ wall switch

kylehendricks:
To test I ran this in the latest version of home assistant.  I also ran `make xmltest`.

Config was derived from the official manual for the HS-WS200+


nechry:
thank's @kylehendricks

PR #1469, opened 2018-03-21
Configs for Inovelli nzw30, nzw31, nzw97

vondit:
Adding configuration options and parameters for Inovelli Rocker switch and dimmer


nechry:
sometime we add xml comment or directly in parameter description, but the default values changes form application version to another so it's not a big deal. If a parameter exist only for a version, sometime we add comment. In a next future, we try to figure a way to apply a config for a specific version, but it's not yet ready.

PR #1473, opened 2018-03-30
Allow build with newer Visual Studio

dotMorten:
This change allows you to build the project with newer versions of Visual Studio.
Without this, VS2017 will complain that the v10.0 C++ toolset isn't installed. With this it'll pick v14.0 or 14.1 (depending on which are installed), and will support future versions as well.

Also added the `Localization.cpp`/`.h`. Without these, linking the generated static libs will fail.


dotMorten:
Ping @Fishwaldo

Fishwaldo:
Thanks

PR #1475, opened 2018-04-03
Add ANZ mappings for Fibaro and Aeotec, removed AU tags

Madelaide:
Added Fibaro AU items (RGBW, and Shutter2 controller), plus added AU to ANZ fibaro and Aeotec devices. This aids configuration debugging as not all devices are identified correctly (fibaro FGD211 in my case)


nechry:
Please remove the AU region in name, this is not relevant.

Madelaide:
I respectfully disagree, and find this additional information useful; especially when trying to expand the Fibaro coverage for ANZ, which is pretty sketchy. 

I have added a [device id] to the description; because different devices were being mapped to the same apparent type, even though they have different codes. In this case a dimmer with id 300a, and id 0107 were both identified as "FGD211 Universal Dimmer 500W".  It is now clearer to me why these different lights did not act the same way even though they were meant to be the same unit(s).

Ideally I would prefer to have the id already present in the definition, but I cannot see how to do this with the IOBROKER interface displaying the node list.

Madelaide:
I missed some entries in the previous commit. I think they are all fixed now in the Dev branch.

The master branch does not work on my system as there are too many missing devices (they just some up blank).

Please let me know if there is anything else required to get this pull request completed.

PR #1480, opened 2018-04-14
Configuration for Hank Flood Sensor

cincuranet:


cincuranet:
Although complete in terms of values, I have a question how to handle parameter 20. It has different default values for US (1040) and Other (400). What value to specify?

Issue #1474, opened 2018-04-02
Zipato bulb COMMAND_CLASS_COLOR gone


jeroenst:
Since bug https://github.com/OpenZWave/open-zwave/issues/1457 was solved, the COMMAND_CLASS_COLOR is not added anymore to the zwave config cache file causing it that colors not working anymore.

I'm aware that this issue is caused by my own bugfix in the past for the missing bluechannel in the color value, but I have not enougd knowledge to solve this bug.

When checking out the commit before the blue value fix everything works again after deleting the ozw cache file (except the blue color channel). When checking out after the commit and removing the ozw cache file it is not possible to read or control the color anymore.

Here is the part of the Zipato node in the cachefile where you clearly can see the COMMAND_CLASS_COLOR is not saved:

         <Node id="11" name="" location="" basic="4" generic="17" specific="1" type="Multilevel Power Switch" listening="true" frequentListening="false" beaming="true" routing="true" max_baud_rate="40000" version="4" query_stage="Probe">
                <Manufacturer id="0" name="">
                        <Product type="0" id="0" name="" />
                </Manufacturer>
                <CommandClasses>
                        <CommandClass id="32" name="COMMAND_CLASS_BASIC" version="1" mapping="38">
                                <Instance index="1" />
                        </CommandClass>
                        <CommandClass id="38" name="COMMAND_CLASS_SWITCH_MULTILEVEL" version="1">
                                <Instance index="1" />
                                <Value type="byte" genre="user" instance="1" index="0" label="Level" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="255" value="0" />
                                <Value type="button" genre="user" instance="1" index="1" label="Bright" units="" read_only="false" write_only="true" verify_changes="false" poll_intensity="0" min="0" max="0" />
                                <Value type="button" genre="user" instance="1" index="2" label="Dim" units="" read_only="false" write_only="true" verify_changes="false" poll_intensity="0" min="0" max="0" />
                                <Value type="bool" genre="system" instance="1" index="3" label="Ignore Start Level" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="True" />
                                <Value type="byte" genre="system" instance="1" index="4" label="Start Level" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="255" value="0" />
                        </CommandClass>
[...]

PR #1490, opened 2018-04-25
Fix of bug #1474 (Color class didn't work anymore on zipato bulb 2)

jeroenst:


nechry:
the zipato bulb 2 work without this change. and set the channel in config will not apply to the device. this PR seem invalid for my point of view,

gizmocuz:
I can confirm it works here to without this patch

jeroenst:
Possible you have newer firmware in your bulb. The bug is fixed in later firnware zipato told me. But afaik it is not possible to update firmware with openzwave

BramJ85:
My 3 Zipato Bulb2's showed no RGBW control (in Domoticz), adding the changes described by @jeroenst to my RGBBulb2.xml fixed my issues 

maybe this issue should be looked at again?

blintwave:
Think it has something to do within Domoticz, running Hass.io i have no problems with Zipato bulb 2 only that i have to restart the instance  a few times to get colors to work but after that it works without any issues. I even have tested this bulb with Openhab and everything works except that dimming sometimes works unexpected. So I think it depends how some domotica software implents openzwave in their own instance, but currently from my point of view this bulb works the best with home assistant.

gizmocuz:
I tested with domoticz and have no issued to use the Zipato Bulb version 2

blintwave:
Out of the box? Or edited with RGBBulb2.xml?

BramJ85:
On another forum I found some people with the same problem as me, at least one fixed it with the editied xml.

@jeroenst mentioned that Zipato said something about diffirent firmware versions? is there a way to figure out what firmware version is in the bulb?

voggur:
Hi I recently got my fist smart-home appliance and dust got two Zipato bulb 2 Model "rgbw2.eu" datecode "1807" and I can not get it do anything other than on/off and dimm. I have tried editing RGBBulb2.xml with no luck

PR #1492, opened 2018-04-29
Initial NetBSD port

tih:
These are the changes I had to make to get open-zwave working (with domoticz) under NetBSD/amd64-current. There's just a very small handful:

- help the build system find libusb-1.0, so it can build the USB bits
- correct for a command line option difference in the fmt(1) utility
- drop a call to an unimplemented phtreads function (the call just sets what's default, anyway, so nfc)
- open the USB serial device O_NONBLOCK, so the call to open(2) doesn't hang

With these changes, domoticz happily uses my Aeotec and Z-Wave.Me USB sticks.

### DIFF
--- a/cpp/build/Makefile
+++ b/cpp/build/Makefile
@@ -57,6 +57,9 @@ endif
 endif
 # For 10.2 and later, use iconv from base, no extra include path required.
 
+else ifeq ($(UNAME),NetBSD)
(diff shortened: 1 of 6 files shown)


Fishwaldo:
Ok. I don't see this breaking existing platforms, so should be ok to merge. Thanks!

PR #1495, opened 2018-04-30
Fix issue with $pkgconfigdir not being used as an absolute path

DvdGiessen:
In the Makefile during the install step, `$pkgconfigdir` is prefixed with `$DESTDIR`. However, `$pkgconfigdir` is an absolute path already and this causes the package configuration files to be placed in an incorrect directory (aka, not recognized by `pkg-config`) whenever the `$DESTDIR` is not `/`. This is for example the case when cross-compiling.

This change fixes the install step for such scenarios without any impact to builds targetting `/`.


Fishwaldo:
I can't remember why I did this... but it seems sane enough. Anything not in the standard pkg-config path would be ignored anyway... so I'll merge this change and see what, if anything breaks!

PR #1507, opened 2018-05-10
Added support for MinGW-W64

Broekman:
1. cpp/src/build/windows/mingw-w64/Makefile - added a Makefile with support for MinGW-W64.
2. cpp/src/Defs.h:38 - added support for MinGW-W64.
3. cpp/src/platform/windows/LogImpl.cpp:84 - switched two values in the initializer list to fix a -Wreorder warning.
4. cpp/src/platform/windows/FileOpsImpl.cpp:62 - changed "int32 ftype" to "uint32 ftype" to match the (unsigned) DWORD return value of the "GetFileAttributesA" function to fix a -Wsign-compare warning.

### DIFF
--- a/cpp/build/windows/mingw-w64/Makefile
+++ b/cpp/build/windows/mingw-w64/Makefile
@@ -1,0 +1,108 @@
+#
+# Makefile for OpenZWave: MinGW-W64 build
+# Stefan Broekman
+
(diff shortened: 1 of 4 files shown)


Fishwaldo:
Thanks

PR #1494, opened 2018-04-30
Added Heltun HE-ZW-THERM-FL2 Thermostat

petterhanssen:
Added support for a new thermostat ive bought. First time working with zwave and the openzwave repository, so just let me know if theres anything im missing. 

Thermostat: https://www.heltun.com/z-wave-heating-thermostat
Documentation: https://drive.google.com/file/d/1AKHXZHyFCfpo_2Wd0XrisgIwilpezM7Q/view

PR #1508, opened 2018-05-14
properly install pkgconfig files to DESTDIR

melvinvermeeren:
currently it always tries to install to the system root, completely
ignoring DESTDIR. this breaks packaging on Arch Linux

https://aur.archlinux.org/packages/openzwave-git/


Fishwaldo:
See #1494 and #1509 

Reverted the change that broke this.

melvinvermeeren:
Thank you for the quick fix. :)

PR #1513, opened 2018-05-19
Fixes line 312 build

asm0dey:
Right now I have following error:

```
Building DoorLockLogging.o
/tmp/finkel/trizen-finkel/openzwave/src/open-zwave-1.5/cpp/src/command_classes/DoorLockLogging.cpp: In member function ‘virtual bool OpenZWave::DoorLockLogging::HandleMsg(const uint8*, uint32, uint32)’:
/tmp/finkel/trizen-finkel/openzwave/src/open-zwave-1.5/cpp/src/command_classes/DoorLockLogging.cpp:312:15: error: passing argument 1 to restrict-qualified parameter aliases with argument 4 [-Werror=restrict]
      snprintf(usercode, sizeof(usercode), "%s %d", usercode, (int)_data[12+i]);
               ^~~~~~~~                             ~~~~~~~~
cc1plus: all warnings being treated as errors
```

### DIFF
--- a/cpp/src/command_classes/DoorLockLogging.cpp
+++ b/cpp/src/command_classes/DoorLockLogging.cpp
@@ -309,7 +309,9 @@ bool DoorLockLogging::HandleMsg
 			if (usercodelength > 0)
 				for (int i = 0; i < usercodelength; i++ )
 				{
-					snprintf(usercode, sizeof(usercode), "%s %d", usercode, (int)_data[12+i]);
+					char* tmp = strndup(usercode, sizeof(usercode));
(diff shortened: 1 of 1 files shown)


stanford-scs:
Note that usercode always contains the string "Usercode:" in this context, so a simpler fix is just:

    snprintf(usercode, sizeof(usercode), "UserCode: %d", (int)_data[12+i]);

What would it take to get this patch upstreamed?  I'm having a horrible time getting python-openzwave to install, I think because it keeps trying to pull and build the C++ code (even in the shared flavor), so just fixing this locally doesn't seem to work.

Fishwaldo:
Fix already applied.

PR #1516, opened 2018-05-22
Make HID support optional (and disabled by default)

### LINKED ISSUE #1316: Continue HID support
I would like to ask this question again as there is not so many feedback from the ozw core developers on the forum here:

https://groups.google.com/forum/#!searchin/openzwave/HID%7Csort:relevance/openzwave/eFge7COuY9w/Z1o52C7AAwAJ

Is this OK ?

My idea is to make a Preprocessor (HID_SUPPORT) that is default enabled in the make script, but can be turned off.

The last week(s) i am struggling making a binary compile that is compatible between armv7 and armv6 with the new debian release.
Removing the HID dependency solves many issues.

Also for those users using a HID controller (and i think there are almost none), for a few dollars they can buy a usb stick these days

dwmw2:
This allows us to eliminate the libudev dependency.

Fixes #1316


dwmw2:
Ping @ratkaj @stintel — this is based on what we have in OpenWRT at https://github.com/openwrt/packages/tree/master/utils/openzwave/patches

gizmocuz:
@Fishwaldo , this is the patch i was referring almost to a year ago.
Is it possible to have a look at this ? I think it is a good idea to have the HID disabled by default as it causes a lot of people that needed to use a patch for this now.
Not sure if HID zwave sticks are still sold on the marked these days, specially when zwave+ was introduced.
 But with this patch those users can enable it when needed

Fishwaldo:
Let me think a while if this should go into master, or the Dev branch... agree that there probably isn't any use of HID anymore, but maybe some users with legacy controllers. 

Pulling this into master might cause them some "grief" if they don't know the change has happened.

dwmw2:
Perhaps just make it optional in master but leave it on by default there.-- dwmw2On 27 Jul 2018 05:37, Justin Hammond <notifications@github.com> wrote:Let me think a while if this should go into master, or the Dev branch... agree that there probably isn't any use of HID anymore, but maybe some users with legacy controllers.
Pulling this into master might cause them some "grief" if they don't know the change has happened.

Fishwaldo:
Done. Thanks!

PR #1525, opened 2018-06-04
tz74.xml

dav76df:
add switch wenzhou tz74


nechry:
Please use list value type for this kind of parameter this is more convenant.
Also do in same PR the manufacturer_specific.xml update.

PR #1527, opened 2018-06-08
Add missing #ifdef __NetBSD__ for a platform specific cast

tih:
When I created the (since merged) pull request #1492 to make open-zwave build on NetBSD, I accidentally omitted to enclose one platform specific change in an ifdef. @ryanwinter points out that this breaks the build on FreeBSD.  This change corrects the omission.

### DIFF
--- a/cpp/hidapi/libusb/hid.c
+++ b/cpp/hidapi/libusb/hid.c
@@ -369,7 +369,11 @@ static wchar_t *get_usb_string(libusb_device_handle *dev, uint8_t idx)
 	inbytes = len-2;
 	outptr = (char*) wbuf;
 	outbytes = sizeof(wbuf);
+#ifdef __NetBSD__
 	res = iconv(ic, (const char ** restrict)&inptr, &inbytes, &outptr, &outbytes);
+#else
+	res = iconv(ic, &inptr, &inbytes, &outptr, &outbytes);
+#endif
 	if (res == (size_t)-1) {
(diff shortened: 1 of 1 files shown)


cpp/hidapi/libusb/hid.c
```diff
@@ -369,7 +369,11 @@ static wchar_t *get_usb_string(libusb_device_handle *dev, uint8_t idx)
 	inbytes = len-2;
 	outptr = (char*) wbuf;
 	outbytes = sizeof(wbuf);
+#ifdef __NetBSD__
```
krytarowski: The usual solution is to ifdef __NetBSD__ define ICONV_CONST const else define ICONV_CONST endif

tih: And putting the ifdef in some widely (preferably globally) included header file, then?

krytarowski: It's usually used in a few places anyway. If this is just a single function it can be inlined in the function body.

tih: In this case, it's just this one use, in a source file that's been copied into this project from elsewhere. I guess if one were to do as you suggest, it would be either right next to the single use, or up near the top of hid.c, between the DEBUG_PRINTF definition and the FreeBSD DETACH_KERNEL_DRIVER stuff.

I'll keep it close, for now, I think.

PR #1529, opened 2018-06-10
Use pkgconf to find the local package directories on NetBSD

tih:
When I submitted the modifications for building on NetBSD, in pull request #1492, I hardcoded a couple of paths to pkgsrc installation directories in a Makefile. Having had this pointed out to me by @krytarowski, I've changed to the more correct use of our pkgconf tool. (/usr/pkg is the default target, but sysadmins may choose to have pkgsrc install in other locations.)


Fishwaldo:
Thanks

PR #1530, opened 2018-06-11
Fix parameter size and allowed values for HS-WD100+

kaburke:
The changes allow the parameters to be set to their full allowable range of values, as specified in the documentation:
https://homeseer.com/wp-content/uploads/2018/06/HS-WD100-Manual-6a.pdf
The documentation specifies the size of parameters 7 and 9 as 1 byte, however defining them as bytes prevented OZW from updating the values: the values would always reset to 1, regardless of the supplied value. Using short works.


nechry:
thank's @kaburke

nechry:
sorry please rebase on master branch

PR #1535, opened 2018-06-15
Create TZWP102.xml

flopp999:
link to manufactor manual
http://download.telldus.com/Manuals/Telldus/TZWP-102/TZWP-102_technical-manual.pdf


nechry:
also change type for param 101,102,103 and 104 with int for 4 bytes length
https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices#configuration-variable-types

flopp999:
i have now added TZWP-102 in manufacturer_specific.xml
also changed from byte to INT

PR #1539, opened 2018-06-20
Update fgwpe.xml

flopp999:
Change all Default settings to a text instead of just a number, which the user never will see.


flopp999:
Sorry I by mistake PR my master, can you still use the changes?
can a Short type have min and max values?
Changes taken from this link
https://manuals.fibaro.com/content/manuals/en/FGWPx-102/FGWPx-102-EN-A-v1.00.pdf

nechry:
yes short type can have min and max values.

PR #1543, opened 2018-06-27
adding Steinel XLED home 2 device

ToSa27:
For some reason dist-update introduced two more changes (two additional files). Maybe the dist-update was missed for earlier updated. I kept them in the pull request for now even though they are not related. Let me know if I should submit a clean version of the pull request without these two changes (means I would need to manually adjust the distfiles after running dist-update).

Device: [https://products.z-wavealliance.org/products/2290/](https://products.z-wavealliance.org/products/2290/)


nechry:
thank's @ToSa27, I'm thinking in the moment to buy this equipment, is good?

PR #1545, opened 2018-06-29
Renamed thermofloor z-trm2 config file following docs specs

robertsLando:


nechry:
@robertsLando please patch also distfiles.mk
https://github.com/OpenZWave/open-zwave/blob/master/distfiles.mk#L449
to avoid break automatic build

robertsLando:
@nechry Now should be ok, sorry for the mistake :)

nechry:
thanks @robertsLando

PR #1547, opened 2018-06-30
Added ramp rate control option to zen22v2.xml

cby016:
The documentation that came with my zooz zen22 version 2 light indicates this option should exist. I've added it on my own setup and it is working as described. Quote from the documentation:

> **Ramp Rate Control**
> Parameter 4: Choose from 2 ramp rate options (fade-in / fade-out effect for on / off operation)
> Values: 0 - light turns on to 100% brightness within 2 seconds, turns off within 4 seconds (default); 1 - light turns on to 100% brightness instantly, turns off within 1 second
> Size: 1 byte dec.

### DIFF
--- a/config/zooz/zen22v2.xml
+++ b/config/zooz/zen22v2.xml
@@ -17,6 +17,11 @@
       <Item value="0" label="Enabled"/>
       <Item value="1" label="Disabled"/>
     </Value>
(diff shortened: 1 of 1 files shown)


nechry:
thanks @cby016

PR #1550, opened 2018-07-05
support building on SmartOS

bahamas10:
relevant issue: https://github.com/OpenZWave/python-openzwave/issues/123

### DIFF
--- a/cpp/build/Makefile
+++ b/cpp/build/Makefile
@@ -60,6 +60,9 @@ endif
 else ifeq ($(UNAME),NetBSD)
 CFLAGS += -I/usr/pkg/include/libusb-1.0
 LDFLAGS+= -shared -L/usr/pkg/lib -lusb-1.0 -Wl,-rpath=/usr/pkg/lib
+else ifeq ($(UNAME),SunOS)
+CFLAGS+=  -I$(PREFIX)/include/libusb-1.0
+LDFLAGS+= -shared -lusb-1.0 -liconv
 else
 LDFLAGS += -shared -Wl,-soname,$(SHARED_LIB_NAME)
 LIBS 	+= -ludev
@@ -76,6 +79,8 @@ SOURCES_HIDAPI =$(top_srcdir)/cpp/hidapi/mac
 else ifeq ($(UNAME),FreeBSD)
 SOURCES_HIDAPI =$(top_srcdir)/cpp/hidapi/libusb
 else ifeq ($(UNAME),NetBSD)
+SOURCES_HIDAPI =$(top_srcdir)/cpp/hidapi/libusb
+else ifeq ($(UNAME),SunOS)
 SOURCES_HIDAPI =$(top_srcdir)/cpp/hidapi/libusb
 else
 SOURCES_HIDAPI =$(top_srcdir)/cpp/hidapi/linux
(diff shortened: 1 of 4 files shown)


cpp/build/Makefile
```diff
@@ -23,6 +23,9 @@ top_srcdir := $(abspath $(dir $(lastword $(MAKEFILE_LIST)))../../)
 
 include $(top_srcdir)/cpp/build/support.mk
 
+# Used to detect SmartOS (variable will be "joyent")
+UNAME_V := $(shell uname -v 2>/dev/null | sed -e 's/_.*//')
```
Fishwaldo: Why can't you use UNAME instead?

bahamas10: The issue here is that `UNAME` on SmartOS overlaps with other operating systems.

```
$ uname 
SunOS
$ uname -v
joyent_20180510T153535Z
```

This will overlap with Solaris, as well as other [illumos](http://illumos.org/) based operating systems.  The issue here is that using `/opt/local` as a prefix is something that is fairly specific to SmartOS.  Just detecting `UNAME == SunOS` would work for the SmartOS case, but would ultimately fail for Solaris or other Illumos based operating systems if support for them were to ever be considered in the future.

bahamas10: I'm testing a patch right now that won't hardcode `/opt/local` and will also not rely on `uname -v` - it should just work with UNAME instead.  I'll know more next week when I can test it better.

bahamas10:
I've updated this commit to use `UNAME` like the rest of the Makefile.  It now no longer hardcodes paths and instead requires the user to pass in the PREFIX if they are building on a SunOS platform that doesn't user `/usr/local` by default.  For example, to build on SmartOS, I would run:

   PREFIX=/opt/local make

This solution is way better than the code I had before... but still not *perfect* in my opinion.

cpp/build/Makefile
```diff
@@ -60,6 +60,9 @@ endif
 else ifeq ($(UNAME),NetBSD)
 CFLAGS += -I/usr/pkg/include/libusb-1.0
 LDFLAGS+= -shared -L/usr/pkg/lib -lusb-1.0 -Wl,-rpath=/usr/pkg/lib
+else ifeq ($(UNAME),SunOS)
+CFLAGS+=  -I$(PREFIX)/include/libusb-1.0
```
bahamas10: This is the only part of the SunOS specific code that requires the PREFIX (or `/opt/local`) variable to be set.  There are a couple alternative approaches that we could use here - I'm open to any of them as I don't personally know which one is preferred or "best".

### 1. Use `pkg-config`.

`pkg-config` is an optional dependency AFAICT with this project so I'd hate to enforce it's presence - though it is incredibly useful in this specific situation.

```
[root@test ~]# uname -a
SunOS test.rapture.com 5.11 joyent_20180510T153535Z i86pc i386 i86pc Solaris
[root@test ~]# pkg-config libusb-1.0 --cflags
-I/opt/local/include/libusb-1.0
```

That will give us the CFLAGS line we need for libusb-1.0

### 2. Use a relative `#include`

Inside `cpp/hidapi/libusb/hid.c`, the include line could be modified to be:

``` c
#if defined(__sun)
#include "libusb-1.0/libusb.h"
#else
#include "libusb.h"
#endif
```

The bonus here is that we could also add `__NetBSD__` to the `defined()` check to the drop it's custom CFLAGS in the lines above.

### 3. Using `$PREFIX`

This is what is currently used in this PR

Fishwaldo:
I'll merge this... if you want to make further changes... open a new PR :)

PR #1562, opened 2018-07-27
Updated manufacture ID.xml and add four configuration xml.

zl0217:
Updated manufacture ID.xml and add four configuration xml.


config/FANTEM/FT080.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- 
+FATEM FT080 Siren
+-->
+<Product xmlns='http://code.google.com/p/open-zwave/'>
+	<Protocol refreshonnodeinfoframe="false"/>
```
Fishwaldo: is this really required?

zl0217: Hi Fishwaldo, nice to meet you, what do you mean that this expression Protocol refreshonnodeinfoframe="false" is required? if yes, I think it is required, and I want to know if we config like this, could you please tell me does it cause any problems?
By the way, our customers have purchased many products which are produced by Fantem, and they complained that these products can't be compatible with the gateway based OpenZwave, so could you please help us handle this imminently, thanks very much.

Fishwaldo:
I can't tell you - As I don' thave this device. It would be unusual to require this (as its for older buggy devices). Have you tested without that line?

Fishwaldo:
ping!

PR #1565, opened 2018-08-01
Updating collection of Inovelli devices 

InovelliUSA:
Adding some additional devices and fixing just a couple of configuration parameters / adding association groups.


Fishwaldo:
Thanks

PR #1567, opened 2018-08-01
GR Multichannel Switches

AlbertoLopSie:
This PR supports the U-Fairy G.R. IOT Tech (http://www.gr-smarthome.com/en/) Multichannel switch Module Embedded 3 gang Switch (http://www.gr-smarthome.com/en/h-pd-46.html#_pp=0_304_3_-1) to be used in OZW and (in my case) on Home Assistant through python-openzwave

That buggy chinese switches incorrectly reports as 0 the COMMAND_CLASS_MULTICHANNEL/MULTIINSTANCE version and causes the V1 of the CC to be used by OZW. The module doesn't support that (after reading the spec I'm not sure if it should, though) and ignores the On/Off commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ ozw_config
 .idea/
 .idea/*
 ozw_config
+
+cpp/src/command_classes/\.DS_Store
--- a/ChangeLog
+++ b/ChangeLog
@@ -256,6 +256,7 @@ Dev Branch:
  - dotNet Wrapper is now maintained by dotMorten https://github.com/OpenZWave/openzwave-dotnet-uwp (Justin)
  - Exceptions are now enabled by Default (Justin)
  - Depreciate the builtin Scene Methods (not related to CENTRAL_SCENE) (Justin)
+ - Fix a build error on OSX - Issue #1593 (Justin)
 
 Version 1.4
  - Released 10th Jan, 2016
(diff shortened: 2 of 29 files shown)


cpp/src/command_classes/MultiInstance.cpp
```diff
@@ -147,6 +148,22 @@ void MultiInstance::ReadXML
 	{
 		m_uniqueendpoints = !strcmp( str, "true");
 	}
+	str = _ccElement->Attribute("version");
```
Fishwaldo: I think this should be done in the CommandClass class then it can be used for any other CC's that might have this issue. Can you move this?

AlbertoLopSie:
No problem... I agree.

Being my first PR on this project I didn't want to introduce that change in a class so deep in the hierarchy  that could affect code I'm not fully aware of.

Anyway... I'll move the fix to the CommandClass class in a few minutes.

AlbertoLopSie:
Requested changes made.

AlbertoLopSie:
Hi Justin!  Is there anything else I have to do before you are able to merge this PR or you just are waiting for the proper time to do it?

Fishwaldo:
Hrm. Only other things I'm thinking off (but I'll try to add it myself) is not allow the version to be set Higher than the MaxVersion the CC supports.

PR #1576, opened 2018-08-10
Fix for issue #997.

osenft:
Add support to clear UserCode by writing empty value or value consisting of NULLs only.

### DIFF
--- a/cpp/src/command_classes/UserCode.cpp
+++ b/cpp/src/command_classes/UserCode.cpp
@@ -298,7 +298,7 @@ bool UserCode::HandleMsg
 
 //-----------------------------------------------------------------------------
 // <UserCode::SetValue>
-// Set a User Code value
+// Set a User Code value or clear a User Code by writing empty or NULLs-only value
 //-----------------------------------------------------------------------------
 bool UserCode::SetValue
 (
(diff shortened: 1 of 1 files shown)


osenft:
Tested against a Schlage BE469 Touchscreen Deadbolt as follows:
- Built python-openzwave with Python 3.6 using patched openzwave code.
- Copied into Home Assistant and restarted HASS.
- Executed lock.get_usercode on populated and unpopulated slots.
- Executed lock.set_usercode on unpopulated slot. Checked correct result via lock.get_usercode.
- Executed lock.clear_usercode on populated slot. Checked correct result via lock.get_usercode.
- Manually confirmed on physical lock that code from cleared slot no longer unlocks the deadbolt lock.

Fishwaldo:
Please see my reply to #997

osenft:
See comments on #997. Are we sure the proposed solution is not sufficient?

edif30:
I have tested this PR and seems to work well on my Schlage BE469.  Slot shows (Available) after clearing.

edif30:
@Fishwaldo based on your comments...

> #1576 makes the assumption that the average joe knows that 0000 or '' or NULL would clear the the code. With the above approach (as most applications iterate over available ValueID's for a command class) they would expose a new Input Box or whatnot that is called "Reset Code" or whateverand wouldn't be required to know the "special" code to clear a code. (other than the index value of what they want to clear).

I don't know that we are going to solve this problem from a standpoint of who knows about special codes or not.  However I do think that anyone with knowledge of the 0000 or NULL can use this patch to their advantage.  This would be the case more so than not.  Local clearing of codes still work as well.  So I don't think this would break anything either.  If anything, it at least improves things for certain users.

Fishwaldo:
Exactly my point. If we just create a new ValueInt called "Clear User Code" then an application just has to present a user with an Input Box (or drop down list or whatever). When we get a Set on that ValueID, we clear the code internally in the class. 

Then the user doesn't have to care if its ascii 0000, decimal 0000, Null etc to clear the code. Devices with quirks, we can handle with a config option inside the class. 

Will applications need to be updated? Yes - But there are lots of updates in the Dev branch, so this is just another change applications will need to take care of over 1.4

Fishwaldo:
I'm working on refactoring this already. So Closing this. Watch #997

PR #1584, opened 2018-08-20
Is group multi instance 

Ryszard-Trojnacki:
Group in ZWave can have multi instance availability or not.
This is correctly programmed in Group.h/Group.cpp but it is not available to end user (user of OpenZWave library).
I have created a patch, that publishesh Group.h `IsMultiInstance` to Manager.h:
```c++
/**
 * \brief Returns true is group supports multi instance.
 * \param _homeId The Home ID of the Z-Wave controller that manages the node.
 * \param _nodeId The ID of the node whose associations we are interested in.
 * \param _groupIdx one-based index of the group (because Z-Wave product manuals use one-based group numbering).
 * \return True if group supports multi instance.
 * \see GetNumGroups, AddAssociation, RemoveAssociation, GetAssociations
 */
bool IsMultiInstance( uint32 const _homeId, uint8 const _nodeId, uint8 const _groupIdx );
```

### DIFF
--- a/config/homeseer/hs-ls100plus.xml
+++ b/config/homeseer/hs-ls100plus.xml
@@ -38,15 +38,15 @@
         <Value type="short" index="20" genre="config" label="High temperature trigger value" units="" min="-670" max="2570" size="2" value="1040">
             <Help>
                 Set the high temperature alarm trigger value.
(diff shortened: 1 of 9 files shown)


Fishwaldo:
Could you please rebase against Dev branch please. Master is for bugfixes only.

PR #1566, opened 2018-08-01
added Multilevel Sensor types and scales for version 6-11

brouwer:
My first pull request here so let me know if I'm doing something wrong. Added 37 Multilevel Sensor Types with corresponding scales according to the Z-Wave specification for Multilevel Sensor Command Class. 

http://zwavepublic.com/sites/default/files/command_class_specs_2017A/SDS13812-2%20Multilevel%20Sensor%20Command%20Class%2C%20list%20of%20assigned%20Multilevel%20Sensor%20types%20and%20scales.pdf

PR #1585, opened 2018-08-21
Support particulate matter 2.5

jviki:
The purpose of this patch series is to add support of PM2.5 to OpenZWave. To achive this, 3 more SensorMultilevel types had to be added as the SensorType is sequential.


Fishwaldo:
Already added in #1566

PR #1598, opened 2018-09-12
Create BE369.xml

smokey1300:
Schlage file for older keypad style lock model BE369


nechry:
please add the manufacturer_specific.xml file with relative products id

tehspaceg:
I know this is an ANCIENT issue, but I was able to use this BE369.xml on my HASS implementation running the OZW 1.4 version.

I added this to the manufacturer_specific.xml file under the Schlage section.

`<Product type="634b" id="5044" name="BE369 Z-Wave enabled lock" config="schlage/BE369.xml"/>`

I'm brand spanking new to the OZW world, so please let me know if I should do a new PR. I also have no idea if this will work on OZW 1.6.

PR #1599, opened 2018-09-14
Drop a number of unused variables

stromnet:
On FreeBSD 11.2 (clang 6.0) build fails due to unused instance variables

```
In file included from /usr/home/johan/dev/open-zwave/cpp/src/value_classes/ValueList.cpp:29:                      
/usr/home/johan/dev/open-zwave/cpp/src/value_classes/ValueList.h:85:11: error: private field 'm_newValueIdx' is not used [-Werror,-Wunused-private-field]            
                int32                   m_newValueIdx;                          // a new value to be set on the appropriate device
                                        ^                                       
1 error generated.                                                                      
```

and similar.

This MR removes the unused variables in the Value classes, and makes use of it in the Group.cpp class (seems sensible to keep m_length together with raw byte byffer m_data)

### DIFF
--- a/cpp/src/Group.cpp
+++ b/cpp/src/Group.cpp
@@ -539,8 +539,8 @@ Group::AssociationCommand::AssociationCommand
 ):
 	m_length( _length )
 {
-	m_data = new uint8[_length];
-	memcpy( m_data, _data, _length );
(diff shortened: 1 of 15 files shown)


Fishwaldo:
Duplicate of #1580

PR #1606, opened 2018-09-26
Fix broken build process for macOS

MasterKale:
I attempted to build OZW on macOS Mojave (10.14) but was blocked by a couple of issues with the **Darwin**-specific build process:

- Xcode's **make** no longer supports the `i386` architecture. I've removed these build targets from the OZW library as well as the MinOZW example application
- The build failed on several unused private fields defined within the various Value classes. The `-Wunused-private-field` gcc flag was being implicitly set somewhere (I couldn't identify where), but I was able to add an additional flag to ignore such errors.

I can confirm that, with these changes, OZW can now be built and installed on macOS 10.14 with a simple `make && sudo make install`.

### DIFF
--- a/cpp/build/Makefile
+++ b/cpp/build/Makefile
@@ -34,13 +34,15 @@ endif
 
 #if we are on a Mac, add these flags and libs to the compile and link phases 
 ifeq ($(UNAME),Darwin)
+CFLAGS	+= -c -DDARWIN
 ifeq ($(DARWIN_MOJAVE_UP),1)
 # Newer macOS releases don't support i386 so only build 64-bit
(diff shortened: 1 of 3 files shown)


Fishwaldo:
This would break for older OSX versions (that might still need i386 fat binaries). Can you update to anything above 10.14 doesn't build with i386 instead please.

MasterKale:
I've updated the code to only set the modified build flag and only target x86_64 on macOS 10.14 and up. This should better-preserve the project's current ability to be built on older versions of the OS.

Fishwaldo:
Thanks

PR #1610, opened 2018-09-28
Fixes an access violation on the NextSibling after a child has been removed

### LINKED ISSUE #1609: Access violation when refreshing a node
When calling RefreshNodeInfo on a node that is not last in the list, an Access Violation is thrown at.

```
>	OpenZWave.dll!OpenZWave::Driver::ReloadNode(const unsigned char _nodeId) Line 7300	C++
 	OpenZWave.dll!OpenZWave::Manager::RefreshNodeInfo(const unsigned int _homeId, const unsigned char _nodeId) Line 785	C++
 	OpenZWave.dll!OpenZWave::ZWManager::[OpenZWave::__IZWManagerPublicNonVirtuals]::RefreshNodeInfo(unsigned int homeId, unsigned char nodeId) Line 385	C++
```
image

This occurs because when the node is found, the node is removed, and the while-loop continues. However when the NextSibling is called on the now-removed node, it returns an invalid memory location and the next ToElement() will throw:

https://github.com/OpenZWave/open-zwave/blob/39695691bc6fe04c24ede7b037782c52b7c286da/cpp/src/Driver.cpp#L7298-L7313

There's two solutions here: The obvious one is to break out of the loop once the node has been found, the second is to get the nextsibling before removing the node (but I'm assuming there's no point in continuing).

The issue has been observed on UWP and Win32 (not sure about any other platforms), but I'm guessing the while check there might be handling this slightly different.

To reproduce run the Windows Form or UWP app in https://github.com/OpenZWave/openzwave-dotnet-uwp/tree/master/src
Right-click the node in WinForms or if UWP select a device and click the settings-icon in the upper-right in UWP and then select "Refresh Node":

image
image

I confirmed adding a break inside this if-check resolves the issue:
https://github.com/OpenZWave/open-zwave/blob/39695691bc6fe04c24ede7b037782c52b7c286da/cpp/src/Driver.cpp#L7307-L7308

Ref: https://github.com/OpenZWave/openzwave-dotnet-uwp/issues/22

dotMorten:
Also there's no reason to continue the loop once the node has been found and removed.
Fixes #1609 / https://github.com/OpenZWave/openzwave-dotnet-uwp/issues/22

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -7304,8 +7304,10 @@ void Driver::ReloadNode
 					// Get the node Id from the XML
 					if( TIXML_SUCCESS == nodeElement->ToElement()->QueryIntAttribute( "id", &intVal ) )
 					{
-						if (intVal == _nodeId)
+						if (intVal == _nodeId) {
 							driverElement->RemoveChild(nodeElement);
+							break;
+						}
 					}
 				}
 			}


Fishwaldo:
Thanks!

PR #1612, opened 2018-10-02
Fix GetLabel for values

dotMorten:
There was some mixup between Help and Label strings, causing GetLabels to not work for most values (SetHelp was used instead of SetLabel when loading the XML)

/cc @Fishwaldo


Fishwaldo:
Thanks!

PR #1619, opened 2018-10-04
Adding device id and type for YKFCON with ZW+/S2

Dinth:
Hi. Ive got a Yale YKFCON lock equipped with the new (dark blue) Zwave+/S2 module and it's device type and id doesnt match the config file, so i amended it.
Sorry if not following the process, im a complete github noob.


nechry:
thank @Dinth

PR #1624, opened 2018-10-09
Fixed Z wave plus command class name

cstrachan88:
The correct command class name for z-wave plus, according to [this wiki](http://wiki.micasaverde.com/index.php/ZWave_Command_Classes) and other official sources, is `COMMAND_CLASS_ZWAVEPLUS_INFO` instead of `COMMAND_CLASS_ZWAVE_PLUS_INFO`.

I updated all of the instances where I found this in the repo.


Fishwaldo:
Thanks

PR #1628, opened 2018-10-17
Heatit Z-TRM2 Update for firmware v2.11

CookieMonster87:
- Fixed parameter 12 from short (not working) to byte. 
- Brightness settings changed from short to byte.
- Association Group Internal relay changed from 2 to 5.
- Association Group Multilevel sensor reports - Internal sensor changed from 3 to 2.
- Association Group Multilevel sensor reports - External sensor changed from 4 to 3.
- Association Group Multilevel sensor reports - Floor sensor changed from 5 to 4.

This product is released with non Z-Wave alliance certificated firmware (2.6) (sale stop until final approved version is released) All Z-TRM2 thermostats will need to update firmware in the future and new ones will be shipped with fw2.11 or later.


nechry:
thanks @CookieMonster87

PR #1630, opened 2018-10-19
Config file for HomeSeer FC200+ Fan Controller

kpine:
New config xml file for the HomeSeer FC200+ Fan Controller: https://shop.homeseer.com/collections/lighting/products/homeseer-hs-fc200-z-wave-plus-fan-controller

Includes the central scene command class. This was initially grabbed from the Home Assistant help page for device-specific configuration: https://www.home-assistant.io/docs/z-wave/device-specific/#homeseer-switches. The scene data that is shown in the OZW log matches the "Top button ID: 1, Bottom ID: 2" in the HomeAssistant docs.

The config parameter help text is copied nearly verbatim from the HS user manual, so it does differ slightly from the other HS switches.


nechry:
thanks @kpine

PR #1642, opened 2018-10-27
Configuration update for Linear GoControl GC-TBZ48

eyager1:
Add missing configuration parameter 23.   This parameter changes what information is reported automatically reported by the thermostat upon a change in its state.  This will allow a fix of the known issue of the thermostat by default not reporting a change in its operating state and the start and end of a call for heat/cool.

Add Association Groups.


nechry:
thank @eyager1

PR #1659, opened 2018-12-01
Fix Qubin PWM Thermostat ThermostatSetpoint

### LINKED ISSUE #1319: Can't set Qubino PWM thermostat temperature
Hello,
I'm trying to set the temperature for Qubino PWM thermosat but it just rollbacks to the default (23) immediately. Even though the thermostat mode is set to "Heat" it still says "Cooling" in the screenshot. When I change the mode to "Cooling", I can set the temperature to whatever I want. After switching the mode back to Heat the temperature stayed. Which is not so bad, but still it's a bug.

screen shot 2017-09-12 at 3 55 41 pm

I was told by the Qubino support team that it's an issue with Open ZWave:
> Thank you for reaching out to us regarding this issue. Unfortunately flush PWM and ON/OFF thermostats are not compatible with OPEN Z-Wave. Because when you configure the set point it set it for cooling but it does not set it as heating.

michiil:
Fixes #1319

Without this you can't set a temperature for the thermostat.

### DIFF
--- a/config/qubino/ZMNHLDx.xml
+++ b/config/qubino/ZMNHLDx.xml
@@ -17,7 +17,7 @@
    		</SupportedModes>
 	</CommandClass>
 
-    <CommandClass id="67" override_precision="2"  />
+    <CommandClass id="67" override_precision="2" base="0" />
     
 	<!-- Configuration -->
 	<CommandClass id="112">


nechry:
thank @michiil I hope this will not break the old version...

PR #1661, opened 2018-12-02
Add Oomi multisensor european version

flinkebernt:
First PR, so be gentle.

Added one line to add type "0002" of the Oomi Multisensor unit. I have this unit myself.
Two commits because of indentation error.


nechry:
thank @flinkebernt

PR #1663, opened 2018-12-04
Add YRD220 with id=aa00

jamesthomas128:
It seems that Assa Abloy (Yale) has YRD220's with an ID of aa00, in addition to 0000. I have added this change to my own xml file and it works with my YRD220.


jamesthomas128:
There are other ID's for this device referenced here:
https://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/321


References (Type:Id) | 0002:0000,0002:0209,0002:AA00,0002:FFFF,2022:0209
-- | --

nechry:
thank @jamesthomas128

PR #1670, opened 2018-12-12
add support for tkb tz56s

chrisns:


chrisns:
Hey @nechry thanks for the feedback, I've added associations block and commented link as requested.
I noticed there's some other Wenzhou products without associations and links.
I'll try and find time to fix up and PR separately

nechry:
thank @chrisns I fix a small type

chrisns:
Good catch @nechry - copy/paste fail. Thanks!

PR #1675, opened 2018-12-24
Added min and max properties to all configuration items for Fibaro FGS2x3.xml

reneklootwijk:
The configuration items of type list did not contain a min and max property. This property could be used in a UI to validate the specified value is within the allowed range.


nechry:
I see no reason to have UI validation rules for list items ... but it's ok to define these 2 attributes. (I speak in general)

PR #1691, opened 2019-01-12
update for MH7H by MCO Home

HWL1968:
Thanks to aseelert there is an update with a new mh7h.xml (proposed) which needs some additional definitions in the manufacturer xml.


Fishwaldo:
Thanks!

PR #1692, opened 2019-01-12
Cleans up static variables during Manager destruction.

### LINKED ISSUE #792: Memory Leaks
I've been testing the v1.4 openzwave library debug DLL with an MFC program. You may be aware that there is an issue with MFC causing false memory leaks for static objects in linked DLLs. This is not technically a fault of the openzwave library, but I suggest that all static objects be cleared when the Manager is destroyed to avoid this problem. There's a few that are not. I can provide more info if interested. It's mainly the result of not clearing static std list and map objects..

However, after a number of days of debugging a couple of real memory leaks were discovered in the openzwave library. The leaks are all in the Driver class:
1. The destructor for the Driver class is sending a few notifications about its progress. There are comments in that function that indicate the problems with doing that. Well, the result is that (in my case) the last two notifications never get processed and are not destroyed. I left the very first notification (that driver is shutting down) and that apparently gets processed OK. But I suggest removing ALL notifications from the Driver destructor or putting a suitable wait after the notification to ensure it's processed.
2. The AuthKey and EncryptKey objects are not destroyed on Driver destruction. There's a further problem related to this -- Driver function initNetworkKeys will allocate new AuthKey and EncryptKey objects every time it's called, even if they already exist from a prior call. This will then leak the prior key objects.

mankyd:
Incorporate jmattioni's changes to fix #792.

### DIFF
--- a/cpp/src/Manager.cpp
+++ b/cpp/src/Manager.cpp
@@ -209,9 +209,10 @@ Manager::~Manager
 	while( !m_pendingDrivers.empty() )
 	{
 		list<Driver*>::iterator it = m_pendingDrivers.begin();
-		delete *it;
+	        delete *it;
 		m_pendingDrivers.erase( it );
 	}
+	m_pendingDrivers.clear();
 
 	// Clear the ready map
 	while( !m_readyDrivers.empty() )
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Thanks! Will merge soon.

PR #1695, opened 2019-01-17
Set "Play Ringtone" to write_only in config/aeotec/zw056.xml

jack-oconnor:
Updated zw056.xml to make "Play Ringtone" write_only="true". Get commands after set commands were causing timeouts. The timeouts eventually trigger the node to be marked as dead.

[zw056.log](https://github.com/OpenZWave/open-zwave/files/2770497/zw056.log)

### DIFF
--- a/config/aeotec/zw056.xml
+++ b/config/aeotec/zw056.xml
@@ -21,7 +21,7 @@ https://aeotec.freshdesk.com/helpdesk/attachments/6009584693
                 Value=1-100, Specify the ringtone as default.
             </Help>
         </Value>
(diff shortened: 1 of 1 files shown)


nechry:
hello I play with the parameter whiout any timeout. This param is not a write_only. we can easily read the config value.

Fishwaldo:
Eh... I saw some discussions on this... is it valid or not @nechry

nechry:
@Fishwaldo I'm waiting Aeotec confirmation. I don't have this device to valid, but this possible.

PR #1703, opened 2019-01-28
Add new product type and id for the 2GIG CT101

mtannertdev:
New product type and id for 2GIG CT101. Confirmed full functionality/testing with actual hardware. While OpenHAB reports this device as a CT100, it does have a humidity sensor which makes it a CT101. This one of the Iris models but strangely not listed yet.


nechry:
thank @mtannertdev

PR #1708, opened 2019-02-02
Adds a new Neo Coolcam Power plug type and ID

kirichkov:
Those new power plugs are just the same as the old ones in terms of looks and packaging, but show up with a different type and ID.


nechry:
thank @kirichkov

PR #1712, opened 2019-02-09
Added Schlage BE469ZP

mnl1121:
Added Schlage Connect Smart Deadbolt model BE469ZP. Tested using OpenZwave Control Panel.


nechry:
thank @mnl1121

PR #1714, opened 2019-02-14
Add support for Heatit Z-TRM2fx

CookieMonster87:
Heatit Z-TRM2 is replaced with Z-TRM2fx. Z-TRM2 is withdrawn from marked and owners can send it back to Thermofloor to get a new Z-TRM2fx replacement.

Since both products have different config but share the same product type/id The XML files should be replaced with config for FX version as this is the only Z-TRM2 product in sale.


nechry:
thank @CookieMonster87

PR #1715, opened 2019-02-16
Updated config parameters for GE/JASCO 45609 Switch based on …

ahertz:
…https://byjasco.com/sites/default/files/product/manuals/45609%2045614%20Manual%20ENG%20070312.pdf#page=10

The current config file for the GE/Jasco 45609 switch lacks one of the possible settings for the LED (2 = always off), and has a note about mounting the switch sideways which makes no sense with this device.  This updates the config to properly match up with the device.


nechry:
thank @ahertz

PR #1717, opened 2019-02-21
update Zooz Zen27 S2 Dimmer Switch ramp rate, min/max brightness parameters

smazurov:
The [Zooz Zen27  S2 Dimmer Switch documentation](https://cdn.shopify.com/s/files/1/0218/7704/files/zooz-z-wave-plus-s2-dimmer-switch-zen27-manual.pdf?7399472298769089473) says that parameter 9, 10, and 11 are 1 byte. The XML config for this device is setting it to a 2-byte value (size=2). It's also a short instead of byte.

I also added some missing descriptions.

`make xmltest` passes with flying colors and device responds correctly once modifications are made.


nechry:
thank @smazurov

PR #1721, opened 2019-02-27
Support for the Heltun HE-ZW-THERM-FL2 Heating Thermostat

gerdesj:
make xmltest passes OK. I have tested this configuration with the OZW control panel and successfully set the clock.  All the settings are straight from the documentation:  https://drive.google.com/open?id=1AKHXZHyFCfpo_2Wd0XrisgIwilpezM7Q  There are so many settings I have a spreadsheet to generate the value items.

### DIFF
--- a/config/heltun/he-zw-therm-fl2.xml
+++ b/config/heltun/he-zw-therm-fl2.xml
@@ -1,6 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Product xmlns='http://code.google.com/p/open-zwave/'>
-  
+  <CommandClass id="67" name="COMMAND_CLASS_THERMOSTAT_SETPOINT" version="1" request_flags="4" create_vars="true" base="0">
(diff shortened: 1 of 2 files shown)


nechry:
thank @gerdesj can you please remove in Help xml element the " for all configuration parameters. the " is not require and will not useful.

gerdesj:
I think the config I have is already useful but not quite complete yet.  I will stick with it because I have one of these devices already but I have another 10 zones at home that need controlling and that is about £1500 so I need to be absolutely sure.  They are not cheap but at the moment look like the most complete option available to control electric underfloor heating (and they look gorgeous as well).  I do note that their web site mentions a lack of formal Zwave certification but they seem to have regular production and distribution.

nechry:
thank @gerdesj I found many typo and mistake, please next time use make xmltest before submit a PR.
Also the param 18 and 19 are suppose to be size 2, did you try to change values ? Your config use size of 1

PR #1727, opened 2019-03-01
Fibaro FGBS-222 Smart Implant

f-bor:
Changes for Fibaro FGBS-222 Smart Implant

Module: https://www.fibaro.com/us/products/smart-implant/
Documentation: https://manuals.fibaro.com/content/manuals/en/FGBS-222/FGBS-222-EN-T-v1.0.pdf

Everything seems working fine to me, except lifeline association. I have to enable the polling to get updated values. I don't have enough zwave knowledge to find what is wrong.

### DIFF
--- a/config/fibaro/fgbs222.xml
+++ b/config/fibaro/fgbs222.xml
@@ -10,22 +10,22 @@
   <CommandClass id="112">
     <Value type="list" genre="config" instance="1" index="20" label="Input 1 - operating mode" value="2" size="1">
       <Help>This parameter allows to choose mode of 1st input (IN1). Change it depending on connected device.</Help>
-      <Item label="Normally closed alarm input (Notification)" value="0" />
-      <Item label="Normally open alarm input (Notification)" value="1" />
-      <Item label="Monostable button (Central Scene)" value="2" />
-      <Item label="Bistable button (Central Scene)" value="3" />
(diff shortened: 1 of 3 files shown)


f-bor:
After more tests, if I include the module using secure mode, there seems to be a problem during the zwave capabilities or variables enumeration. I mean that I end up with only one input, one output, and the internal temperature. If I include the module without secure mode, the association with the controller takes a lot more time and I end up with everything detected properly (but without automatic status updates).

Fishwaldo:
Probably related to #1639 based on your comments.

f-bor:
ah it's funny : I have a ZMNHXD1 on my desk just waiting to be wired and put into service :)

Should I open an issue and provide the same logs you asked for #1639 ?

eltoro81:
Looking forward to get this device added. Thanks for your help @f-bor. I´m getting my device in a week or two and will also test. Although my Z-Wave skills are very limited...

f-bor:
hi @eltoro81,

I'm in the EU too. I won't have time to really work on this until next month.

The id 32 is for zwave COMMAND_CLASS_BASIC. You can easily find information on its purpose on internet (for example: https://www.vesternet.com/resources/technology-indepth/how-z-wave-controllers-work/).
Endpoints (id 96) are for getting information on sensors and switches (see COMMAND_CLASS_MULTI_CHANNEL).

PR #1734, opened 2019-03-08
philio relay pan05 ru

PDarkTemplar:


nechry:
the manufacturer_specific.xml already contain type and id for this device. this config will produce collision.

PDarkTemplar:
oh, copied wrong id. fixed.

nechry:
thank @PDarkTemplar

PR #1736, opened 2019-03-09
Add custom config for Leviton VRCZ4

ahertz:
The Leviton VRCZ4 uses association groups 1-4 to set devices controlled (with basic set commands) with buttons 1-4.  It doesn't make sense to auto-add the controller to group 1 for this device.  Added a custom config to allow disabling that auto association, and give a slightly more descriptive name for those groups.


nechry:
great, thank @ahertz

PR #1748, opened 2019-03-16
add support for Tronika Dimmer ZV9040 

Lotte3500:

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ManufacturerSpecificData xmlns='http://code.google.com/p/open-zwave/'>
-	<Manufacturer id="0028" name="2B Electronics">
+	<Manufacturer id="0040" name="2B Electronics">
 	</Manufacturer>
 	<Manufacturer id="0098" name="2GIG Technologies">
 		<Product type="3200" id="015e" name="CT50e Thermostat" config="2gig/ct50e.xml"/>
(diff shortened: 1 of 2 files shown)


nechry:
thank @Lotte3500

TheRealLouCipher:
Hi, I started working on adding support for the brand Sunricher(manufacturerid 0330) to OpenZWave just to find that manufacturer id 0330 is already occupied by Tronika which I believe is only a reseller of Sunricher products among others. Looking in tronika/ folder I see three devices, zsc01d.xml and zv9040.xml are Sunricher products, the last one zv9101.xml is a Everspring device(manufacturerid 0060) device(Nexa i just a branded version) so it shouldn't be there either. 

https://www.sunricher.com/z-wave-series.html

Lotte3500:
Hi, I think you are right. I only added the configuration for zv9040.xml. At that time the manufacturer id already existed, and my device showed up as a Tronika device. The device or manual does not say Sunricher, so I assumed the manufacturer was Tronika and focused on the configuration file. For me it is okay to change this to Sunricher.

zipperten:
For me the ZV-9101 also showed as tronika but the product branding tell its a Nexa, its a big swedish 433 remote control company, maby we only should use the product number ZV-×× in the config file and the certifier for the folder?

zipperten:
@TheRealLouCipher how do you mean with ZV-9101 is a everspring device, i think it is a sunricher without the power meter? https://www.sunricher.com/z-wave-series/z-wave-ac-dimmer.html

TheRealLouCipher:
Yes I might been wrong abot the Nexa/Everspring device. Seems to be
Sunricher as well. I'm gonna rearrange these devices so we get them right.
Probably on friday and then make a pull request. Will need your input on
existing devices to put them under correct manufacturerer.

Den ons 20 mars 2019 18:46zipperten <notifications@github.com> skrev:

PR #1750, opened 2019-03-18
Added useDestAsSource parameter in xml to use destination endpoint as…

jduquennoy:
This pull request adds a dest_as_source attribute to the node.
This opt-in options is meant to fix a problem with the greenwave powernode6 device, that reports its measurements with the source in the destination field (see ticket OpenZWave#1700 )

The source compiles, but I did not find an easy way to test it with my zwave setup. I did run MiniOZW, and I was able to see the zwave data frames, but nothing more.

Thanks a lot to @sarakha63 for his guidance.

### DIFF
--- a/config/device_configuration.xsd
+++ b/config/device_configuration.xsd
@@ -124,6 +124,14 @@
      </xs:restriction>
     </xs:simpleType>
    </xs:attribute>
+   <xs:attribute name='invertEndpointSource' use='optional'>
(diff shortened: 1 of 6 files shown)


Fishwaldo:
Thanks! I updated it a bit

jduquennoy:
Indeed, I can see a full part of the work was missing !
Thanks for the update and the merge 👍

PR #1752, opened 2019-03-22
Added brand Sunricher and moved miscategorized devices

TheRealLouCipher:
The devices where miscategorized as Tronika which is not a manufacturer. These devices are Sunricher. Created new folder sunricher and moved files from tronika folder to there.


nechry:
You remove 2 devices entries in Manufacturer file

TheRealLouCipher:
Sorry, my mistake. Not sure of the id's of those two devices but I think it's better to keep them in the file but under the brand Sunricher. I added them again to the manufacturer file.

nechry:
thank, 
one last thing, can you also update the distfiles.mk file, remove old entries and add the news one, this will avoid a test job failure.

https://github.com/OpenZWave/open-zwave/blob/master/distfiles.mk#L494

TheRealLouCipher:
Sure, i'll do that. I need to close this PR and do a new one right, after the changes?

nechry:
thank @TheRealLouCipher

PR #1757, opened 2019-04-05
Add Honeywell Lyric T6 Pro Z-Wave Plus Smart Thermostat TH6320ZW2003

jaminh:
Add Honeywell Lyric T6 Pro Z-Wave Plus Smart Thermostat TH6320ZW2003
https://products.z-wavealliance.org/products/2583


nechry:
thank @jaminh

PR #1760, opened 2019-04-10
Add HomeSeer Technologies HS-FS100+ Flex Sensor

nellering:


nechry:
you didn't include the config files somfy/1811265_ZRTSI.xml and homeseer/hs-fs100plus.xml

PR #1766, opened 2019-04-16
Added the Flush Thermostat OnOff 2

petrovcicklemen:
Added the configuration xml for the "Qubino Flush Thermostat OnOff 2" device and an entry into the manufacturer_specific.xml.


petrovcicklemen:
The device belong to the Qubino/Goap manufacturer.

PR #1767, opened 2019-04-17
Bugfixes Extended Status, tested on small test-network.

Fishwaldo:


petergebruers:
Right, better but not good enough, it is my first PR and my C/C++ is rusty...
Does not fix "output may be truncated copying 7 bytes from a string" on gcc 8.3
I need to improve testing before I push changes...

petergebruers:
Tested on armbian 64 bit gcc 8.3 and Apple LLVM 10.0.1

Fishwaldo:
👍 

Thanks

PR #1775, opened 2019-04-25
replace getVersionAsString() with getVersionLongAsString()

petergebruers:
This would help me Without this change, all Dev logs start with "OpenZwave Version 1.5.3770 Starting Up" (based on tag info in git) so I am not sure if I'm linking agains a build that was based on the latest commit. With this change it prints "OpenZwave Version V1.5-3770-gd11ab62b-dirty Starting Up".


Fishwaldo:
Thanks!

PR #1776, opened 2019-04-25
Implement Manager::HasExtendedTxStatus

petergebruers:
After testing and logging the controller response, to the "enable extended status" serial command, this method should return true if this status is supported.

MinOZW updated to print this status.

Tested on a UZB1 with recent firmware, not tested on a controller which lacks the capability, because I do not have one...


Fishwaldo:
Thanks

PR #1778, opened 2019-04-27
Fix GCC 8 build + return in GetXMLTagName

petergebruers:
Build on Mac with gmake -j 4 CXX=c++-8 LD=C++-8 revealed these issues.


Fishwaldo:
I removed the default case - the compiler will usually warn if a enum is not handled in a switch case - Thats a easy sanity check for us Devs that we have updated all the code dealing with enums.

petergebruers:
I see, in this specific case of translating enum (int) -> string you want the compiler to complain. I understand. I'm kind of used to "cover all cases" but I see why you don't want that here. Thanks!

Issue #1762, opened 2019-04-11
Use of non-reentrant time functions (i.e  localtime()).


nihilus:
In TimeParameters.cpp, LogImpl.cpp and  TimeStampImpl.cpp there are calls to localtime() which should be replaced by calls to localtime_r() to ensure thread safety.

Fishwaldo:
Patches Welcome!

nihilus:
`Index: openzwave-1.4.164/cpp/src/platform/unix/TimeStampImpl.cpp
===================================================================
--- openzwave-1.4.164/cpp/src/platform/unix/TimeStampImpl.cpp	(revision 23000)
+++ openzwave-1.4.164/cpp/src/platform/unix/TimeStampImpl.cpp	(working copy)
@@ -112,8 +112,9 @@
 )
 {
 	char str[100];
-	struct tm *tm;
-	tm = localtime( &m_stamp.tv_sec );
+	struct tm *tm, xtm;
+	memset(&xtm, 0, sizeof(xtm));
+	tm = localtime_r( &m_stamp.tv_sec, &xtm );
 
 	snprintf( str, sizeof(str), "%04d-%02d-%02d %02d:%02d:%02d:%03d ", 
 		  tm->tm_year + 1900, tm->tm_mon + 1, tm->tm_mday,
Index: openzwave-1.4.164/cpp/src/platform/unix/LogImpl.cpp
===================================================================
--- openzwave-1.4.164/cpp/src/platform/unix/LogImpl.cpp	(revision 23000)
+++ openzwave-1.4.164/cpp/src/platform/unix/LogImpl.cpp	(working copy)
@@ -261,8 +261,9 @@
 	// Get a timestamp
 	struct timeval tv;
 	gettimeofday(&tv, NULL);
-	struct tm *tm;
-	tm = localtime( &tv.tv_sec );
+	struct tm *tm, xtm;
+	memset(&xtm, 0, sizeof(xtm));
+	tm = localtime_r( &tv.tv_sec, &xtm);
 
 	// create a time stamp string for the log message
 	char buf[100];
Index: openzwave-1.4.164/cpp/src/command_classes/TimeParameters.cpp
===================================================================
--- openzwave-1.4.164/cpp/src/command_classes/TimeParameters.cpp	(revision 23000)
+++ openzwave-1.4.164/cpp/src/command_classes/TimeParameters.cpp	(working copy)
@@ -176,13 +176,14 @@
[...]

PR #1779, opened 2019-04-27
Replace localtime with localtime_r for thread safety

petergebruers:
#1762 
Tested on Mac llvm + gcc, not tested on Windows.


Fishwaldo:
I went for the localtime_s version instead... we try to use those versions on windows instead. (otherwise for example, windows throws a warning)

petergebruers:
OK!

PR #1780, opened 2019-04-29
Add KwikSet SmartCode 914

derekbelrose:
Added the Kwikset SmartCode 914 Z-Wave plus deadbolt

The SmartCode 914 shows as "Unknown", this should fix that problem.


nechry:
thanks @derekbelrose

PR #1781, opened 2019-04-30
Vs2019

petergebruers:
To be able to do building & debugging on windows I'd like to merge these setup files + 1 fix for "linker error unresolved external symbol "struct tm * __cdecl OpenZWave::localtime_r". This is for VS2019, the existing projects for VS2010, VS2008 are untouched. MinOZW app runs with VS2019, I did not test older versions of VS.


Fishwaldo:
As discussed, not needed.

PR #1790, opened 2019-05-13
Set thread names on NetBSD

tih:
The parameter list for pthread_setname_np() is slightly different in the NetBSD implementation.


Fishwaldo:
Thanks!

PR #1791, opened 2019-05-14
Add config for Schlage BE468ZP Connect Smart Deadbolt

alex-mohr:
Add support for the Schlage BE468ZP, which is now available on the market, and similar to the BE469ZP, but without the built-in alarm.  (It also has a white keypad instead of blue.)
https://products.z-wavealliance.org/products/3028


alex-mohr:
And fwiw, I did follow https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices and confirm that "make xmltest" validates the new file.

PR #1794, opened 2019-05-15
Fixed crash when AddCommandClass returned NULL

gizmocuz:
A crash was triggered here. Seems the only place where this was not checked


Fishwaldo:
Thanks

PR #1795, opened 2019-05-15
BeNext AlarmSound : configuration update

jano42:
The aim of this PR is to add all configuration items available for product.

I also added all metadata tags

Prior to this PR the product could not be condigured as explained in manual.


nechry:
can be nice to also name the Association group
https://www.benext.eu/static/manual/alarmsound.pdf

PR #1803, opened 2019-05-17
Fix bogus Node IDs in logging

petergebruers:
Some code used _data[3] in to log Node ID,
but not all serial data frames report back node number.
Use GetNodeNumber() instead.

Without this bugfix you'll see log lines like this:

```
Info, contrlr, Received reply to FUNC_ID_ZW_GET_VERSION:
Info, contrlr,     Static Controller library, version Z-Wave 4.61
Detail, Node045,   Expected reply was received
```

The 'Node045' part is incorrect. After this patch:
```
Info, contrlr, Received reply to FUNC_ID_ZW_GET_VERSION:
Info, contrlr,     Static Controller library, version Z-Wave 4.61
Detail, contrlr,   Expected reply was received
```

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -2153,14 +2153,17 @@ void Driver::ProcessMsg
 		}
 		case FUNC_ID_ZW_REQUEST_NODE_INFO:
 		{
+			// This code used _data[3] to log Node ID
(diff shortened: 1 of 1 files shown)


Fishwaldo:
haha! One of the most annoying things that I always forget to fix!

Finally!

PR #1806, opened 2019-05-17
Added functional header to fix compile error under VStudio

gizmocuz:


Fishwaldo:
Thanks. I was having dinner while that was compiling :)

Issue #1414, opened 2018-01-07
zwcfg xml displays incorrect max_baud_rate


ben423423n32j14e:
Example (secure included Aeotec SmartSwitch 6 Z-Wave Plus device) with UZB1 controller running on Domoticz:

<Node id="33" name="Fish Tank Pump" location="" basic="4" generic="16" specific="1" roletype="5" devicetype="1792" nodetype="0" type="Binary Power Switch" listening="true" frequentListening="false" beaming="true" routing="true" max_baud_rate="40000" version="4" secured="true" refreshonnodeinfoframe="false" query_stage="Complete">

All devices report baud rate of 40000

Either this is a bug in the reporting or the network is actually running at baud 40000.

Many people asking why devices are not displayed as Z-Wave Plus speeds on the Home Assistant forums as well.

Would be good to either get this resolved or definitively clarified from a member of the team.

jwillaz:
@Fishwaldo Would you be able to enlighten us regarding how baud rate is being set?  What device data is being read to set it? I see this in **node.cpp**:

> m_maxBaudRate = 9600;
	if( ( _data[0] & 0x38 ) == 0x10 )
	{
		m_maxBaudRate = 40000;
	}

ben423423n32j14e:
Unfortunately I don’t have the knowledge required to be able to answer your question (the above was discovered using a Domoticz installation) and the same issue appears to exist in Home Assistant based on a number of forum threads I read.

Even if the issue does not get resolved here, are you able to advise what the impact of baud being set in the zwcfg xml has on network performance?

Is the network actually running at 40000? If I change the value for the controller to 100000 does that now make it use 100kbps rate if available?

I have a large almost entirely Z-Wave + network but I'm seeing some delays and wondering if it's really performing at + speeds?

Fishwaldo:
This is information only and changing any setting in zwcfg_*.xml doesn't actually affect the network. Someone has to reverse engineer the serial api around that to figure out what the new encoding is for 100K.

ben423423n32j14e:
Would be nice if Sigma Designs stepped in to assist... Considering how almost every open source home automation project is depending on OpenZWave (for Z-Wave support), it's in their interest to provide the best customer experience. They should be providing this info to the team directly...

Fishwaldo:
[...]

PR #1808, opened 2019-05-19
Decode "speed extension" in protocol field to detect 100 kbit devices

petergebruers:
Fixes issue #1414

The most up-to-date specification of the NIF (that I could find) is in SDS13781 "Z-Wave Application Command Class Specification.pdf" chapter "3.4 Node Information Frame". However, it does not explain the 3rd byte, it still lists that byte as "Reserved". By crafting packets and decoding them with Zniffer I could determine the speed bits, which encode the "speed extension". PC Controller by Silabs also lists this property as "speed extension".

### DIFF
--- a/cpp/src/Node.cpp
+++ b/cpp/src/Node.cpp
@@ -1476,6 +1476,30 @@ void Node::UpdateProtocolInfo
 		m_maxBaudRate = 40000;
 	}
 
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Thanks!

PR #1812, opened 2019-05-21
fix spelling issues detected by lintian

alteholz:


Fishwaldo:
Thanks. 
Are you updating the Debian package? If so - there are several other changes needed...

alteholz:
Yes, already uploaded to Debian/experimental. I found two other commits to make the thing compile. Do I need others as well?

601e5fb16232a7984885e67fdddaf5b9c9dd8105
3b029a467e83bc7f0054e4dbba1e77e6eac7bc7f

Fishwaldo:
Firstly - Can you please use the versions published on the Homepage. We don't tag minor releases on GitHub. So you are always packaging up potentially older, buggy versions. We are already upto 1.6.168 and there are lots of additional bug fixes over and above those that you pointed out already. 

second - the config files in /etc/openzwave need to be writable by applications. by default they are owned by root and 644 permissions. 

I had a brief discussion with the Fedora Folks - The preferred way from them is to create a system user "zwave" and make the files ownership root:zwave with 664 permissions. I have update the local Debian files to do that. Can the official package please do the same. 

(applications will have to add the user they are running as to the dialout group already, so asking them to add to the zwave group as well shouldn't be a issue).

Finally - Not sure if you picked it up - but there are flags you can pass to the Makefile to disable using the builtin TinyXML library, and the HID controller is being depreciated and no longer compiles so the files in hidapi are no longer used/compiled. 

and last question - How can I work with you or Debian to have more frequent updates to the packages? the old 1.5 package you did caused so much confusion for users and issues as it was buggy and a based on the dev tree so not even a stable release. (you should have packaged up 1.4).

alteholz:
> Firstly - Can you please use the versions published on the Homepage. We don't tag minor releases on GitHub.

Sure, out of curiosity, what is the reason for not tagging minor releases?
 
> second - the config files in /etc/openzwave need to be writable by applications.  Can the official package please do the same.

Same answer, sure

> Finally - Not sure if you picked it up - but there are flags you can pass to the Makefile to disable using the builtin TinyXML library, and the HID controller is being depreciated and no longer compiles so the files in hidapi are no longer used/compiled.

Ok, thanks for the info.

> 
> and last question - How can I work with you or Debian to have more frequent updates to the packages?

As soon as the Debian watch file notices an update, it is on my todo-list. At the moment it looks at Github, so updates had been rather seldom.


> the old 1.5 package you did caused so much confusion for users and issues as it was buggy and a based on the dev tree so not even a stable release.

But nobody filed a bug in the Debian BTS!?

Fishwaldo:
> Sure, out of curiosity, what is the reason for not tagging minor releases?

Our CI/CD pipeline doesn't have write access to the Github repo. Releases only appear on the website after they have passed all the tests when building out OZW etc. Everything (plus more you can't see) has to be green [here](http://bamboo.my-ho.st/bamboo/browse/OZW-OZWLNX)

And the fact that there might be multiple releases "per day" when we are busy :)

> But nobody filed a bug in the Debian BTS!?

I told everyone and anyone not to use the Debian/Ubuntu/Rasbian packages not to use them :) haha
Don't worry - Fedora has the same issue :)

Thanks for your help

Issue #1667, opened 2018-12-09
Unhandled Command Class 0x2b when FGD212 sends SceneActivation after restart


stromnet:
### Issue Guidelines

While experimenting with Dev and the new CC support, I've hit some other issues.. I've got two Fibaro dimmers which on S2 sends a scene activation command.

With Dev cleanly built on FreeBSD 11.2, with an Aeon S2 stick, I just get the following when pressing S2:

```
2018-12-09 17:19:51.755 Detail, Node003,   Received: 0x01, 0x0a, 0x00, 0x04, 0x00, 0x03, 0x04, 0x2b, 0x01, 0x1a, 0xff, 0x39
2018-12-09 17:19:51.755 Detail,
2018-12-09 17:19:51.755 Warning, Node003, CommandClass COMMAND_CLASS_SCENE_ACTIVATION HandleIncommingMsg returned false
```

From what I can interpret, the 0x04 means 4 bytes long and 0x01 means `data[0] == 0x01` inside `SceneActivation::HandleMsg`.. so I cannot at all understand how HandleMsg can return false...

I've also had issues with my application (agocontrol) initially failing to detect all classes on the device, resulting in the following:
```
2018-12-09 14:23:08.683621 [0x0000000808c52d00]  [  DEBUG] OZW Node031,   Received: 0x01, 0x0a, 0x00, 0x04, 0x00, 0x1f, 0x04, 0x2b, 0x01, 0x10, 0xff, 0x2f
2018-12-09 14:23:08.683671 [0x0000000808c52d00]  [  DEBUG] OZW
2018-12-09 14:23:08.683690 [0x0000000808c52d00]  [   INFO] OZW Node031, ApplicationCommandHandler - Unhandled Command Class 0x2b
```
After a manual `RefreshNodeInfo` on the node, it managed to obtain the classes and instead now gives the HandleIncommingMsg error instead.

Both Node031 and Node003 are Fibaro FGD211.

I can provide more logs from above runs, and perhaps debugging aid, but until then I will move back to master to get a working system again.

stromnet:
Back on master:

```
2018-12-09 17:44:41.211557 [0x0000000808bf9e00]  [  DEBUG] OZW Node003,   Received: 0x01, 0x0a, 0x00, 0x04, 0x00, 0x03, 0x04, 0x2b, 0x01, 0x10, 0xff, 0x33
[...]

PR #1819, opened 2019-05-27
Fix SceneActivation

petergebruers:
Reported by @gizmocuz


petergebruers:
This is a necessary fix, but only fixes FGD212 scene activation if the node is not in the cache. It has to do with m_advertisedCommandClassMap no containing the scene CC after cacheload, leading to "Unhandled Command Class 0x2b". Reopening #1667

petergebruers:
Fix integrated in https://github.com/OpenZWave/open-zwave/commit/a3f4ccd17a3f39990cbdaf179599cc715accaab0

PR #1834, opened 2019-06-10
Fixes crash when node = NULL

gizmocuz:
Experienced a crash here. node was NULL. This is checked everywhere in this function except on this location

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -1184,7 +1184,7 @@ bool Driver::WriteMsg(string const &msg)
 
 	Log::Write(LogLevel_Detail, "");
 
-	if (m_nonceReportSent > 0)
+	if (m_nonceReportSent > 0 && node != NULL)
 	{
 		/* send a new NONCE report */
 		SendNonceKey(m_nonceReportSent, node->GenerateNonceKey());


Fishwaldo:
Can you provide a LogFile leading upto the Crash - Encrypting something without a Node means somethings is screwed up somewhere.

gizmocuz:
Sorry , i tried to replicate this, but was unable to do so.

This is what happened:

- Included (ZWave+) a door sensor
- tested the sensor (alarm class)
- Excluded the sensor

...

and it happened

Looking at this function, it seems to be possible to get a node pointer that = NULL, this was also the only place where it was not checked against NULL.

Fishwaldo:
Was there "Traffic" with the Sensor when you removed the Device? Maybe something was left over in one of the SendQueues when you removed the device.(which shouldn't happen).

gizmocuz:
Yes that could very well be, because i needed to pickup the sensor, which triggers a motion And also a burglar alarm as the unit was lift from the table. (small pin at the back checks if it is mounted). This small pin is also used to exclude, you have to flip it 3 times.
(Node was a Philio PST02-1A)

https://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/2625/Manual.pdf

petergebruers:
Patch kindly accepted, though we need to keep in mind the real cause of the problem is somewhere else.

PR #1835, opened 2019-06-10
Silence compiler warning for Visual Studio

gizmocuz:


petergebruers:
You might think we do not want to merge this, or forgot about it.

On the contrary... I was very keen to find out about this but to be honest I compile 95% of all times on mac/linux and only 5% on Windows. 

But I got thoroughly intrigued by your PR and the somewhat ominous warning "warning C4251 - ...  needs to have dll-interface to be used by clients of class ..."

I have spent several hours reading on this topic.

Took me a while to find a good description of the possible consequences of ignoring this warning, and I think this one does a nice job:

https://stackoverflow.com/questions/5661738/how-can-i-use-standard-library-stl-classes-in-my-dll-interface-or-abi

This is the MS explanation:

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251?view=vs-2019

I may grossly over-simplify, please tell me if that is the case, but here is how I see this. Due to lack of ABI, if you distribute the library as a binary (dll) and do not use the same compiler and toolset then the objects might not match...

Since you build the DLL for Domoticz, then build Domoticz based on that (supposedly using the same toolset) and distribute both at the same time, the warning can be ignored. All of them.

So why not set the compiler to ignore that warning? Is it because, in some cases, you really do want to get warned about this C4251?

Instead af selectively disabling it, you could instead of selectively disabling this per member of the class, apply that to the whole class or the whole file for that matter. After all, you have been building OZW + Domoticz for years and you have always ignored the warning, and you want to suppress more of them. 

After all, OPENZWAVE_EXPORT_WARNINGS_OFF is only a wrapper around <__pragma( warning(disable: 4251) )> and C4275

I also noticed, at least on my build system, this PR does not silence all warnings (only those about "string") - is there a reason for that?

Can you tell a bit more about this subject?

gizmocuz:
@petergebruers , I have compiled many Libraries/DLL's, and with this project, it's the first time i received it.
Yes of course we use the same compiler.

The reason not all messages have been silenced is that there are quite a few.

Maybe if a windows (VStudio) compiler is detected the error/message in the whole file/project could be silenced ?

I don't want to add a 'silence' in the Domoticz project as this is about the OZW project

petergebruers:
Thanks for the feedback. I understand. I'll have a look at it and I am inclined to silence it as you suggest, let me test that.

petergebruers:
Incorporated and superseded by PR "Fix DLL export warnings on windows" #1944 aka "OpenZWave Version 1.6-919-g51b78018"

gizmocuz:
Thanks !!!

PR #1843, opened 2019-06-17
Added MCO Home MH-P511 dimmer switch configuration files

stuartwishart:


stuartwishart:
In my original pull request I forgot to increment config/manufacturer_specific.xml Revision number.  I've corrected this but there is now a conflict that I can't resolve.

PR #1845, opened 2019-06-17
Feature/static-GetMetaDataId

ekarak:
This helper function (Node::GetMetaDataId) only maps strings to enum values, and is very useful to be able to get the metadata member from its name.
However, its a member function and therefore requires access to the Node object. It can(and should) be made static so as to be callable from wrappers (like node-openzwave-shared)


Fishwaldo:
Eh? Missing a Commit or something? only change I see is you made it public on the Node Class... 
Wouldn't you want to make it Static on the Manger Class instead?

(You shouldn't be calling anything on Node Class directly - But I can't move it into the Internal Namespace as we have a few Structures NodeData etc) that are exposed by the Manager Class... thats a refactor for another day)

ekarak:
ok, I exposed the method through Manager. Do you think it is necessary to keep the Node method private? (its not mutating anything)

Fishwaldo:
This is fixed already via the ValueID Class.

PR #1858, opened 2019-06-27
localtime_r first argument should be const

gizmocuz:
Currently it is not declared as 'const' and this is causing compilation issues for applications that also introduces their own localtime_r implementation that is using the arguments as declared in the spec


Fishwaldo:
Although the prototype was wrong, the problem is that Utils.h should have never been exposed to the application. Its not part of our API. The real Fix here would be ensuring that the Utils.h was not included in a public header file. Removing it from Value.h ensures that its not exposed and thus doesn't influence any Application Compiles.

petergebruers:
Thanks, reducing the footprint of header files seems a very good idea to me...

PR #1859, opened 2019-06-27
Added missing include for shared_ptr

gizmocuz:
Solves the following compile issue under travis:

> /home/travis/build/open-zwave-read-only/cpp/src/Node.h:566:32: error: ‘std::shared_ptr’ has not been declared

### DIFF
--- a/cpp/src/Node.h
+++ b/cpp/src/Node.h
@@ -32,6 +32,7 @@
 #include <vector>
 #include <list>
 #include <map>
+#include <memory>
 #include "Defs.h"
 #include "value_classes/ValueID.h"
 #include "value_classes/ValueList.h"


petergebruers:
I saw that, "memory" is already included in Defs.h so not sure why your build fails. I also tested the same gcc as the one you have in Domoticz ci and it passes (and ozw CI also does not complain)

But I have no issue with this PR. I am an inquisitive guy, I always like to know the real reason.

You copy the source into your tree, iirc, could there be some mixup?

Fishwaldo:
The fix here is to ensure you compiler is running in c++11 mode... old compilers don't. Added this flag to the MinOZW build.

PR #1864, opened 2019-07-01
Add Zooz ZEN25 to list of devices in manufacturer_specfic.xml

### LINKED ISSUE #1863: Zooz ZEN25 missing from `manufacturer_specific.xml`
Looks like there is a valid Zooz ZEN25 device configuration file in the Zooz directory, but it is never referenced from `manufacturer_specifc.xml`.

zlalanne:
Resolves #1863.


petergebruers:
Thank you for adding this. Still not ready to merge though, but almost there..

run "make xmltest" - it will tell you if the edits are OK.

Then push the xml file and also testconfigversions.cfg file.

An explanation of the process in more detail is on the wiki:

https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices

Fishwaldo:
Thanks!

PR #1873, opened 2019-07-10
Added Fibaro FGWD111 Walli Dimmer

TerrorSource:
Added "FGWD111 Walli Dimmer" to manufacturer_specifiec.xml
Added /config/fibaro/fgwd111.xml file with all z-wave config's from https://products.z-wavealliance.org/products/3392/configs


petergebruers:
Thanks!

PR #1879, opened 2019-07-13
adds node id to a controller state change notification 

kdschlosser:
When receiving a controller state notification the node that a command was invoked on (if any) was always set to 0. 

In pyozw (python-openzwave) because there is no way to identify a node that was involved in a command when a change in the state of the controller takes place. what was done to handle this was pyozw was designed to only allow a single command to be sent at any given time. If that command took a long time to run either the program would stall or the attempt to send another command would return False 

A thread lock was added and this lock would stop any other commands from being issued if ozw has not sent a notification that the first command has finished or it has failed. ozw was designed to queue commands being sent and pyozw is designed to not use this queue because there is no way to identify the node that a controller state update has been issued for if the same command has been issued on more then a single node there simply is no way of being able to track what node failed the command being issued on  and what node completed successfully This is the reason why pyozw was designed to only send one and wait for completion before sending another. This lock is either going to cause the program to stall or the returned value is going to be False.. The issue with stalling is obvious.. the issue with returning False is that if the lock is not in place pyozw will return the call to the method in the manager. and the returned value from the method is True/False.. False indicating that there is an issue with any parameters that have been passed. An application has no way to know if False means there is a lock. or if False means there is an issue with data that is being passed. 

This lock does not need to be there at all if there is a way to identify the node that is involved in the controller state change. The mechanics in ozw have already been created to handle this. I believe it may have been an oversight. there are 2 other locations in the driver that set the node ID to 0, and I believe these are correct in setting them to 0.


petergebruers:
Hi! That should work, I'll double check and maybe add a comment about which controller commands return a node ID.

kdschlosser:
I can do that. Give me until the end of the day.

petergebruers:
OK

kdschlosser:
I added the comments you requested.

petergebruers:
Thanks. I'll run a few tests, maybe need to clarify a few more things about the validity of the Node ID eg when removing a node.

petergebruers:
I've run the debugger on "add node" and "remove node".

I have no issue with this PR, I only want to point out this type of notification still might confuse you when you've never used them before... Please have a look at this comment and tell me if it is OK...

Might seem weird, but I've discussed it with Justin and he thinks not many software builders have had a thorough look at the controller state notifications, proven in part by the fact that node ID at the moment gets indeed forced to be 0, which it the reason for your PR after all.

I may be telling something you already know, but this part of the OZW code is new to me 🙂 

I'm setting a breakpoint before the notification gets queued...

First, I click on "Add Device" in ozwcp.

This causes log line:

2019-07-15 10:37:52.497 Detail, Queuing (Controller) Add Device

And data:

(OpenZWave::Notification) $0 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 0, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x01'
  m_command = '\x01'
  m_useralerttype = Alert_None
  m_comport = ""
}

The node ID is 0 because indeed there is no active node at this point.

uint8 GetNodeId() const
{
    return ((uint8) ((m_id & 0xff000000) >> 24));
}

No surprise.

Next stop is:

2019-07-15 10:38:36.600 Info, contrlr, ADD_NODE_STATUS_LEARN_READY

And UZB1 starts blinking, it is ready to accept a NIF

(OpenZWave::Notification) $1 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 0, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x04'
  m_command = '\x01'
  m_useralerttype = Alert_None
  m_comport = ""
}

OZW and controller then wait on inclusion event... 3 click my neo double switch:

2019-07-15 10:39:58.696 Info, ADD_NODE_STATUS_NODE_FOUND

(OpenZWave::Notification) $2 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 0, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x06'
  m_command = '\x01'
  m_useralerttype = Alert_None
  m_comport = ""
}

2019-07-15 10:40:49.912 Info, Adding node ID 6 - Non-Secure

(OpenZWave::Notification) $3 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 100663296, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x06'
  m_command = '\x01'
  m_useralerttype = Alert_None
  m_comport = ""
}

Node ID:
100663296>>24
= 6

2019-07-15 10:43:28.875 Info, Node006, Initializing Node. New Node: true (true)

(OpenZWave::Notification) $5 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 100663296, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\a'
  m_command = '\x01'
  m_useralerttype = Alert_None
  m_comport = ""
}

You can see it progresses by looking at m_event, available through uint8 GetEvent() const - and '\a' is ControllerState_Completed and the NodeID is available on some notifications.

Now let's remove this node:

Call "Remove Device":

2019-07-15 13:05:18.713 Detail, Queuing (Controller) Remove Device

(OpenZWave::Notification) $0 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 0, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x01'
  m_command = '\x04'
  m_useralerttype = Alert_None
  m_comport = ""
}

No surprise, node ID is 0 because we don't know the ID yet...

2019-07-15 13:06:13.210 Info, REMOVE_NODE_STATUS_LEARN_READY

(OpenZWave::Notification) $1 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 0, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x04'
  m_command = '\x04'
  m_useralerttype = Alert_None
  m_comport = ""
}

Now the controller is in "exclusion mode" and waits on manual action on the device to exclude it.

2019-07-15 13:07:20.214 Info, REMOVE_NODE_STATUS_NODE_FOUND

(OpenZWave::Notification) $3 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 0, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x06'
  m_command = '\x04'
  m_useralerttype = Alert_None
  m_comport = ""
}

Still ID 0.

2019-07-15 13:08:05.079 Info, Removing node ID 6

(OpenZWave::Notification) $4 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 100663296, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\x06'
  m_command = '\x04'
  m_useralerttype = Alert_None
  m_comport = ""
}

Check node stored in ValueID:

100663296>>24
= 6

Check!

2019-07-15 13:08:56.583 Info, REMOVE_NODE_STATUS_DONE

(OpenZWave::Notification) $5 = {
  m_type = Type_ControllerCommand
  m_valueId = (m_id = 100663296, m_id1 = 0, m_homeId = 3923952579)
  m_byte = '\0'
  m_event = '\a'
  m_command = '\x04'
  m_useralerttype = Alert_None
  m_comport = ""
}

After this, node is gone...

kdschlosser:
OK so for the "startup" of the inclusion and removal process the node ID that is seen in the Add and Remove notifications the ID is going to be set to 0. This seems logical because the command is not being carried out on a specific node. It is a network wide command. If a node is removed or a node is added then the notification will then attach that nodeID to the remainder of the notifications. This is how it is supposed to work as per our previous discussion when we did some investigation on this. This is how it should be done as per the ZWave specification. 

If the command gets canceled then the nodeId should be set to 0xFF if I remember correctly.

kdschlosser:
The reason why this change should be made is because of the queuing process that is done internally in ozw.

There are controller commands that carried out on specific nodes. in which the user would supply the nodeId to the manager method (I outlined which manager methods trigger which controller commands in the comments). Because these controller commands are node specific they can be "stacked". If for some reason someone wanted to iterate over all of the nodes and call Manager::RequestNodeNeighborUpdate with the current code there is no way to know what notification applies to what node. There would be no possible way to display some kind of an informative banner in a GUI. and if a node was sleeping then this would cause another set of issues. How pyozw was written to handle this wold only allow for a single command being run at a time. stalling the program until each command has reported a success or failure before allowing the user to send another command.

A real use case example would be there is a new user that is a wee bit click happy. I am sure this would happen more then we like to think it does. But a user starts clicking on things the current code would fore the program to come to a grinding halt (if using pyozw) and the application would appear to be unresponsive and have performance issues. when in fact this is by the original design. If we have the ability to remove that issue from happening I feel that it would be best for everyone. 

Justin is going to be the best person to ask on what the nodeId gets set to and during what portion of the process it gets set. I believe both you and I would be able to locate this exact information through a lot of reading of the code and quite a bit of time spent in the process. I feel that this is an important enough thing that a section in the documentation should be made outlining this process. and when various elements of the notification get changed/set.  I personally do not know how the documentation gets compiled for ozw. If someone was willing to explain the process I am willing to learn. If someone wants to spear head it, that will work as well.

@Fishwaldo you can chime in here with any of this and correct my understanding. which is often incorrect. 

This is my understanding of how the notifications should work with relation to the nodeId.
controller commands can be broken into 2 sections. those that are directed at a specific node and those that are network wide. when I state that a command is directed at a node that means that a nodeId needs to be supplied in order for the command to be carried out. RemoveFailedNode technically speaking is not directed at a node because the node is no longer available on the network. But a nodeId is required in order to carry out the command. a network wide command needs no nodeId supplied in order for the command to be run. 

Network wide commands will have a nodeId value of 0 throughout the command process. Unless the command has involved a node in some manner. The inclusion and removal of a node is an example of this. If this is the cause once that node has been involved in the command process the involved node nodeId then gets attached to the notifications.

If a controller command has been canceled then the nodeID of the notification will be set to 0xFF if the command is a "network wide" command.

petergebruers:
> Because these controller commands are node specific they can be "stacked"

I would be careful with that. The are all kind of special commands. For example, the controller command "Update Neighbors" will make the node work very hard for you and then at the end it responds by delivering data to the protocol, which ends the update technically. But the command actually only delivers data and spits out the nodes in the log file when you see this:

2019-07-15 09:56:13.222 Info, Node005, Received reply to FUNC_ID_ZW_GET_ROUTING_INFO

That result is handled in Driver::HandleGetRoutingInfoResponse(uint8* _data)

> A real use case example would be there is a new user that is a wee bit click happy.

Depends on which controller command, "Neighbor Update" could be queued but would warrant a message like "are you sure you want this because will block Z-Wave for up to X seconds per node". Afaik a "neighbor update in progress" cannot be cancelled because you cannot tell the node to stop...

And Add and Remove cannot be queued but have to be cancelled.

A command like "HasNodeFailed" retrieves data from the controller cache so it is fast and does not communicate with the node. But To get a node on that "Failed Node" list you might want to queue a "NOP" to see if it is alive, at least in some cases. Then NOP takes a long time to time out on a node with several neighbors and troutes. And there is one more detail to "Failed Node"...

See "INS13954-Instruction-Z-Wave-500-Series-Appl-Programmers-Guide-v6_81_0x v7.pdf"

4.4.18 ZW_IsFailedNode

"Used to test if a node ID is stored in the failed node ID list."

"The failed node list is limited to 5 entries.If a node is added to a full list the node with the lowest failed count will be replaced with the new node ID."

> Justin is going to be the best person to ask on what the nodeId gets set to and during what portion of the process it gets set. I believe both you and I would be able to locate this exact information through a lot of reading of the code and quite a bit of time spent in the process.

Yes, Justin may know the details of each controller command. Still I would recommend to have a look at that INS13954 because of the special nature of these commands... It is not easy.

IMHO this does not prevent us to merge the PR, does it? Maybe wait and discuss if some of the explanation here should be in the source code?

kdschlosser:
You have to remember I am looking at this from an application point of view and not the low level network communications. If the driver in ozw is busy with a command and an application requests to send another command that second command gets placed in a queue. If a command takes a really long time to process this is going to block the network. and not the application. the application can continue along it's merry way. if a command that is being run is going to take a while to process this will cause a lag. but not what appears to be a hung application This is a very simple thing to deal with because of controller status updates. an indication can be given to the user informing the user of what command is currently processing. letting the user know that the controller is busy and delays can be expected.

If there is a known list of what controller commands can take a long while to process we can also provide the application with some kind of an indication that a lengthy process has begun and an application would be able to disable interaction with the devices until that process has completed. The end result is that the application is not going to come to a standstill. 

blocking the ZWave network and blocking the application are 2 completely different things. pyozw uses a thread lock to block controller commands from being sent if the controller is busy. a thread lock in python i would assume functions the same as a thread lock in many other programming languages. It simply causes the calling thread to stop and wait until the lock has been removed. until that lock is removed an application is going to appear as being "hung" if it takes a long time before the lock gets removed.

I can code around this whole process by creating a thread worker to handle this whole process. and it would not allow an application direct access to the lock. and it would always release the application. My issue with this is that ozw has this handled internally already so why create 2 threads that do almost exactly the same thing???... 


This is the biggest question I have.. and depending on how this gets answered is going to determine what changes I can make on the python side of things.

I am going to really oversimplify this process. This is an example and is not factual. please do not compare this to the specification as I already know it is not going to be the same it gives a rough outline of what takes place and is not exact. 

so lets say a command is a series of communication between the driver and the network, node, controller, it doesn't matter who the destination is. but say the first step is a packet sent out that is a request to run a command. and a packet gets sent back stating the command is running. after however long it takes the command to run another packet gets sent back stating completion of the command (success or failure does not matter)

so if a request to run a command is sent does it automatically move onto the next item in the queue and send it?... or does it will wait for the 1st response that indicates the command was received and is running. then it moves onto the next commend. or maybe it will sit there and wait for the completed reply before moving onto the next. This is what I do not know. does the network allow for parallel tasks to be performed? or is it dependent on the command that is issued? Now i do realize that the network is not dual band so only one packet can be flying at any given point. so this negates true parallel processing. one could say the same thing about threading as well. a processor on a PC is only able to do a single thing at a time. today's processors that have multi core can be considered as multiple processors in the same container. just like a multiple CPU machine would have more then a single CPU in the same chassis. Its the appearance of parallel i am speaking of, not the actual task that is being performed. 

I do not know if the order in which commands are issued will finish up in the same order they are sent in. 

I send commands to be performed on nodes 10, 11, 12 and 13. will those commands always finished up in the same order 10, 11, 12 and 13? or can they finish up as 12, 10, 13, and 11 ???

kdschlosser:
I am trying to gain a better understanding of the process in order to provide the best solution thus giving the user the best overall experience. I do not believe thread locking of any kind should be used if there is a queue in place. The purpose to the queue is to allow a program to drop off a task that is to run and then to go about it's business. stalling a program to wait for a task to be run that has been placed into a queue I do not believe is the best way to go about it. 

In threadworkers that I have written in the past i have written 2 entry points. 
the first will drop off a task and continue about it's way.. These tasks can also have a completion callback associated with them. there is a flag that can get set to for the execution of the callback in a new thread new thread and not from the thread worker thread. This ensures that code that is executed in the callback will not delay the threadworker. if the flag is not set this would be for synchronization between tasks being processed. keeps then in order. 
 
the second is going to drop off a task and wait for it's completion. The second type are usually for high priority tasks where there may be data that has to be returned. these tasks will skip the queue and will get run as soon as the task that is currently running gets completed. once the high priority task has completed then the queue resumes.

This is how I design my queuing systems. You mentioned Add and Remove and they are not queued. so what happens when a user wants to issue one of these command and there is already a command currently running? and what happens if the UpdateNeighbors is running and cannot be canceled?

petergebruers:
Sorry for the delay... I understand your issues with this from an end-user and gateway builder perspective. I'll try to answer in more detail next week, for now assume the controller will bark at you when you try something when it already doing something else.

I'll merge this because I think this change might help you although it does not fix all "queue issues" as you call them.

kdschlosser:
There is simply a lot of ????'s I have on how OZW handles controller commands and how the queue process works. I can go digging through the code to get the answers. I was hoping that if someone knew the answers it would be easier to get them by asking.

Now that this has been merged I am going to create a new branch on my fork of pyozw. If there is anyone interested in doing some testing they are more then welcome to give it a go. I do know it compiles properly on windows. I have not yet had the chance to do any testing for the various linux flavors and on mac .

I am going to finish up some documentation first. This way someone that is testing will be able to build it.

kdschlosser:
Here is the link to the branch I mentioned above. for those that want to give it a test run.

https://github.com/kdschlosser/python-openzwave/tree/test

report any issues using the above link.

PR #1891, opened 2019-07-21
Correct nept user code report code index BeNext Tag Reader

gizmocuz:
_data[8] contains the index of the code/tag that was used to control home/arm.
_data[11] is always zero


petergebruers:
You have told me you've tested the change and besides that the index used is in an "if" statement saying "else if (EventParamLength == 1)" which would mean only _data[8] would be valid anyway.

AFAIK does not (yet) use NEPT_UserCodeReport, so this won't affect users right now, correct?

Although I cannot test it I shall squash and merge...

gizmocuz:
@petergebruers , yes this has been tested. _data[8] contains the index of the code used

petergebruers:
Thanks!

PR #1898, opened 2019-07-28
Fix Philio door sensors

gizmocuz:


gizmocuz:
@nechry , thank you ! Always forget this...

PR #1901, opened 2019-08-08
Add HANK motion sensor type 300

GeertvanHorrik:
A config file was already available, but was not added to manufacturer_specific.xml


petergebruers:
Thank you for your contribution. Please, at the start of the config,  set Revision="2". Under `<ChangeLog>` then add a line with contact info and some description of the change pointing to that revision. Set <ManufacturerSpecificData Revision="46" then run "make xmltest" to see if everything is OK.

GeertvanHorrik:
Thanks for the review, should be fixed.

petergebruers:
Thanks!

PR #1902, opened 2019-08-09
Add support for FIBARO Walli Outlet

sjorge:
This config adds support for the [FIBARO Walli Outlet](https://manuals.fibaro.com/content/manuals/en/FGWOEF-011/FGWOEF-011-T-EN-v1.0.pdf), I tested this against a model E.

When I ran `make dist-update` it also picked up the config and image for zw164, which is already commit to the repo, I guess this step was forgotten when this one was added.


petergebruers:
Thank you for contributing!

A few small issues, we're nearly there.

> When I ran `make dist-update` it also picked up the config and image for zw164, which is already commit to the repo, I guess this step was forgotten when this one was added.

You are correct! What you did is OK.

I've noticed the "alarm" parameters have max="4" imho that should be max="4294967295"

You'll see the build has failed, please run "make xmltest" and it will tell you what to do, and more info on the config format is here:

https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices

If everything passes I'll squash merge.

petergebruers:
Oh and please remove auto="true" on lifeline, OZW does that automatically for all Z-Wave Plus devices.

sjorge:
Ah interesting, locally make xmltest complains about some utf8 encoding issues but says everything is fine, I winder if my perl is too old. But it easy to figure out thanks to the PR checks.

petergebruers:
Thanks for the update. I had quite some fun getting perl to work with the test script op my mac. I've noticed a small issue, the script does not seem to have written the correct version of the manufacturer_specific.xml file into testconfigversions.cfg... I've seen it happen as well. I'll push an update and merge.

petergebruers:
(permission denied)

-> either give me rights or remove "config/manufacturer_specific.xml" entry from the testconfigversions.cfg file and rerun make xmltest, it should have different version and hash...

sjorge:
I've removed config/manufacturer_specific.xml and reran make xmltest.

PR #1909, opened 2019-08-19
Added GoControl LB30Z-1 BR30 bulb

mtannertdev:
The GoControl LB30Z is functionally the same as the LB60Z, just in a different form factor. Adding the specific device (which I own and validated) for support.


nechry:
thanks @mtannertdev

PR #1910, opened 2019-08-19
Linear WT001 and WS15Z group 1 associations

ratatine:
These devices support group 1 associations to update status of associated devices. (On off and dim in the case of the WD500Z.


petergebruers:
Thanks!

PR #1913, opened 2019-08-22

DeviantEng:
Added new Alfred DB2 Deadbolt.  No configuration items that I'm aware of, but I've reached out to the manufacturer to confirm.  Wanted to at least add a clean name.


nechry:
thanks @DeviantEng

PR #1915, opened 2019-08-23
Fix sigsegv after refresh node. Fix crash when ozwcache holds data to deleted nodes

petergebruers:
Second crash observed when Log has "Dropping Notification as ValueID does not exist"


petergebruers:
@Fishwaldo - forgot to mention I requested a review because you have been working on this and I am not sure if I am fixing symptoms or root cause...

PR #1928, opened 2019-09-03
Popp Window/Door Sensor with Tilt Config

Tristan79:
Note: Parameter: 10 & 11: Are also the same in the manual.
Parameter: 3 Bit 0 naming is strange. But yeah, it is from the official
manual.
I do not think it set ‘No Indications’ cuz if the default value has
this bit enabled, next to bit 1 and 2. What is the point? But I leave
it as is, like in the manual.


petergebruers:
Thanks! Fixed a small error (param size). Thanks also for commenting on the "weird things", indeed that is what the manual says...

Issue #1899, opened 2019-07-28
Labels of PST02 in Secure Mode are wrong in ozwcp


petergebruers:
I do have a strange issue with the PST02 !

When i included the node 'secure', and operating the device magnet to simulate a door open/close,
i am receiving strange index values with strange labels, or no labels at all:

![image](https://user-images.githubusercontent.com/2265299/62005486-41b04e80-b134-11e9-9757-254199804d6f.png)

I excluded it, included it in the normal way, and it seems to work much better !!

Excluded it... included it secure again....

Again strange values are received:

![image](https://user-images.githubusercontent.com/2265299/62005550-2eea4980-b135-11e9-969f-34095f2e2a1c.png)

Last test, exclude, include normal again, working without any issue !

I think this is unrelated to this issue, but there is something not correct here...

_Originally posted by @gizmocuz in https://github.com/OpenZWave/open-zwave/issues/1896#issuecomment-515752364_

petergebruers:
You don't happen to have an OZW_Log.txt from the start of inclusion (secure mode) up to the point where you call a refresh on this device in ozwcp?

gizmocuz:
@petergebruers , hereby a log from the include moment, till some strange index are received:

[OZW_PST02_Secure_Log.zip](https://github.com/OpenZWave/open-zwave/files/3439434/OZW_PST02_Secure_Log.zip)

And here a log after a restart (node was already included secure)... not doing anything (sensor is under my hands while typing)...

[OZW_PST02_After_Restart_Log.zip](https://github.com/OpenZWave/open-zwave/files/3439437/OZW_PST02_After_Restart_Log.zip)

petergebruers:
Thanks for the log, that really helped.

I can see what causes this, but I still have to work out the details in the source code, and a fix... But explaining it is step 1.

The PST02 sends "encapsulated multi-command" meaning it can pack multiple CCs into one or more transmissions, this makes the data transfer more efficient. 

If the number of commmands is low, it fits in to one encrypted packet and OZW decodes this as expected. For example:

```
2019-07-28 20:08:45.489 Info, Node010, Received encapsulated multi-command from node 10
2019-07-28 20:08:45.489 Info, Node010, Received SensorMultiLevel report from node 10, instance 1, Air Temperature: value=84F
2019-07-28 20:08:45.490 Detail, Node010, Refreshed Value: old value=-40.0, new value=84, type=decimal
[...]

PR #1929, opened 2019-09-04
MultiCmd: implement sanity checks and improve logging

petergebruers:
This stops creating bogus data, as reported here:
"Labels of PST02 in Secure Mode are wrong in ozwcp"
https://github.com/OpenZWave/open-zwave/issues/1899

This is the first step in fixing that issue, the next step is supporting multipart secure messages

This PR was tested by using both a real PST02 and data simulated by injecting valid and invalid data.

The PST02 sends a multipart secure message, the first part now logs:

```
2019-09-04 09:51:13.429 Info, Node013, Multi-command frame received, encapsulates 4 command(s)
2019-09-04 09:51:13.431 Info, Node013, Received Battery report from node 13: level=100
2019-09-04 09:51:13.434 Detail, Node013, Refreshed Value: old value=100, new value=100, type=byte
2019-09-04 09:51:13.435 Detail, Node013, Changes to this value are not verified
2019-09-04 09:51:13.437 Info, Node013, Received Notification report (>v1): Type: Home Security (7) Event: Motion Detected at Unknown Location (8) Status: true, Param Length: 0
2019-09-04 09:51:13.440 Detail, Node013, Refreshed Value: old value=2, new value=2, type=list
2019-09-04 09:51:13.445 Detail, Node013, Changes to this value are not verified
2019-09-04 09:51:13.447 Info, Node013, Received SensorMultiLevel report from node 13, instance 1, Illuminance: value=99%
2019-09-04 09:51:13.449 Detail, Node013, Refreshed Value: old value=98, new value=99, type=decimal
2019-09-04 09:51:13.451 Detail, Node013, Changes to this value are not verified
2019-09-04 09:51:13.454 Error, Node013, Multi-command command part 4 with base 22 is invalid, end > highest_index (28 > 24)
2019-09-04 09:51:13.456 Warning, Node013, CommandClass COMMAND_CLASS_MULTI_CMD HandlerMsg Returned False
```

This stops the creation of a 4th value, based on whatever data is after the truncated frame...

Also tested if it still works on valid data:

```
2019-09-04 09:51:04.057 Info, Node013, Multi-command frame received, encapsulates 1 command(s)
2019-09-04 09:51:04.059 Info, Node013, Received SensorMultiLevel report from node 13, instance 1, Air Temperature: value=71F
2019-09-04 09:51:04.062 Detail, Node013, Refreshed Value: old value=-40.0, new value=71, type=decimal
2019-09-04 09:51:04.064 Detail, Node013, Changes to this value are not verified
2019-09-04 09:51:04.066 Info, Node013, Multi-command, all 1 command(s) processed
```

### DIFF
--- a/cpp/src/command_classes/MultiCmd.cpp
+++ b/cpp/src/command_classes/MultiCmd.cpp
@@ -49,18 +49,49 @@ namespace OpenZWave
 			{
 				if (MultiCmdCmd_Encap == (MultiCmdCmd) _data[0])
 				{
-					Log::Write(LogLevel_Info, GetNodeId(), "Received encapsulated multi-command from node %d", GetNodeId());
+					int commands;
 
-					if (Node const* node = GetNodeUnsafe())
+					if (_length < 3)
+					{
+						Log::Write(LogLevel_Error, GetNodeId(), "Multi-command frame received is invalid, _length is < 3");
+						return false;
+					}
+					else
+					{
+						commands = _data[1];
+						Log::Write(LogLevel_Info, GetNodeId(), "Multi-command frame received, encapsulates %d command(s)", commands);
+					}
+
+					if (Node const *node = GetNodeUnsafe())
 					{
 						// Iterate over commands
-						uint8 base = 2;
-						for (uint8 i = 0; i < _data[1]; ++i)
+						int base = 2;
+						// Highest possible index is _length minus two because
+						// array is zero-based and _data starts at second byte of command frame
+						int highest_index = _length - 2;
+						for (uint8 i = 1; i <= commands; ++i)
 						{
+							if (base > highest_index)
+							{
+								Log::Write(LogLevel_Error, GetNodeId(),
+										   "Multi-command command part %d is invalid, frame is too short: base > highest_index (%d > %d)",
+										   i, base, highest_index);
+								return false;
+							}
+
 							uint8 length = _data[base];
+							int end = base + length;
+							if (end > highest_index)
+							{
+								Log::Write(LogLevel_Error, GetNodeId(),
+										   "Multi-command command part %d with base %d is invalid, end > highest_index (%d > %d)",
+										   i, base, end, highest_index);
+								return false;
+							}
+
 							uint8 commandClassId = _data[base + 1];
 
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Looks good to me. I don't think I've actually ever seen a device that does the MultiCommand :)

Out of interest - What is the Controllers NIF? (if you can capture it or show me what we send in FUNC_ID_SERIAL_API_APPL_NODE_INFORMATION). I would have expected us to advertise MultiCommand in the NIF for a device to that encapsulated message to us.

petergebruers:
Inclusion NIF:

Received: 0x01, 0x1c, 0x00, 0x4a, 0x0d, 0x03, 0x0a, 0x15, 0x04, 0x07, 0x01, 0x5e, 0x80, 0x71, 0x85, 0x70, 0x72, 0x86, 0x30, 0x31, 0x84, 0x59, 0x73, 0x5a, 0x8f, 0x98, 0x7a, 0xef, 0x20, 0x43

That's ZW->HOST: REQ | 0x4A | funcID | bStatus | bSource | bLen | basic | generic | specific | cmdclasses[ ]

And indeed in the log, decodes as

Node010,   Optional CommandClasses for node 10:
(...)
COMMAND_CLASS_MULTI_CMD

I've only checked this device in secure mode, because then it produces one multicmd frame that is 4 bytes "too long" to fit in a single secure encap because of the 20 byte secure header... Then OZW does not handle the joining of the split packet. I expect the problem to be non-existent in non-secure mode but to be honest I haven't checked it. You want me to capture some non-secure data?

petergebruers:
BTW here's something the PST02 sends (I think it only does that once, at startup):

```
Decrypted Packet: 0x00, 0x8f, 0x01, 0x00
Multi-command frame received, encapsulates 0 command(s)
```

That is valid... But I do not get the point of sending such a frame...

Judging by the weird things we've discussed on this issue tracker so far, I guess developing Z-Wave devices probably isn't very easy. I have modified one of the SDKs examples but haven't started from scratch which is totally different I guess -  ;)

Fishwaldo:
>Inclusion NIF: 
ahhh sorry, I meant the Controller NIF that we set at startup. It should only contain the ZW+ CC.

PR #1932, opened 2019-09-04
Added device 12725 (GE Dimmer)

sfeakes:


sfeakes:
Updated the revision#

PR #1944, opened 2019-09-19
Fix DLL export warnings on windows

petergebruers:
Discussion https://github.com/OpenZWave/open-zwave/pull/1835
Tested with Domoticz.
Does not affect other platforms, but cleans up clutter in
header files.

### DIFF
--- a/cpp/src/Bitfield.h
+++ b/cpp/src/Bitfield.h
@@ -75,8 +75,7 @@ namespace OpenZWave
 				Iterator End() const;
 
 			private:
-				OPENZWAVE_EXPORT_WARNINGS_OFF
-				vector<uint32> m_bits;OPENZWAVE_EXPORT_WARNINGS_ON
+				vector<uint32> m_bits;
 				uint32 m_numSetBits;
 				uint32 m_value;
 		};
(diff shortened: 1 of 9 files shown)


petergebruers:
@Fishwaldo just a quick check before merge: I also checked in "dist/openzwave.spec" - is that OK? This update to headers does not mess with any work you are doing on Dev?

gizmocuz:
Thanks for the PR, maybe also add 'std::' before all std types and remove the namespace.
In the code i see a lot of std::vector, std::string, also string without std::

https://akrzemi1.wordpress.com/2019/03/14/not-using-namespace-std/#targetText=using%20namespace%20std%3B,easily%2C%20without%20bothering%20with%20namespaces.

petergebruers:
> maybe also add 'std::' before all std types and remove the namespace.

Sure, it is on my todo list. @Fishwaldo has already done a lot of work to add proper namespace support. Code style probably changed over time, or maybe it depends on the contributor. Similarly, I guess nowadays we would not define int8 in Defs.h but use int8_t from <cstdint>

petergebruers:
@Fishwaldo OK to merge?

Issue #465, opened 2015-03-14
Log File Rotation. 


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Run a program that uses OpenZwave with logging turned on
2. View log file
3. Logrotate the log file with copy truncate
4. Watch log file return to original size
5. View log file and notice null values at the head of the file
6. Log file grows endlessly until it fills up the partition

This issue is explained here:
http://serverfault.com/questions/221337/logrotate-successful-original-file-goes-
back-to-original-size

What is the expected output? What do you see instead?
The log file should start in size at zero and grow from there. Also, there 
should not be null values at the head of the file.

What version of the product are you using? On what operating system?
Domoticz v2.2248 using OpenZwave Version 1.3.1008 on Linux raspberrypi 3.12.36+ 

The fix is to not keep the offset when writing to the log file, but to append 
to the file when writing.

```

Original issue reported on code.google.com by `martywes...@gmail.com` on 24 Feb 2015 at 6:28

michapr:
I think a daily log rotation by default would help everyone.
Restarting OZW is really not a solution, for larger networks it maybe a problem, for battery devices the last status will be lost etc.
I like to wotk with the OZW log for debugging issues, but do not want to restart every week my service...

Hope it will be implemented very soon ;)

apexo:
The usual way to do this is to let logrotate *rename* the log file (no copying, no truncation) and then trigger the daemon (e.g. via a SIGHUP) to re-open its log files.

michapr:
> and then trigger the daemon (e.g. via a SIGHUP) to re-open its log files.

Yes,... but how to trigger without restarting all the OpenZwave process?
There is missing any "graceful"  (or like this)

bibi21000:
> I think a daily log rotation by default would help everyone.
No, I don't log in debug mode, so a weekly rotation is nice for me :)
why not using the copytruncate option of logrotate ?
http://linux.die.net/man/8/logrotate

2016-04-18 17:33 GMT+02:00 Michael P. <notifications@github.com>:

Fishwaldo:
Signal Handlers in a Library are a bad idea. 

What if the Application wants to do something different with SIGHUP?

When I do this, it will be a function call. Upto the application to decide when to call the function. 

[...]

PR #1946, opened 2019-09-19
Implemented simple ReopenLogFile

electrofloat:
This is a base for https://github.com/OpenZWave/open-zwave/issues/465


Fishwaldo:
This doesn't really do anything (especially if AppendLog is true) as it just closes and opens the log file. For rotation, we need to rename the old log file and start a new one (regardless of AppendLog) so yes - a base, but still needs some work.

electrofloat:
> This doesn't really do anything (especially if AppendLog is true) as it just closes and opens the log file.
Exactly.

> For rotation, we need to rename the old log file and start a new one (regardless of AppendLog) so yes - a base, but still needs some work.

The rotation is the not job of the lib itself. Logrotate does the rotation (on linux), sends a SIGHUP to the app, the app calls ReopenLogfile in the lib, which reopens the same logfile and everything works as expected.

Fishwaldo:
Your assuming Linux and Systems with logrotate. Windows/Mac/Solaris etc etc etc may not/do not have such a thing. Hence the actual rename of the log file should be done in the library.

I appriciate the PR - I'll update it eventually. :)

electrofloat:
That is true. Since the issue was mainly talking about logrotate (https://github.com/OpenZWave/open-zwave/issues/465#issuecomment-211425247) and not handling SIGHUP in the lib (https://github.com/OpenZWave/open-zwave/issues/465#issuecomment-211913649)

That is what I implemented. I'd really appreciate some kind of logrotation for the log it creates either by logrotate or by any other means.

Fishwaldo:
It just needs a rename between the close/open calls... (but needs to implemented in the platform abstraction classes so its portable). 

Question begs - What do we rename to, and how do we handle duplicates etc. 

I'll get to it one day :)

Fishwaldo:
Related to #2146

Fishwaldo:
Merged to Dev Branch. Thanks!

PR #1952, opened 2019-09-24
Update manufacturer_specific.xml for FGFS101 Zwave+ Flood Sensor

nemccarthy:
Add Fibaro System device FGFS101 Zwave+ Flood Sensor - added entry for new device ID for the AU region


petergebruers:
Thanks!

PR #1954, opened 2019-09-24
Update ZMNHDDx.xml

chrisns:
make the label for `Input 1 switch type` match the `Input 2 contact type`
annoying whitespace difference


nechry:
thank's @chrisns

PR #1968, opened 2019-10-07
Update 26933-motion-dimmer.xml

tech165:
Option 8,10,12 and 15 should be size 2 or short as per data sheet. Could not config dimmer with int or byte.

Thanks


petergebruers:
Thanks!

Issue #1981, opened 2019-10-18
(Qubino ZMNHYD1 .. and also others) crash on first time SetValue(bool) after include


gizmocuz:
I have received some reports about OZW crashing the first time when a (boolean) value is set after a node got included.
This time i can replicate this with a Qubino ZMNHYD1)

It does not matter if it is included secure or not.

I also found where it goes wrong:

`bool SwitchBinary::SetState(uint8 const _instance, bool const _state)`

And it goes wrong here:

```
				if (GetVersion() >= 2)
				{
					Internal::VC::ValueByte* durationValue = static_cast<Internal::VC::ValueByte*>(GetValue(_instance, ValueID_Index_SwitchBinary::Duration));

```

At this point durationValue = NULL, and is used right below:

`uint8 duration = durationValue->GetValue();`

I have no idea why this happens, but I would propose a fix like this:

```
					uint8 duration = 0xff;

					Internal::VC::ValueByte* durationValue = static_cast<Internal::VC::ValueByte*>(GetValue(_instance, ValueID_Index_SwitchBinary::Duration));
					if (durationValue != nullptr)
					{
						uint8 duration = durationValue->GetValue();
						durationValue->Release();
						if (duration == 0xff)
						{
							Log::Write(LogLevel_Info, GetNodeId(), "  Duration: Default");
						}
						else if (duration >= 0x80)
						{
							Log::Write(LogLevel_Info, GetNodeId(), "  Duration: %d minutes", duration - 0x7f);
						}
						else
						{
							Log::Write(LogLevel_Info, GetNodeId(), "  Duration: %d seconds", duration);
						}
					}

```

This way we always use the default duration (0xff) when we can not retrieve the (durationValue)

Hope that's OK, i will create a PR for this. (It is also good to check against NULL pointer results)

Fishwaldo:
Known issue. This is caused by the interview order that I’ve started working on in Dev branch. No need for a PR.

gizmocuz:
@Fishwaldo , sorry, I just noticed your response.
Please feel free to delete/close/merge, but checking against a nullptr is always good right ?

petergebruers:
@gizmocuz it affects more devices and CCs and the problem is more fundamental, that is why your patch is *good* but I am not inclined to merge it because it works around the problem, it does not fundamentally fix the cause.

[...]

PR #1982, opened 2019-10-18
Check durationvalue against nullptr, using 0xff as default

gizmocuz:


petergebruers:
"error: unused variable ?duration? [-Werror=unused-variable]"
You've declared (a new copy of) duration in the "if" statement.
But let's discuss this in #1981 ...

gizmocuz:
@petergebruers , corrected, thanks for reporting!

Fishwaldo:
As mentioned, this was a side effect of a different bug that we should have corrected with the Interview Reordering that has been in master for a while now.

PR #1984, opened 2019-10-18
ClimateControlSchedule, Releasing values

gizmocuz:
Making sure values are released, and checking for nullptr


gizmocuz:
I noticed that 95% of the code gets/using values with a 'if' case like:

```
					if (Internal::VC::ValueList* value = static_cast<Internal::VC::ValueList*>(GetValue(_instance, sceneID)))
					{
						/* plus one, as we have our own "inactive" entry at index 0 */
						value->OnValueRefreshed(keyAttribute + 1);

```

But not on all places, and patches like this will make it persistent..
 
This patch also calls **value->Release()** that was not called here

PR #1985, opened 2019-10-18
Making GetValue code more consistent, Releasing values

gizmocuz:
I noticed throughout application that 95% of the time GetValue is retreived/used via a 'if statement like:

```
			if (Internal::VC::ValueSchedule* value = static_cast<Internal::VC::ValueSchedule*>(driver->GetValue(_id)))
			{
				res = value->.....

```

This is not done on some locations, and could also aid checking against nullptr's
This patch also 'releases' the values

### DIFF
--- a/cpp/src/command_classes/ClimateControlSchedule.cpp
+++ b/cpp/src/command_classes/ClimateControlSchedule.cpp
@@ -317,6 +317,8 @@ namespace OpenZWave
 							if (Internal::VC::ValueByte* setback = static_cast<Internal::VC::ValueByte*>(GetValue(instance, ValueID_Index_ClimateControlSchedule::OverrideSetback)))
 							{
 								Msg* msg = new Msg("ClimateControlScheduleCmd_OverrideSet", GetNodeId(), REQUEST, FUNC_ID_ZW_SEND_DATA, true, true, FUNC_ID_APPLICATION_COMMAND_HANDLER, GetCommandClassId());
+								// nullptr check after "new" is kind of placebo, but it makes for consistent code
+								// discussion: https://github.com/OpenZWave/open-zwave/pull/1985
 								if (msg != nullptr)
 								{
 									msg->SetInstance(this, instance);
@@ -331,8 +333,20 @@ namespace OpenZWave
 								}
 								setback->Release();
 							}
+							else
+							{
+								Log::Write(LogLevel_Warning, GetNodeId(), "ClimateControlSchedule::SetValue couldn't Find ValueID_Index_ClimateControlSchedule::OverrideSetback");
+							}
+						}
+						else
+						{
+							Log::Write(LogLevel_Warning, GetNodeId(), "ClimateControlSchedule::SetValue state->GetItem() returned nullptr");
 						}
 						state->Release();
+					}
+					else
+					{
+						Log::Write(LogLevel_Warning, GetNodeId(), "ClimateControlSchedule::SetValue couldn't Find ValueID_Index_ClimateControlSchedule::OverrideState");
 					}
 				}
 
--- a/cpp/src/command_classes/ClimateControlSchedule.cpp
+++ b/cpp/src/command_classes/ClimateControlSchedule.cpp
@@ -316,7 +316,6 @@ namespace OpenZWave
 						{
 							if (Internal::VC::ValueByte* setback = static_cast<Internal::VC::ValueByte*>(GetValue(instance, ValueID_Index_ClimateControlSchedule::OverrideSetback)))
 							{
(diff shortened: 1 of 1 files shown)


petergebruers:
Hi! I like the way you've rewritten that as nested ifs... But I am not convinced about the nullptr check. Please be patient with me, maybe I misunderstand the point.

I have to say I am not a noob, that would be false modesty, but I am also certainly not a seasoned C++ coder like you @gizmocuz  and @Fishwaldo - I am still learning.

But something told me, this check has a high level of "placebo effect"

```cpp
Msg* msg = new Msg("ClimateControlSche...
if (msg != nullptr)
{
    ...
}
```

The spec says "new" can either throw an exception or return a null pointer, and in both cases the only reason is "memory issues".

So that would be rare, we've got plenty of memory, but this leaves us with two questions...

- nullptr or exception?
- if "new" fails, what options do you have? How to handle?

In your code, the second question seems to be answered like "we don't have a msg object so don't send anything because we cannot do anything and pretend it did not happen.". I am not sure what you can do after "new" fails, maybe log a few lines and abort? Maybe pretending it did not happen is an option but I got a feeling if nobody cleans up memory, this error won't go away.

I could not resist, and wrote a small test program:

```cpp
#include <iostream>

struct blah
{
    int blah[1000000000];
};

int main(int /*unused*/, char ** /*unused*/)
{
    float how_many = 0;
    try
    {
        while (auto ptr = new blah != nullptr)
        {
            how_many++;
        }
    }
    catch (const std::exception &e)
    {
        std::cerr << "Exception caught: " << e.what() << '\n';
    }

    std::cout << "how_many = " << how_many << '\n';
}
```

Result on my Mac:

```
new(88512,0x116d25d40) malloc: can't allocate region
*** mach_vm_map(size=4000002048) failed (error code=3)
new(88512,0x116d25d40) malloc: *** set a breakpoint in malloc_error_break to debug
Exception caught: std::bad_alloc
how_many = 35181
```

This might depend on the compiler and the operating system, maybe you get a nullptr...

I'd say the nullptr check isn't working on my system...

If there is no memory, we get std::bad_alloc and everything goes down, or we get a nullptr and at the next dereference, a "segmentation fault". In either case we were out of memory and the whole process goes down anyway.

So what is the advantage of doing that check? Does it have security implications?

gizmocuz:
Who knows what the feature bring, checking against nullptr seems always good and can not harm
As you know there should be a lot more pointer checks, and yes sometimes you don't have to, but as pointed out sometimes your do.
When you make this a default habit, everything is checked

But this is also about releasing the values which is more interesting

petergebruers:
Thank you for reflecting on this. I sure don't mind about the general rule and the habit. My comments apply to this very specific case only, doing the nullptr check after "new". Talking about general rules I would say nullptr check is good but only when every "if" part has an "else" part and the question here remains: what to put in the "else" part. What if you did not get a "new msg object"? Bail out? Throw a custom exception? Pretend it did not happen?

Yeah, sure, thanks for fixing the release counter. @Fishwaldo mentioned, since we're on C++11, OZW could move to shared_ptr (or unique pointers in some cases) instead of doing its own reference counting, but that is of course easier said than done.

I'll let @Fishwaldo have the final word on this, I am inclined to add a comment to that check referencing this PR and merge.

petergebruers:
Had a quick chat with @Fishwaldo and we'd say nullptr check is placebo but that is surely not the essence of this PR. The "nested ifs" are much clearer than the original code... Before we merge, we'd like to reflect a bit more on the "what if we do not get a ..." part (eg value list), I mean, what should be in the "else". Some logging?

gizmocuz:
(Almost!) Nowhere in the code this is checked/logged, if you want to do so, this should be done on all places.
I would prefer a debug/warning line in the log, and a notification (debug/warning) that applications can display in their log to indicate a possible issue in the underlying library/library call/result

I would suggest a general look/change in the complete code to make changes like this, and check against nullptr's ... fair enough, not for 'new' pointers

petergebruers:
Sure, ... we have more work to do...

At the moment, I am trying to implement "test cases" based on the gtest framework that's already there. See for example cpp/test/ValueID_test.cpp. Please be patient, it is not as easy as that might sound.

I'd like to keep this PR open and push some commits when I have an idea.

petergebruers:
I am still fighting a bit with the "test" system in OpenZWave but I am getting there. I am able to test the Value reference count is wrong before this PR and correct after this PR. I am able to do so but I am also working on a proposal for a new function to get "End Point" from a ValueID and also a proposal to improve the message of the exception thrown when you ask OZW to handle an invalid ValueID.

petergebruers:
Thanks @gizmocuz. Tests will be added as separate PR.

PR #2008, opened 2019-11-24
Make set after get optional

### LINKED ISSUE #1980: Ability to disable confirmation get for devices with reliable automatic state reporting
Check out the following log. It contains simple action of me toggling (dimming) an Aeotec LED Bulb 6 Multi-Color (E27).

[OZW_Log.txt](https://github.com/OpenZWave/open-zwave/files/3737382/OZW_Log.txt)

From what I understand, this happened:

* Single multilevel set is sent to the bulb
* Because OZW apparently assumes bulb does not properly report its own state, it also sends another get.
* When bulb starts actually dimming, it sends automatic value update
* On first value update, OZW sends THREE ADDITIONAL GET REQUESTS. It looks like device cannot handle those at once, so OZW of course receives timeout, which starts cascading and bogging down the network with retries

Could an option be added in the future to disable confirmation gets for specific devices (that behave well)?

This is from Home Assistant, which is supposedly still based on OZW 1.4 (my apologies if I should not report HA OZW issues here).

matejdro:
Resolves #1980 

Patch adds additional parameter to every Z-wave value in config xml: `refresh_after_set`. By default it is turned on (no changed behavior by default), but user can turn it off manually for devices that do not need constant verification.

Smart lighting in my Z-wave network consists solely of recent Aeotec and Fibaro devices that all send SwitchMultiLevel report on their own whenever you update them. By disabling redundant get commands, my network's responsiveness has increased significantly, especially when changing multiple lights at once or when changing color on bulbs.

### DIFF
--- a/cpp/src/value_classes/Value.cpp
+++ b/cpp/src/value_classes/Value.cpp
@@ -57,7 +57,7 @@ namespace OpenZWave
 // Constructor
 //-----------------------------------------------------------------------------
 			Value::Value(uint32 const _homeId, uint8 const _nodeId, ValueID::ValueGenre const _genre, uint8 const _commandClassId, uint8 const _instance, uint16 const _index, ValueID::ValueType const _type, string const& _label, string const& _units, bool const _readOnly, bool const _writeOnly, bool const _isSet, uint8 const _pollIntensity) :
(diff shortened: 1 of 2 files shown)


Fishwaldo:
You didn't even compile test this?

Issue #1971, opened 2019-10-09
FIBARO FGT001 Heat Controller does not report Setpoint


michapr:
Hi,
Using Version: 1.6-808-gbb0a4bf7-dirty and checking with OZW Control Panel

Seems that Fibaro Thermostat is reporting the setpoint in two different ways:
1) If changing the Setpoint manual at the thermostat - all is fine and will be reported in OZW Panel (and in Domoticz).
```
2019-10-09 17:20:00.133 Detail, Node008,   Received: 0x01, 0x10, 0x00, 0x04, 0x00, 0x08, 0x0a, 0x60, 0x0d, 0x01, 0x01, 0x43, 0x03, 0x01, 0x22, 0x00, 0xb4, 0x53
2019-10-09 17:20:00.133 Detail,
2019-10-09 17:20:00.134 Info, Node008, Received a MultiChannelEncap from node 8, endpoint 1 for Command Class COMMAND_CLASS_THERMOSTAT_SETPOINT
2019-10-09 17:20:00.134 Detail, Node008, Refreshed Value: old value=20.0, new value=18.0, type=decimal
2019-10-09 17:20:00.134 Detail, Node008, Changes to this value are not verified
2019-10-09 17:20:00.135 Info, Node008, Received thermostat setpoint report: Setpoint Heating 1 = 18.0C
2019-10-09 17:20:00.135 Detail, Node008, Notification: ValueChanged
```
If changing the setpoint via Z-Wave command the new Setpoint value will not be shown in Panel (and not in Domoticz) but will be changed at Thermostat.
The command for return value seems to be different, and not to be detected by OZW in this case:
```
2019-10-09 17:18:46.413 Info, Node008, Value::Set - COMMAND_CLASS_THERMOSTAT_SETPOINT - Heating 1 - 1 - 1 - 24
2019-10-09 17:18:46.414 Detail, Node008, Queuing (Send) MultiChannel Encapsulated (instance=1): ThermostatSetpointCmd_Set (Node=8): 0x01, 0x10, 0x00, 0x13, 0x08, 0x09, 0x60, 0x0d, 0x01, 0x01, 0x43, 0x01, 0x01, 0x01, 0x18, 0x25, 0xe9, 0x06
[...]

Issue #1827, opened 2019-06-06
S2 Tasks


Fishwaldo:
Task List for S2 Support. Lots of Refactoring is going to need to happen before we can start implementing it....

- [ ] Refactor Msg Class - Constructor in the CC's, ValueID Indexes and Command ENUM's, success/failure callbacks
- [ ] Encoding affected ValueID's into each Msg Class so we can fail gracefully on a Failure Callback.  
- [ ] Move Transport related CC into new TransportClass Base. 
- [ ] Msg Class calls directly into TransportClasses for encapsulation
- [ ] Msg Class interactions directly with the WriteNextMsg related functions. (eg, Msg->Write(Driver *) type function)
- [ ] Refactor/Introduce new Msg Class for handling received Messages (Encode in our HandleApplicationCommand related methods?)
- [ ] Received Msg calls directly into Transport Classes for unencapsulation
- [ ] Update all CommandClasses to use new Msg Class instead of raw byte array.
- [ ] Link sent/received messages together.
- [ ] Supervision CC - How will it live together with VerifyChangedValues and our Set/Get process (so older devices without Supervision still work as expected)
- [ ] Figure out how Sequencing of multiencapsulated messages is handled with the TransportCommandClass Specs. 
- [ ] Interactive? Callback to get DSK for Access and Authentication S2? Or should it be provided at the AddNode stage?

ccokme:
Thanks for your works 👍 , i am looking forward this feature.

petergebruers:
> * Figure out how Sequencing of multiencapsulated messages is handled with the TransportCommandClass Specs.

Pointers to spec:

SDS13783 Z-Wave Transport-Encapsulation Command Class Specification
2.3.5 Encapsulation order overview
Text + especially "Figure 1, Encapsulation overview"

And "3.9 Transport Service Command Class, version 2"

I've looked at quite a few Zniffer captures and I haven't found an example of that CC, but I have been thinking this might get used when doing firmware update (which is not in OZW).

xeanhort:
Hi guys, is there any progress/schedule?

I'm interested in S2 and I might help you somehow even though I don't have any idea about how ozw works internally. Is there any developer documentation about how it is structured?

petergebruers:
@xeanhort, thank you for your kind offer. I started working on OZW a fe months ago so I can relate to your question: "how does OZW (code) work".

[...]

PR #2010, opened 2019-11-25
Make set after get optional - Fixed version

matejdro:
[Fixed version of #2008]

Really sorry about that.

I was testing on a linux machine (Raspberry Pi where I run my Z-wave setup from) and then I commited it on a windows machine (my main work PC). Looks like something went wrong while transfering edits from linux to windows.

I have set up Visual Studio now, compiled the code on the Windows machine as well and fixed duplicate method.


petergebruers:
We have a merge conflict, probably because you started your work on this quite a while ago.

What I would do is update your local repository (git pull) then "rebase" your "optional_get_after_set" branch onto master. You will have a "conflict" because a file was edited both by you and someone else. You have to figure out what to change to make your patch compatible.

After fixing the conflict, "force" push your local branch to your remote branch and Github will pickup your changes.

If you want more detailed instructions, I can write something...

matejdro:
Merge conflicts fixed.

petergebruers:
@matejdro thanks. I've fixed some whitespace
@Fishwaldo LGTM, what I checked:
- Builds on my mac
- Test with a single device
- Does not change the semantics ie after this PR still does "get after set" and writes refresh_after_set="true" to ozwcache.
- If I edit a switch binary value in ozwcache, setting refresh_after_set="false", skips get as expected.

Fishwaldo:
So ideally, we would want to add this flag as a CompatOption so it can be set in the distribution config files without having to specify the full <value> block. 
Something like:
```xml
<CommandClass id="xx">
	<Compatibility>
		<NoRefreshOnSet>1</NoRefreshOnSet>
		<NoRefreshOnSet>2</NoRefreshOnSet>
	</Compatibility>
</CommandClass>
```
Where 1 and 2 refer to the Index of the Value you dont need refreshed.

But unfortuantly CompatOptionManager isn't really setup to work on Value's just yet (its on my to long todo list). If you want to have a stab at it, let me know and I'll explain how I see it working... otherwise give me some time.

petergebruers:
Was thinking the same. Thinking out loud. The way this PR works now requires gateway builders and library users to add the function to their software. Eg I could add it to ozwcp as proof of concept. But wouldn't that be cleaner with methods on "manager"? If they do not add it, users may be tempted to edit their ozwcache file. Isn't that something we would like to avoid? Or did I miss part o& the discussion?

Fishwaldo:
As the current Get/Set methods is on the Value Class - its theoretically impossible for a application to Set. So it either requires adding a full ```<value.../>``` block to their distribution config file, or editing ozwcache. Neither is preferable right now. 

And I see this as a enhancement that applies at the device level, rather than a installation level - hence why I'd want to add this as Compatibility block to config files we distribute so all users benefit.

We *could* add the Get/Set methods to the Manager class so apps/users could override it if needed..... But how many users average users would even know if its needed or not? (a legit question - would the Manager interface become some technical debt that no one uses/implements or is it something that would be useful?)

Fishwaldo:
Actually we can implemented Compatibility on values right now. See https://github.com/OpenZWave/open-zwave/blob/7d792c05db181a833f8fba7388e985fa24d6f91a/cpp/src/value_classes/ValueStore.cpp#L86 for a Example. 

Just duplicate that for this option and we are good :)

Question still remains if we need a Manager method to override it.

petergebruers:
I am reasonably sure this would be a _device type specific_ thing, so config file only. I can imagine someone wanted to override it when the device has config set to "do not get after set" and for some reason, device status turns out to be unreliable. But imho the user can either call "refresh value" on his gateway or implement a mild form of polling. I think this question is related to the fact that some dimmers and blind aka roller shutter modules report value(s) while they ramp up (and you always suggest to look at Target Value, CC V3 and higher iirc). Doing a "get" right after "set" will make the device appear to have some (random) intermediate level. On HASS forums the suggested workaround now is to add a delay of about 5 seconds then do a get.

If a device sends an unsolicited report with state, after some change, then doing this "get" seems unnecessary indeed. I have not really scrutinized the way other gateways do it, but I can look at eg Zniffer capture of a HC2. I am guessing quite a few Z-Wave gateways do not implement get after set for binary set or multilevel set but silently assume the devices accepts the message... Which, strictly speaking, is not true in all cases. An example that comes to my mind: multilevel 0xFF turns on the last dimming level > 0, so the get will return the actual dimming level. I never checked, but I bet an an unsolicited report might have some intermediate value while it ramps up, then a bit later it should report the "set" value. I should really check that with Zniffer or OZW_Log.

Sorry for rambling about this but I am trying to see the bigger picture here.

@matejdro do you have an example (or two) of a device on which you would set refresh_after_set="false" and why?

matejdro:
Two example devices:

1. Aeotec Nano Switch/Dimmer

    This device supports two ways of toggling the light, via connected physical switch and via z-wave network:

    Physical switch:
        1. User toggles switch
        2. Device self-reports its state to OZW

    Z-wave:
        1. OZW sends set state
        2. OZW sends get state
        3. Device self-reports its state

    While self-reporting can be turned off, that also turns off self reporting for physical switch (meaning OZW would not know light is actually toggled when user uses physical switch), so the only proper way to remove the redundancy is to get rid off get command.

2. Aeotec Z-wave Color 6 Bulb

    Since this bulb is dimmable, user can have it "fade" instead of just turning on/off. In that case:

    1. OZW sends set state
    2. OZW sends get state
    3. Device self-reports its state over time until it reaches desired brightness.

    Note that in this case, get actually reports misleading number, since bulb can be mid-fade when response to get is returned. So if user turns off self-reporting, OZW will have wrong brightness numbers. Once again, only proper way to remove the redundancy is to get rid off get command.


Note that both devices have config option to turn off self reporting. So just knowing device type is not enough (get might still be needed if user turns off self reporting).

This does not apply to all devices though. Changing heating modes on eurotronic Spirit Z-Wave thermostat will not trigger any self reporting and thus get is mandatory here.

petergebruers:
@matejdro thank you for your detailed examples.

> 1. Aeotec Nano Switch/Dimmer
...
> 1. While self-reporting can be turned off, that also turns off self reporting for physical switch

I am not familiar with this device so I had to look it up... You're talking about parameter 80 of the Dual Nano Switch, I quote from the manual, right?

To set which notification would be sent to the associated nodes in association group 1 when the state of output load is changed.

0 = Nothing
1 = Hail CC
2 = Basic Report CC
3 = Hail CC when using the external switch to switch the loads. --> default.

I agree with you, it makes sense to disable "get" in OZW and leave the device to report. I cannot think of a reason to set this to "0 = Nothing" because then you would have to poll the device. This makes me wonder... Why does this option exist?

I own a lot of Fibaro devices and afaik none have a parameter to disable this for "main action" like on/off. They do have settings to turn down the volume of Meter reports.

> Aeotec Z-wave Color 6 Bulb
> 
> Since this bulb is dimmable, user can have it "fade" instead of just turning on/off. In that case:
> 
> 1. OZW sends set state
> 2. OZW sends get state
> 3. Device self-reports its state over time until it reaches desired brightness.

That is what HASS users complain about and that is iirc the way Fibaro Roller Shutter 3 and Dimmer 2 work.

> Note that both devices have config option to turn off self reporting. So just knowing device type is not enough (get might still be needed if user turns off self reporting).

Indeed.

I expect parameter settings to available on all software based on OZW so I would say... Although the user can change this setting it is easier to change that, than having to add rules to OZW to handle option "0 = Nothing" for this Nano device for example.

> Changing heating modes on eurotronic Spirit Z-Wave thermostat will not trigger any self reporting and thus get is mandatory here.

That is a rather modern device and it might fall into the same category as the Fibaro FGT-001

See "FIBARO FGT001 Heat Controller does not report Setpoint #1971"

The gist... Those devices support COMMAND_CLASS_SUPERVISION

This is on the todo list, under "S2 Tasks":

#1827

Public spec SDS13783-12 Z-Wave Transport-Encapsulation Command Class Specification

```
3.7 Supervision Command Class, version 1
3.7.1 Terminology
The controlling application initiates an operation by sending a Supervision encapsulated command to a supporting node. The supporting node returns an immediate confirmation for the reception while advertising its ability to perform the requested operation. The confirmation may advertise the application status <Working>, <Success>, <No Support> or <Fail> depending on the condition of the supporting node. One or more application status updates may follow later to report problems, delays or the completion of the requested operation.
```

Thermostats are interesting because you can ask eg a certain mode that does not exist, or set a temperature outside its available range. So get makes sense but in case of the FGT-001 a get throws a "device busy" at you (which does not happen in case supervision CC is used).

matejdro:
> Why does this option exist?

Well, without this get patch it would actually make sense to disable self reporting if you do not use physical switch feature to reduce number of packets when toggling multiple lights at once.

Fishwaldo:
@matejdro I've just converted it to use CompatOptions instead. Can you do some testing:
For the config file of your device - please add the following:
```xml
<CommandClass id="xx">
	<Compatibility>
		<NoRefreshAfterSet index="yy">true</NoRefreshAfterSet>
		<NoRefreshAfterSet index="zz">true</NoRefreshAfterSet>
	</Compatibility>
</CommandClass>
```
Where xx equals the CommandClass of the Value you don't want refreshed
and yy / zz are the index's of the Values you don't want refreshed.

Make sure any ozwcache*.xml is wiped. You should see a message in the logs along the lines of 
```Setting NoRefreshAfterSet Flag on Value %d for CC %s``` after restarting

You can also test you got the config file correct by running ```make xmltest```

**Comment by matejdro** [comment]
I can do that, but it will take few days unfortunately (a bit busy at the moment).

**Comment by matejdro** [comment]
Thanks for the changes @Fishwaldo and sorry for the delay. I  have tested the changeset and I can confirm it's working with `<Compatibility>` tag (tested with ozwcp).

**Comment by Fishwaldo** [comment]
Thanks!

**Comment by danielbrunt57** [comment]
> I am reasonably sure this would be a _device type specific_ thing, so config file only. I can imagine someone wanted to override it when the device has config set to "do not get after set" and for some reason, device status turns out to be unreliable. But imho the user can either call "refresh value" on his gateway or implement a mild form of polling._after_set="false" and why?

I am trying to implement this for HAB iblinds.  They do not take kindly to get after set and become unresponsive.  I've added the NoRefreshAfterSet compatibility flag but now the device status never updates.  How would I "refresh value" for CC37/CC38 and/or implement a mild form of polling?  I've tried publishing the following to no avail:
```
topic: OpenZWave/1/command/refreshvalue/
payload: '{ "ValueIDKey": 357138449 }'
```
```
topic: OpenZWave/1/command/requestnodedynamic/
payload: '{ "node": 21 }'
```

PR #2013, opened 2019-11-25
Add GetValueEndPoint() to Manager

petergebruers:
Introduces helper function:

uint8 Manager::GetValueEndPoint(ValueID const &_id)

The purpose is
- To allow printing of "End Point" for diagnosis. End Point is what is used in conversation with a device, not "Instance".
- To allow selection of Values with certain End Point for use in "Multi Channel Association". This class does not accept "Instance".

Adds tests to cpp/test/Manager_test.cpp - verifies a (fake) mapping and tests exception thrown.

Adds "TestHelper" as a friend to allow testing of existing OZW classes without refactoring and without having to start a "real" Manager and Driver (which would have its merits but would be to involved and heavy)

If this PR gets accepted, I will update ozwcp to display the "endpoint".

The "TestHelper" can (will) be re-used to test a recent commit, 79663e1e
by gizmocuz, "Making GetValue code more consistent, Releasing values", mainly as a proof-of-concept on how to test things.

I am not totally new to testing, but I am new to using "google test" and testing of C++ code so please give an honest opinion 🙂

### DIFF
--- a/cpp/src/Driver.h
+++ b/cpp/src/Driver.h
@@ -77,6 +77,10 @@ namespace OpenZWave
 		class Msg;
 		class TimerThread;
 	}
+	namespace Testing
+	{
+		class TestHelper;
+	}
 
 	/** \brief The Driver class handles communication between OpenZWave
 	 *  and a device attached via a serial port (typically a controller).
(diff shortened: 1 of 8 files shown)


Fishwaldo:
Thanks. Will review soon.

petergebruers:
Wasn't aware that Github would close my PR if I removed the branch. Was doing some spring cleaning 🙂...

PR #2014, opened 2019-11-26
Add missing Philio relay

quarcko:
I have a Philio relay PAN06 which has ID=13.
Please add it to database, thank you!


petergebruers:
Thanks! Edits cherry-picked and fix-up revision.

PR #2016, opened 2019-11-28
Make metaData function static

jano42:
Change MetaDataField methods as static. (functions to convert MetaDataField to/from string).
This allow an easier usage of MetaDataField (calling code can now have the name of the MetaDataField and avoid a duplicated declaration)


jano42:
@Fishwaldo : check fails from an unknown reason for me

Fishwaldo:
Failing checks are known issue right now. (Github Actions are nerfed for pull requests unfortunately). 

I don’t quite follow your reasoning here. Can you show me a code sample how this helps. 

I’m also weary of breaking existing apps (we try to stay API stable) and not sure how flicking to static would impact apps that are calling from the manager instance instead.

jano42:
In fact here is my code:

````c++
void COpenZWaveController::getNodeMetaData(uint32 homeId, uint8 nodeId, shared::CDataContainer& result)
{
   for (int i = 0; i < OpenZWave::Node::MetaData_Invalid; ++i)
   {
      try
      {
         const auto mdi = static_cast<OpenZWave::Node::MetaDataFields>(i);
         std::string metaName = OpenZWave::Node::GetMetaDataString(mdi);
         if (metaName.empty())
         {
            i = OpenZWave::Node::MetaData_Invalid;
         }
         else
         {
            std::string metaValue = OpenZWave::Manager::Get()->GetMetaData(homeId, nodeId, mdi);
            if (!metaValue.empty())
            {
               result.set(metaName, metaValue);
            }
         }
      }
      catch(...)
      {
         break;
      }
   }
}
````

The aim : collecting all the node metaData without specifying each entry. => in case of OZW evolution with new meta, nothing to do for me. Then another reason is to get the metaData name, I don't wnat to add the string to my code, has it is already defined in your.

The last reason for making those methods static, is that they are "static" ! (no interaction with class instance,... => my resharper tools recommandation told me to make it static)

Finaly adding static, just allow other code to use it without getting node instance (OpenZWave::Node::GetMetaDataString() ), and then adding static do not change anything to current code.

Fishwaldo:
Ahhhh.

So we are actually trying to move away from a app having to access any class other than Manager (Node is a hangover due to some of these ENUM's though). So My preference would be to add a Method to the Manager Class to retrieve the metadata names, and that can call into the static Node Method to get it

If you could add a method to Manager.h/cpp I'll merge the PR.

jano42:
Ok you're right. This is a good idea to have only Manager class as public api.
I've added methods as requested.

petergebruers:
@jano42 typo I guess:

metaDataFiledtoParse -> metaDataFieldtoParse

Fishwaldo:
I put some comments. Spelling Errors there :)

PR #2023, opened 2019-12-05
Fix comparison uint8 with -1, also test if controller node is 0.

petergebruers:
lldb complained:
Driver.cpp:2701:43: error:
      result of comparison of constant -1 with expression of type
      'uint8' (aka 'unsigned char') is always false
      [-Werror,-Wtautological-constant-out-of-range-compare]

Add some info to debug logging


petergebruers:
Pff had a brainfreeze... Working on "master" branch so mixed up commits with PR 2024... Fixed it...

Issue #2020, opened 2019-11-29
Problems with ZME_WALLC-S after Update from 1.4 to 1.6 - states missing


fbatsb17:
Hi, 
I am using Zwave in IOBroker. Everything's fine until updating iobroker.zwave from 1.4 to 1.6. So I would say thank you for your work. By running the update, also the openzwave Version changes from 1.4 to 1.6.
After the update, the Group "CLASSES91" was missing and only two states (Scene_Count_1 and Scene_Reset_Timeout) were found in "CENTRAL_SCENE". Many different tries of installing the Zwave Modul leading to the same Problem. Especially state Scene_Number_1 is still missing - which I was using to read the Commands, coming from the wall plug.
In the Logfile I found:
```
2019-11-07 19:30:54.529 Info, Node005, Received Central Scene set from node 5: scene id=4 with key Attribute 0. Sending event notification.
2019-11-07 19:30:54.529 Warning, Node005, No ValueID created for Scene 4
2019-11-07 19:30:54.529 Warning, Node005, CommandClass COMMAND_CLASS_CENTRAL_SCENE HandlerMsg Returned False

``` 
Discussions in the Github and in the forum of iobroker indicates on same Problems on "Devolo wall switch" and "heatit z-push button 8" (take a look at https://github.com/ioBroker/ioBroker.zwave/issues/99 and https://forum.iobroker.net/topic/26440/probleme-nach-zwave-update-mit-zme_wallc-s/28).

Hope, anyone get's a solution for this Problem.
Thanks in advance

petergebruers:
I own a ZME_WALLC-S but I am not using this device... This means, I don't know it very well but I was able to make a test.

But first... I am a volunteer and you have to be very patient with me, I am (very) slow...

I can tell you what is going on, when you see this: "No ValueID created for Scene 4" the "interview process is not finished yet and OpenZWave does no know yet what that wall switch can "do".

This is a battery (CR2032) device and it does not stay awake very long... And there is something which I have to investigate, OZW is querying certain data and the device does not respond which makes OZW time out, then the device goes back to sleep. In other words, you have to keep waking it (push 4 buttons for 5 seconds, led turns green, then press key 2). This is in OZW_Log:

```

2019-11-29 20:00:07.656 Info, Node027, MultiInstanceCmd_Get for COMMAND_CLASS_BASIC
[...]

PR #2024, opened 2019-12-05
Don't request Instances is Multichannel is "after mark"

petergebruers:
IMHO if the Multichannel CC is net before, but only after the "mark" then the device is not multi channel.

Adds log:

2019-12-05 10:34:57.454 Detail, Node004, Skipping RequestInstances() because MultiChannel CC is "after mark"

Fixes "Problems with ZME_WALLC-S after Update from 1.4 to 1.6 - states missing" #2020 by not doing 17 requests that all time out...


Fishwaldo:
Thanks

There is probably a few other CC’s that a similar check should be made for. Or separate the list of CC’s into controlled versus controlling

PR #2025, opened 2019-12-05
Fix comparison uint8 with -1, also test if controller node is 0.

petergebruers:
Ported from OZW 1.6

lldb complained:
Driver.cpp:2701:43: error:
      result of comparison of constant -1 with expression of type
      'uint8' (aka 'unsigned char') is always false
      [-Werror,-Wtautological-constant-out-of-range-compare]

Add some info to debug logging

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -550,7 +550,7 @@ bool Driver::Init
 		return false;
 	}
 
-	m_Controller_nodeId = -1;
+	m_Controller_nodeId = 255;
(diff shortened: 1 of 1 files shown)


petergebruers:
Shall I open a PR agains HASS fork?

Fishwaldo:
Sure.

petergebruers:
Done... Also had to fix Makefile to make that fork build with lldb (C++11)

https://github.com/home-assistant/open-zwave/pull/16

PR #2026, opened 2019-12-05
Make flock(m_hSerialController... actually do something

petergebruers:
I noticed... I could run many instances of MinOZW on the same port, on my Mac, or MinOZW and "picocom" on the same port...

This will make OZW retry as if it failed to open the port.

Prints this if it cannot get an exclusive lock:

ERROR: Cannot get exclusive lock for serial port /dev/cu.usbmodem14101. Error code 35

### DIFF
--- a/cpp/src/platform/unix/SerialControllerImpl.cpp
+++ b/cpp/src/platform/unix/SerialControllerImpl.cpp
@@ -188,9 +188,10 @@ namespace OpenZWave
 					goto SerialOpenFailure;
 				}
 
-				if (flock(m_hSerialController, LOCK_EX | LOCK_NB) == -1)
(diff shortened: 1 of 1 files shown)


Fishwaldo:
Lets see what, if anything breaks here. :)

petergebruers:
> Lets see what, if anything breaks here. :)

Yes, I keep an eye on this. I know a case of a guy who started 2 instances of HASS without realizing it (probably a scripting error). Everything looked normal except all the "out of frame flow" messages and barely working OZW. a "ps" revealed every process was loaded twice. After this commit, one of them won't start the driver.

petergebruers:
Should I write about this change on "google groups"? Or any other place?

Fishwaldo:
I don’t think it’s big enuf for that. Anyone of the mailing list should have seen the commit.

PR #2027, opened 2019-12-06
config CT32: correct instance/endpoint mapping

kpine:
Enable mapping root to endpoint. The root and endpoint 1 seem to have
identical functionality. Enabling this mapping removes the duplicate
values (battery, setpoint, etc).

Temperature and humidity are still duplicated since those are provided
by endpoint 2 as well.

Endpoint 2 doesn't seem to get any unsolicited updates, even after enabling reporting thresholds for the temperature (param 1) and humidity (5). I don't quite understand what it's purpose is for. Refreshing the value does work.


Fishwaldo:
Thanks!

PR #2031, opened 2019-12-08
Added config for GE (Jasco) 46203 In-Wall Smart Dimmer.

radovanbauer:
This is the 2nd gen of the dimmer wih QuickFit and SimpleWire enhancements.

This configuration is based on the xml spec from
https://products.z-wavealliance.org/products/3323 . I have tested all
configuration parameters on an actual switch.


petergebruers:
Thanks!

Issue #2038, opened 2019-12-18
Ozwcache corruption leading to memory corruption and crash


petergebruers:
The problem might be less scary than the title suggests, because this happened while I was debugging OZW, frequently starting and stopping, adding and deleting nodes and so on... It might not happen very often "in real life" but I think it was worth investigating/mitigating.

At one point, I got a SIGSEGV while testing ozwcp or MinOZW, similar to this:

`./MinOZW: line 11: 89769 Segmentation fault: 11`

It halted in "vector" code while retrieving the list. That's odd. I added some logging:

```cpp
	ValueList::Item const *ValueList::GetItem() const
	{
		try
		{
			Log::Write(LogLevel_Warning, "m_items.size: %zu, requesting m_valueIdx: %d %s", m_items.size(), m_valueIdx, GetID().GetAsString().c_str());
			if (m_items.size()>255)
			{
				Log::Write(LogLevel_Warning, "Crash Ahead!");
			}
			return &m_items.at(m_valueIdx);
		}
```

That happened when I was retrieving a List value and after adding debugging code I got an implausible size for the list size, something like:

`18446744073684291199`

That looked a lot like memory corruption, so I decided to find the root cause.

The Value Tyoe is determined after loading ozwcache and I noticed...

```xml
<CommandClass id="112" name="COMMAND_CLASS_CONFIGURATION">
	<Compatibility />
	<State>
		<CCVersion>1</CCVersion>
		<InNif>true</InNif>
		<StaticRequests>4</StaticRequests>
	</State>
	<Instance index="1" />
	<Value type="list" genre="config" instance="0" index="1" label="Ambient Light Intensity Calibration" units="" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="8" max="255" vindex="1" size="1">
		<Help>Adapt the sensitivity of the motion detection.</Help>
		<Item label="High Sensitivity" value="8" />
		<Item label="Normal Sensitivity" value="12" />
		<Item label="Lower Sensitivity" value="128" />
		<Item label="Lowest Sensitivity" value="255" />
	</Value>
	<Value type="short" genre="config" instance="0" index="1" label="Ambient Light Intensity Calibration" units="second" read_only="false" write_only="false" verify_changes="false" poll_intensity="0" min="5" max="600" value="30">
		<Help>Adapt the sensitivity of the motion detection.</Help>
	</Value>
[...]

PR #2040, opened 2019-12-18
Add ValueID::GetAsString() for debugging purposes

petergebruers:
This function allows to convert a valueID to human readable form

Example (from tests):

ValueID(static_cast<uint64>(0x1), 0x02, ValueID::ValueGenre_System, 0x04, 5, 6, ValueID::ValueType_String).GetAsString()

"HomeID: 0x00000001, ValueID: (Id 0x0006000002c10057, NodeID 2, Genre system, CC 0x04, Instance 5, Index 6, Type string)"

Related to https://github.com/OpenZWave/open-zwave/issues/2038 "Ozwcache corruption leading to memory corruption and crash"

If this PR is OK I would like to use it to add warnings described in 2038, and also, later, add it to the exceptions thrown in Manager if something is wrong about the ValueID in the request. This would help gateway builders to better understand the cause of the exception.


petergebruers:
I've pushed a commit because it think this PR makes more sense if I provide an application for GetAsString.

As discussed here  #2038 "Ozwcache corruption leading to memory corruption and crash"

I "fake" a form of cache corruption by starting MinOZW

Then I take this value under Node 1

`<Value type="int" genre="system" instance="1" index="0" label="Loaded Config Revision"`

And change the type to

`<Value type="list" genre="system" instance="1" index="0" label="Loaded Config Revision"`

This is of course a rather "synthetic" example but easy to reproduce. Then OZW will log:

`2019-12-19 16:36:23.033 Error, Node001, Node::GetValue called with: HomeID: 0xd96a77cd, ValueID: (Id 0x0000000001dc8013, NodeID 1, Genre system, CC 0x72, Instance 1, Index 0, Type int) but GetValueStore returned: HomeID: 0xd96a77cd, ValueID: (Id 0x0000000001dc8014, NodeID 1, Genre system, CC 0x72, Instance 1, Index 0, Type list)
2019-12-19 16:36:23.033 Info, Node001, Dropping Notification as ValueID does not exist`

Note... In this PR I also return nullptr if the IDs don't match, leading to the `Dropping Notification as ValueID does not exist`

PR #2043, opened 2019-12-21
Update Kwikset Convert

jeffsanicola:
Update existing device:
Manufacturer ID=0090
Product:
config=kwikset/914c.xml, id=0446
name=Convert Electronic Deadbolt
type=0003

Convert doesn't support all the features the Kwikset SmartCode locks do, such as User Codes
Split out the Convert specific functions into its own file
Bumped manufacturer_specific.xml version


petergebruers:
Thanks!

chickenandpork:
I believe the Kwikset Signature also identifies as the same IDs, and matches the same capabilities.  It’s possible the Convert is merely the controller side without the Signature’s deadbolt.

jeffsanicola:
> I believe the Kwikset Signature also identifies as the same IDs, and matches the same capabilities. It’s possible the Convert is merely the controller side without the Signature’s deadbolt.

Certainly looks that way to me too.

Is there a preferred method for handling this type of ambiguity in products?

chickenandpork:
I would assume it’s just a product marketing differentiation; what about just changing the description to be “Convert/Signature” or similar so that the human looking at it is assured the correct device was discovered ?

PR #2044, opened 2019-12-21
Port Fibaro Walli FGWOx-011 from 1.6

gyorsok:


Fishwaldo:
1.4 is no longer maintained. We only accept updates on 1.6

PR #2048, opened 2019-12-25
Add support for Radio Thermostat CT32 for Devices with ID 102

aaronseaton:
This issue was solved on the HA Community forums [here](https://community.home-assistant.io/t/radio-thermostat-ct32-target-temperature-setpoint-not-working/157153/2).  The issue was solved by duplicating the existing Radio Thermostat CT32 entry on line 10 of the manufacturer_specific.xml to also cover devices with ID 102 as follows:

`		<Product type="2002" id="0100" name="CT32 Thermostat" config="2gig/ct32.xml"/>
		<Product type="2002" id="0102" name="CT32 Thermostat" config="2gig/ct32.xml"/>`

Please add this additional line to ensure future compatibility for this common device in future versions of HA and HASSIO.


Fishwaldo:
Firstly - this PR is invalid. It’s brining it’s lots more than just your changes. 
Second. We are only maintaining config for 1.6. 1.4 is not being updated.

PR #2050, opened 2019-12-25
Add fibaro flood sensor

ymanton:
Came across a Fibaro flood sensor variant with type/id=0b01/2003.

Also updated `distfiles.mk` for files that were missed by a previous commit.


Fishwaldo:
Thanks!

PR #2051, opened 2019-12-26
Add new device Aeotec zwa021 Thermostatic Valve

benvm:
I added the Aeotec ZWA02 which is a Thermostatic Radiator Valve.
It looks like a clone of the Eurotronic Spirit, so I mostly copied the data from there.
I've been using this config successfully in Home Assistant.

This is my first contribution. I hope everything is correct.


nechry:
thank @benvm I just fix the ProductPic path, otherwise everything is ok. I also tested with my ZWA021 device and also no issue so far.

PR #2064, opened 2020-01-09
Update ID and type for HomeSeer sensors

funwhilelost:
These were misidentified in my HomeAssistant setup but it appears it's just needs leading zeros.


Fishwaldo:
Thanks!

PR #2065, opened 2020-01-09
Update configuration parameter descriptions for Aeotec ZW175

gertvdijk:
```
I own a few of these devices (with firmware 1.02) and found the
configuration applied did not match the description in Open-Zwave.

- Parameter 101 controls the periodic power reporting. The description
  before this change was all wrong. The Aeotec manual updated November 2019
  [1] appears to be wrong about bits 2 & 3 (values 4 and 8); they seem
  reversed in practice.
- Parameter 111 controls the interval for the periodic power reporting. The
  Aeotec manual updated November 2019 [1] mentions that the minimal
  interval is 30 seconds, not 600. The factory default is 600, though.

[1]: https://aeotec.freshdesk.com/support/solutions/articles/6000219911-smart-switch-7-user-guide-
     (version "Modified on: Wed, 13 Nov, 2019 at 10:45 PM")
```

### DIFF
--- a/config/aeotec/zw175.xml
+++ b/config/aeotec/zw175.xml
@@ -1,4 +1,4 @@
-<Product Revision="2" xmlns="https://github.com/OpenZWave/open-zwave">
+<Product Revision="3" xmlns="https://github.com/OpenZWave/open-zwave">
     <MetaData>
         <MetaDataItem name="Name">ZW175 Smart Switch 7</MetaDataItem>
         <MetaDataItem name="Description">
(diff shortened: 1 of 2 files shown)


config/aeotec/zw175.xml
```diff
@@ -146,10 +147,10 @@
         <Value genre="config" index="101" label="Configure which meter reading will be periodically report via Lifeline" type="int" units="" value="15">
             <Help>
                 Defines the type of report sent for reporting group 1.
```
gertvdijk: Note for reviewer: the description of the bits (L156-L160 in the base) below appear correct to me, it's just the integer representation that appear wrong.

gertvdijk:
I see an unrelated CI failure "Password is not specified"... :frowning:

petergebruers:
Makes sense. Thanks!

PR #2066, opened 2020-01-09
Update Hank HKZW-MS02 configuration parameter 18 description

gertvdijk:
According to the product manual, this parameter controls how much
additional delay will be used to control directly associated devices, and
it has nothing to do with a luminance level (this is not a luminance
sensor).


petergebruers:
Thanks!

PR #2076, opened 2020-01-22
CT32 thermostat config file updates

kpine:
Fix the 1) setpoint detection and some 2) incorrect config parameters.

1. Disable the AltTypeInterpretation compat flag. When this flag is enabled (the default), OZW will process the thermostat setpoint supported report bitmask value as Interpretation B. This causes the Energy Save Heating (A bit 1.7) and Energy Save Heating (A bit 2.0) to be interpreted as Furnace (B bit 1.7) and Dry Air (B bit 2.0). The spec says all implementations must comply with Interpretation A, so I'm not quite sure why it isn't the default.

2. Some config params had incorrect or minimal help text, and incorrect default values. The Temperature Swing setting was missing several values.

### DIFF
--- a/config/2gig/ct32.xml
+++ b/config/2gig/ct32.xml
@@ -35,7 +35,7 @@
     </Compatibility>
   </CommandClass>
   <CommandClass id="112">
-    <Value genre="config" index="1" label="Temperature Reporting Threshold" max="4" min="0" size="1" type="list" units="" value="0">
+    <Value genre="config" index="1" label="Temperature Reporting Threshold" max="4" min="0" size="1" type="list" units="" value="2">
       <Help>
         The Temperature Reporting Threshold Configuration Set Command sets the reporting threshold for changes in the ambient temperature as detected by the thermostat.
       </Help>
@@ -116,9 +116,7 @@
       <Item label="HVAC: Normal, Aux Stages: 2, Aux Setup: Elec, Heat Pump Stages: 2, Cool Stages: 2" value="19005954"/>
       <!-- 0x01,0x2,0x2,0x02,0x02 -->
       <Item label="HVAC: Heat Pump, Aux Stages: 2, Aux Setup: Elec, Heat Pump Stages: 2, Cool Stages: 2" value="35783170"/>
-    
-    
-    <!-- 0x02,0x2,0x2,0x02,0x02 -->
+      <!-- 0x02,0x2,0x2,0x02,0x02 -->
     </Value>
     <Value genre="config" index="3" label="Utility Lock" max="1" min="0" size="1" type="list" units="" value="0">
       <Help>
@@ -136,9 +134,9 @@
       <Item label="C-Wire" value="1"/>
       <Item label="Battery" value="2"/>
     </Value>
-    <Value genre="config" index="5" label="Humidity Reporting Threshold" max="3" min="0" size="1" type="list" units="" value="0">
+    <Value genre="config" index="5" label="Humidity Reporting Threshold" max="3" min="0" size="1" type="list" units="" value="2">
       <Help>
-        The Temperature Reporting Threshold Configuration Set Command sets the reporting threshold for changes in the ambient temperature as detected by the thermostat.
+        The Humidity Reporting Threshold Configuration Set Command sets the reporting threshold for changes in the ambient humidity as detected by the thermostat.
       </Help>
       <Item label="Disabled" value="0"/>
       <Item label="3% RH" value="1"/>
@@ -156,13 +154,17 @@
       <Item label="Disabled" value="0"/>
       <Item label="Enabled" value="1"/>
     </Value>
-    <Value genre="config" index="7" label="Thermostat Swing Temperature" max="8" min="1" size="1" type="list" units="" value="0">
+    <Value genre="config" index="7" label="Thermostat Swing Temperature" max="8" min="1" size="1" type="list" units="" value="2">
       <Help>
-        The Auxiliary/Emergency configuration command enables or disables auxiliary/emergency heating in the thermostat. Auxiliary/emergency heating is only available if the thermostat is configured in heat pump mode and with at least one stage of auxiliary heating. This command enables auxiliary / emergency heating when the thermostat is in Auto mode. The Thermostat Set Mode command with mode Auxiliary/Emergency Heat will enable emergency heating but only if the thermostat is in Heat
-        mode. This command should only be used on thermsotats that support Auxiliary/Emergency Heat thermostat mode.
+        The Temperate Swing (HVAC cycling rate) is the desired variance in temperature between the thermostat setting and the room temperature required before the heating or cooling system will turn on.
       </Help>
-      <Item label="0.05F" value="1"/>
-      <Item label="0.1F" value="2"/>
+      <Item label="0.5F" value="1"/>
+      <Item label="1.0F" value="2"/>
+      <Item label="1.5F" value="3"/>
+      <Item label="2.0F" value="4"/>
+      <Item label="2.5F" value="5"/>
+      <Item label="3.0F" value="6"/>
+      <Item label="3.5F" value="7"/>
       <Item label="4.0F" value="8"/>
     </Value>
     <Value genre="config" index="8" label="Thermostat Differential Temperature" max="32767" min="2" size="2" type="list" units="F" value="4">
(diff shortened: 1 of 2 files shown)


config/2gig/ct32.xml
```diff
@@ -115,9 +117,7 @@
       <Item label="HVAC: Normal, Aux Stages: 2, Aux Setup: Elec, Heat Pump Stages: 2, Cool Stages: 2" value="19005954"/>
       <!-- 0x01,0x2,0x2,0x02,0x02 -->
       <Item label="HVAC: Heat Pump, Aux Stages: 2, Aux Setup: Elec, Heat Pump Stages: 2, Cool Stages: 2" value="35783170"/>
-    
-    
-    <!-- 0x02,0x2,0x2,0x02,0x02 -->
+      <!-- 0x02,0x2,0x2,0x02,0x02 -->
     </Value>
     <Value genre="config" index="3" label="Utility Lock" max="1" min="0" size="1" type="list" units="" value="0">
```
kpine: The manual says this is write only, but I don't know how I would verify that a GET doesn't work? The OZWCP does not let me refresh individual values.

config/2gig/ct32.xml
```diff
@@ -155,13 +155,17 @@
...
+      <Item label="0.5F" value="1"/>
+      <Item label="1.0F" value="2"/>
+      <Item label="1.5F" value="3"/>
+      <Item label="2.0F" value="4"/>
+      <Item label="2.5F" value="5"/>
+      <Item label="3.0F" value="6"/>
+      <Item label="3.5F" value="7"/>
       <Item label="4.0F" value="8"/>
     </Value>
     <Value genre="config" index="8" label="Thermostat Differential Temperature" max="32767" min="2" size="2" type="list" units="F" value="4">
```
kpine: This parameter has different data for the Set and Report command definitions. The Set takes a single differential mode (heat or cool) and a differential. The Report returns both temperatures. Doesn't look like there is any way to actually represent that in OZW.

Fishwaldo:
Previously versions of the spec were not clear how to interpret it, and a “recent” version Clarified it’s type B. 

I was hesitant to change as it means all those devices that use A would “break”. But yeah, maybe time for a flag day and set the default to B and users with broken devices will need to update the config files.

Fishwaldo:
Ehhh. Swap A and B in my comment. Haha

config/2gig/ct32.xml
kpine: Since I can't add comments on non-modified lines, the Parameter 10 below is a mess. It has two entries. Looks like OZW chooses the first. The first is the Report format. It returns two bytes, while the Set takes 4 bytes. Not sure how I would fix that one.

kpine:
> Previously versions of the spec were not clear how to interpret it, and a “recent” version Clarified it’s type B.
> 
> I was hesitant to change as it means all those devices that use A would “break”. But yeah, maybe time for a flag day and set the default to B and users with broken devices will need to update the config files.

I assumed that was why, but I was hoping to get some clarification on that. The other flag I have issue with is the "Base" one, which defaults to 1. It seems like most config files are already overriding it to 0. I presume there were some early thermostats that were 1-based.

Yes, I can understand the hesitation. It could certainly change the behavior some people are accustomed to. On the other hand, I would expect (they are certified, right??) modern thermostats to follow the spec. When they do, it means they don't function properly w/o creating a config file and setting the flags properly.

My guess is most people haven't been bothered by this because either a) the thermostats only support heating and/or cooling, or b) they just don't use the affected setpoints.

Would it be useful if I submitted an issue for the default flag values, to at least track it as a known thing?

config/2gig/ct32.xml
Fishwaldo: You can add ``` write_only="true"``` to the Value entity to set that up.

config/2gig/ct32.xml
Fishwaldo: Seriously what? thats braindead :(

No way to handle it. 

Different sizes based on Set/Report is against the specifications. (it doesnt make it).

config/2gig/ct32.xml
```diff
@@ -217,14 +221,14 @@
...
-        Simple UI Mode Enable/Disable
+        If the value is set to Disable then Normal Mode is enabled. If the value is set to Enable then Simple Mode is enabled.
       </Help>
       <Item label="Enable" value="0"/>
       <Item label="Disable" value="1"/>
     </Value>
     <Value genre="config" index="12" label="Multicast" max="1" min="0" size="1" type="byte" units="" value="0">
       <Help>
-        Multicasting Enable/Disable
+        If set to 0, multicast is disabled, if set to 1, will enable the multicast.
```
Fishwaldo: You could potentially change this to a list (or bool)

Fishwaldo:
Let me review how many devices potentially are affected and I'll look at switching the default to A and set the other devices to Type B if they don't already force A.

config/2gig/ct32.xml
kpine: I'll do that, I verified a SET will be ignored.

config/2gig/ct32.xml
kpine: I thought of doing that, but didn't want to break anything. I'll go ahead and update it considering the state of all the others.

kpine:
> Let me review how many devices potentially are affected and I'll look at switching the default to A and set the other devices to Type B if they don't already force A.

I thought that would be a reasonable approach. Consider also defaulting Base to 0 instead of 1 also. There are a lot of config files setting it to 0. I'm assuming there are more of those than there are of config files not setting it though...

Fishwaldo:
Question - Does the documentation mention what versions of CC's the CT32 supports?

kpine:
Nothing that I can find. Here are the versions that are reported in the cache file. Also attaching the complete file in case it's useful. I'm still in the process of making additional updates, hopefully can do that this weekend.
[ozwcache_0xf4e76156.xml.txt](https://github.com/OpenZWave/open-zwave/files/4143245/ozwcache_0xf4e76156.xml.txt)


```
<CommandClass id="32" name="COMMAND_CLASS_BASIC">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="49" name="COMMAND_CLASS_SENSOR_MULTILEVEL">
		<CCVersion>5</CCVersion>
</CommandClass>
<CommandClass id="64" name="COMMAND_CLASS_THERMOSTAT_MODE">
		<CCVersion>2</CCVersion>
</CommandClass>
<CommandClass id="66" name="COMMAND_CLASS_THERMOSTAT_OPERATING_STATE">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="67" name="COMMAND_CLASS_THERMOSTAT_SETPOINT">
		<CCVersion>2</CCVersion>
</CommandClass>
<CommandClass id="68" name="COMMAND_CLASS_THERMOSTAT_FAN_MODE">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="69" name="COMMAND_CLASS_THERMOSTAT_FAN_STATE">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="90" name="COMMAND_CLASS_DEVICE_RESET_LOCALLY">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="94" name="COMMAND_CLASS_ZWAVEPLUS_INFO">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="96" name="COMMAND_CLASS_MULTI_INSTANCE/CHANNEL">
		<CCVersion>3</CCVersion>
</CommandClass>
<CommandClass id="112" name="COMMAND_CLASS_CONFIGURATION">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="114" name="COMMAND_CLASS_MANUFACTURER_SPECIFIC">
		<CCVersion>2</CCVersion>
</CommandClass>
<CommandClass id="115" name="COMMAND_CLASS_POWERLEVEL">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="128" name="COMMAND_CLASS_BATTERY">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="129" name="COMMAND_CLASS_CLOCK">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="133" name="COMMAND_CLASS_ASSOCIATION">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="134" name="COMMAND_CLASS_VERSION">
		<CCVersion>1</CCVersion>
</CommandClass>
<CommandClass id="135" name="COMMAND_CLASS_INDICATOR">
		<CCVersion>1</CCVersion>
</CommandClass>
```

Fishwaldo:
Ok. Commiting this. I got a idea for flag day :)

Issue #2083, opened 2020-01-25
Shenzhen PD03Z confused by PD02Z.. and PD03Z missing - OZW 1.4


billiaz:
Hi All,

The Device I have on my hands, is missing parameters. Per manufacturer manual, the PD03Z is a newer version of the old device with added Temperature control.
It also has parameters moved around.

In the past I helped another database to add this device (Example of this device: https://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/973)

I would like to add the same details and correct the device recognition. Should I submit a pull request in this case?

Can I also alternatively try modify/test this on my local install? (Using now home assistant)

billiaz:
Some history on this: https://community.openhab.org/t/neo-coolcam-motion-sensor-pd02z-and-previous-models/60372/9

one of the devices changes these parameters on a firmware version

petergebruers:
It think it is not as easy as as you say.

I have investigated as similar case, or possibly the same case a while ago (July 2019)

Can you please have a look at this PR

https://github.com/OpenZWave/open-zwave/pull/1867

And please also read this:

https://github.com/OpenZWave/open-zwave/issues/1866

The main conclusion at that time was:

```
I suddenly realize all posts up to now do not properly explain the main issue and proposed fix.

So with same IDs there is a temperature model and a non-temperature model.

The device flashes a bright LED when it detects motion, and this can be disabled but they used a different parameter for this.

With the current nas-pd02z.xml it is not possible to turn off the LED if you have the newer "temperature" model.

I propose a new config file with these parameters

index="10" label="Ambient Temperature Differential Level Report or Led Blink Enable (0 or 1 - non-temperature version)"
index="11" label="Led Blink Enable (temperature reporting version)"
Parameter 10 is the only "conflicting" parameter but imho that explanation should help. Parameters 11,12 and 99 are new
```

aretakisv:
I have had the same investigation with another z-wave module.
That module had the way to filter on device firmware version.

We tested multiple devices and found that specific device code 0003:108d have this pattern:
Anything from firmware 3.80 has Temperature and the three extra params, anything else not!

The device with 3.80 firmware and higher is also called PD03Z and not PD02Z.
I suspect (but is only thoughts) manufacturer forgot to change the 108d to 208d in early devices.

should I update my pull request with your suggestions, in order to cover both cases?

petergebruers:
[...]

PR #2084, opened 2020-01-25
Add Shenzhen Neo PD03Z device, replaces the PD02Z which is wrong

aretakisv:
PD03Z device is covering the IDs 0003:108d and 0003:208d for manufacturer 0258

Based on manual:
Moved Parameter 10 to 11 ( enable disable Led )
Added Parameter 10 which adjusts the temperature report threshold
Added 99 which is the sensor calibration for Luminance

### DIFF
--- a/config/manufacturer_specific.xml
+++ b/config/manufacturer_specific.xml
@@ -1797,7 +1797,9 @@
 		<Product type="0003" id="2087" name="Power plug 12A" config="shenzen_neo/nas-wr01z.xml"/>
 		<Product type="0003" id="0083" name="Battery Powered PIR Sensor" config="shenzen_neo/nas-pd01z.xml"/>
 		<Product type="0003" id="1083" name="Battery Powered PIR Sensor" config="shenzen_neo/nas-pd01z.xml"/>
-		<Product type="0003" id="108d" name="Battery Powered PIR Sensor V2" config="shenzen_neo/nas-pd02z.xml"/>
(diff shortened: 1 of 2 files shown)


petergebruers:
Thank you for contributing. 

Please read my comments and discuss further the issue here, solving the problem probably isn't as easy as you think:

https://github.com/OpenZWave/open-zwave/issues/2083

You'll understand I cannot merge this without proper investigation and conclusion...

aretakisv:
What I am saying is.
I am trying to add the new device on 1.4 version of openzwave and your changes. which do not exist.
This impacts HASS.

the new device list is not picked by the python_openzwave. and on my HASS i still see only 10 params even with PD02Z

petergebruers:
A few things...

It does not look like my patch, that is not how I remember it, but I must admit I have not run a" diff", please make sure you do not break existing config of NAS PD02 owners!

OpenZWave 1.4 is no longer supported, @Fishwaldo will agree the way forward is OpenZWave 1.6, so we tend to spend not much energy on 1.4

And lastly, HASS does not use this repository, they have forked OpenZWave 1.4, so if you want it on HASS then you'll have to open a PR here:

https://github.com/home-assistant/open-zwave

If you can guarantee that this PR is sane and won't break stuff than I'll accept it though.

aretakisv:
Tganks for the update, I think I will close this one as makes no sense to do so.
I will take your changes and Device PD03Z in order to apply them to Home assistant. I will also submit an addition for PD03Z, for 1.6

petergebruers:
Thanks!

PR #2087, opened 2020-01-26
Update product ID for Fakro ZWS12 and ZWS230

pepeEL:
Update product ID for Fakro ZWS12 and ZWS230

Update by manual Fakro
https://www.fakro.pl/att/COMMON/offer/servis/user_manuals/PL_ZWS12_FAKRO.pdf


Fishwaldo:
are you sure that the ID's are incorrect? Several vendors use different ID's for different regions.

also as mentioned, Please run xmltest and submit all your changes in a Single PR.

pepeEL:
I have model ZWS230 and it was found as ZWS12. My Product ID is 0001 and type is 0003
But in openzwave it was add to type ZWS12 and it is problem that identify for me as ZWS12 not as ZWS230.
In manula fakro look ID and type for model ZWS12 is other that i decied to change it.

PR #2095, opened 2020-01-31
Added Honeywell 39357 / ZW3004 toggle dimmer

individualist:
As far as I can tell this device is identical to the GE toggle dimmer which is already in the DB, but with Honeywell branding and the Honeywell manufacturer id. Both are made by Jasco.


individualist:
Hmm. Did I do something wrong? When I try to use this locally I have to bump the revision # of the device file up to 6 or higher or the OZW log warns me that the revision is out of date, renames the file to 39357-ZW3004.xml.1, fails to download a replacement (because it's not in the db)  and then doesn't use it - even if I delete ozwcache_*.xml first...

petergebruers:
Let me have a look. I can already give one advice... it is easier to work on a another branch, not "master", before you open a PR. It is no big deal but I'll probably "squash merge" your edits and then your copy of master will diverge from OZW master and git will complain. No big deal though.

petergebruers:
Thanks! Your commits have been squashed, you'll have to check if you still have local unsaved work on your master branch and if you have none then do "git fetch origin" and "git reset --hard origin/master # Destroys your local work"

individualist:
No problem. Thanks!

PR #2100, opened 2020-02-05

rubimart:
I added the same 700342 10-Year Smoke Detector to the manufacturer_specifix.xml with different id.


Fishwaldo:
Thanks

Issue #1866, opened 2019-07-04
Neo Coolcam PIR Motion 


poudenes:
Hi All,

I found out that the config for Neo Coolcam PIR Motion is not correct
`open-zwave/config/shenzen_neo/nas-pd02z.xml`

With a other user (he created the new XML) we have all the options for those motion sensors.

We changed the line in manufacturer_specific.xml

`<Product config="shenzen_neo/nas-pd02z.xml" id="108d" name="Battery Powered PIR Sensor V2" type="0003"/>`

to 

`<Product config="shenzen_neo/nas-pd02ze.xml" id="108d" name="Battery Powered PIR Sensor V2" type="0003"/>`

Added the new nas-pd02ze.xml in shenzen_neo/

```
<Product sourceFile="shenzen_neo/nas-pd02ze.xml">
  <CommandClass id="112">
    <Value type="byte" genre="config" instance="1" index="1" label="Sensitivity Level Setting" value="12" min="8" max="255" size="1">
       <Help>This parameter defines the sensitivity of PIR detector, it is recommended to test the detector with movements from a farthest end of the coverage area at first time of use.         If movements cannot be detected sensitively, simply adjust the sensitivity level with this parameter.         This Parameter can be configured with the value 8 trough 255, where 8 means high sensitivity and 255 means lowest sensitivity.       </Help>
    </Value>
    <Value type="short" genre="config" instance="1" index="2" label="On/Off Duration" units="second" value="30" min="5" max="600" size="2">
       <Help>This parameter can be determined how long the associated devices should stay ON status.         For instance, this parameter is set to 30(second), the PIR detector will send a BASIC_SET Command to an associated device with value basic set level if PIR detector is triggered and the associated device will be turned on 30(second) before it is turned off.         This Parameter value must be large than Parameter 6#.         If user set this parameter to default by Configure CC, the parameter #6 will be set to default value.         Available Settings:5 to 600(second).       </Help>
    </Value>
    <Value type="byte" genre="config" instance="1" index="3" label="Basic Set Level" value="255" min="0" max="255" size="1">
       <Help>Basic Set Command will be sent where contains a value when PIR detector is triggered, the receiver will take it for consideration; for instance, if a lamp module is received the Basic Set Command of which value is decisive as to how bright of dim level of lamp module shall be.         This Parameter is used to some associated devices.         Available Settings: 0, 1 to 99 or 255.       </Help>
    </Value>
[...]

PR #2105, opened 2020-02-08
Added Haozee PIR Motion Sensor (Z52NAS-PD02Z)

individualist:


petergebruers:
Thank you for contributing. Looks live you've done everything by the book...

While I could accept your PR, I would prefer you to check something and try something else... It is a long sotry.

IMHO the device configuration already exists, what you do is add a regional variant... What I mean to say is this...

You add:

`<Product config="shenzen_neo/z52nas-pd02z.xml" id="008d" name="Z52NAS-PD02Z Haozee Zwave PIR Motion Sensor" type="0003"/>`

The 00 in "008d" is probably "region" and the 8d is "device model".

We already have:

`<Product config="shenzen_neo/nas-pd02z.xml" id="108d" name="NAS-PD02Z Battery Powered PIR Sensor V2" type="0003"/>`

So could you try your device with this line in manufacturer-specific?

<Product config="shenzen_neo/nas-pd02z.xml" id="008d" name="NAS-PD02Z Battery Powered PIR Sensor V2" type="0003"/>

If that works, I would prefer to have one config file because the NAS-PD02Z is known to exist in 2 variants, one with and one without temperature sensor. It depends on their firmware version.

To cater for both variants, the parameter descriptions are specially crafted like this:

index="10" label="Ambient Temperature Differential Level Report or Led Blink Enable (0 or 1 - non-temperature version)"

index="11" label="Led Blink Enable (temperature reporting version)"

index="12" label="Motion Event Report One Time Enable (temperature reporting version)"

If you bought this device recently, then you'll have the "temperature variant" (iirc firmware >= 3.80). If it is older, it might not have temperature but still allow you to enable/disable the red LED - but using the other parameter.

I know... It is confusing... Neo should have assigned a different model number and ID to the temperature version... But they didn't... This config file was tested and discussed, there is a closed issue with all the details.

petergebruers:
Ah, found the previous discussion of NAS-PD02Z

Neo Coolcam PIR Motion #1866

individualist:
I read that discussion before I created a new config, I did not think it applied due to the Haozee branded model I have actually being identifiable by **different** id.

And I can already tell you that the existing nas-pd02z.xml *will* work. Other than the text, it's almost functionally identical to my file. I've got not problem with reusing the xml file and current description. If that's the way you'd prefer to go.

I do have one question:

* In nas-pd02z.xml the sensitivity setting is defined as a list of 4 values, 8,12 (the default), 128, and 255.
* In the documentation for the non-temperature version for NAS-PD02z, it claims this setting takes 1-4. 
* In the documentation for the temperature version of the NAS-PD02z and for this "Haozee" branded version of the device it is a range of 8-255 (with a default of 12) where I can choose **any** value in that range. 

Is the definition as a list a workaround to allow support for both versions of this device in a single file?

individualist:
Commit amended with your suggested solution.

petergebruers:
Thank you for testing that. You are right about the IDs being different, but we cannot be sure if this "new" ID isn't in fact the same device but only with different frequency, and unless people post details about what they have go, we cannot know if they own a temp or a non-temp version.

If they ever make a "03Z" I hope they will change that "8d" in the ID...

About the "sensitivity setting" - that parameter was not added by me, but I'll look into that. There are quite a few different versions of the manual, someone told me the small leaflet he got with the device did not match the settings on his device, so it is a real mess... I mean the settings, the device is OK, I have a few of them and they perform as expected.

Fishwaldo:
@petergebruers go or no go?

petergebruers:
Sorry for the delay. Thank you for contributing, @individualist!

magpern:
What is the status of this PIR device?
Both PD02T and PD02Z reference the same xml file in config, but the config does not match the 
[documentation](https://fccid.io/Z52NAS-PD02Z/User-Manual/User-manual-2887191) for PD02Z.

In the documentation is says it has two association groups, xml says four.
Sensitivity settings in xml refer to value between 0 and 255, in documentation it is 1-4

I cannot get the associations to work
I cannot alter any settings in the device. They all reset to default after a refresh.

Anyone successfully used this PIR?

PR #2114, opened 2020-02-16
config: Auto-reformat whitespaces

andras-tim:
I tried to do things with config XMLs, but there are lots of left whitespaces.


andras-tim:
Can anyone help what should I do to pass the test? Where can I find the required changelog?

Fishwaldo:
Hi,
if you run ```make xmltest``` you will see all the errors - Basically any config file that is changed would need a new ChangeLog Entry updated in the MetaData Section. 

But - This is a loooot  of churn to existing config files, but IMHO, a better approach would be updating (or creating a new version in python) cpp/build/testconfig.pl to make sure future submissions are also "tidy" 

Accepting this change - We might have to discuss further. I've I merge this, then 10's of thousands of "downloads" against my server for updated config files that only have whitespace changes... I think we need to figure out a different approach.

andras-tim:
I tried to backport several config files to 1.4 (for Home Assistant...) but the files contain several whitespace issues that I didn't want to avoid.
I populated this idea to a bit bigger, but I feel now I parented a monster... and the main reason why I will drop the whole thing is the missing tool for avoiding this kind of issue in the future.

Issue #2134, opened 2020-02-27
First day here, help with new device 39351-ZW3010


crxporter:
Hello I'm new and getting my zwave network set up with zwave2mqtt. I've added a switch and I believe I have a new version of the Honeywell 39351 dimmer. I ordered from Amazon and was sent a version that isn't listed or pictured anywhere that I can find. The box I received says "simple wire" and "quick fit" like the GE 46203 dimmers.

I'd like to put together a pull request for the new dimmer but I'm not 100% sure all of the files I need to put together.

The device is in the device database [here](http://www.openzwave.com/device-database/0039:3235:4944). It seems to be a honeywell version of the [GE 46203](https://github.com/OpenZWave/open-zwave/blob/master/config/ge/46203-dimmer.xml).

I believe the new line for `manufacturer_specific.xml` will be:
```xml
    <Product config="honeywell/39351-ZW3010.xml" id="3235" name="39351 / ZW3010 In-Wall Smart Dimmer" type="4944"/>
```

And I think I need to make a combination of [the ge version](https://github.com/OpenZWave/open-zwave/blob/master/config/ge/46203-dimmer.xml) and [the previous honeywell version](https://github.com/OpenZWave/open-zwave/blob/master/config/honeywell/39351-ZW3005.xml).

I've found the dimmer [here](https://products.z-wavealliance.org/products/3600?selectedFrequencyId=2) on the z-wave alliance page. 

Based on the wiki page for adding a device, I think I just need to put together the config file - but I'm not sure what should come from the GE 46203 config and what should come from the Honeywell 39351 config.

Thank you!

Fishwaldo:
Thanks!

PR #2135, opened 2020-02-28
Add honeywell 39351-ZW3010

crxporter:
I think I have this right. Please review. 

See comments at #2134


crxporter:
Note I decided to use the GE 46203 as a starting point for the `39351-ZW3010.xml` file since these two devices should be the same thing with different brand markings (both made by Jasco).

I updated descriptions, part numbers, etc. to reflect Honeywell and the product description on z-wavealliance.org as instructed in the "adding devices" wiki page. But of course I'm still very new to this project so please let me know if I need to make corrections.

I actually bought some GE 46203 from Jasco and some 39351 and apart from what is printed on the switch, they look exactly the same as each other.

Fishwaldo:
Thanks!

PR #2136, opened 2020-03-02
Add support for Popp device

Janco59:
Add support for the Popp Mold Detector 701202.

I have virtually no experience with GitHub so please forgive me if I'm doing something wrong.
I wrote the configuration file based on the specs from the Open Z-Wave Alliance and the manufacturer. I tried most of the settings in a real setup (Domoticz) and am able to change them.
However, it would be a good idea if some-one could look at the config files.


Fishwaldo:
Thanks!

PR #2151, opened 2020-03-07
Fix maybe-uninitialized warning

travishaagen:
Fix maybe-uninitialized warning:

```
#5 80.71 Building src/command_classes/Configuration.cpp
#5 82.11 /src/open-zwave/cpp/src/command_classes/Configuration.cpp: In member function 'uint32 OpenZWave::Internal::CC::Configuration::getField(const uint8*, OpenZWave::Internal::CC::Configuration::CC_Param_Size, uint8&)':
#5 82.11 /src/open-zwave/cpp/src/command_classes/Configuration.cpp:84:12: error: 'value' may be used uninitialized in this function [-Werror=maybe-uninitialized]
#5 82.11    84 |     return value;
#5 82.11       |            ^~~~~
#5 82.88 cc1plus: all warnings being treated as errors
#5 82.90 make[1]: *** [/src/open-zwave/cpp/build/support.mk:192: /src/open-zwave/.lib/Configuration.o] Error 1
#5 82.90 make[1]: Leaving directory '/src/open-zwave/cpp/build'
#5 82.90 make: *** [Makefile:23: all] Error 2

```

### DIFF
--- a/cpp/src/command_classes/Configuration.cpp
+++ b/cpp/src/command_classes/Configuration.cpp
@@ -66,7 +66,7 @@ namespace OpenZWave
 			};
 
 			uint32 Configuration::getField(const uint8 *data, CC_Param_Size size, uint8 &pos) {
-				uint32 value;
+				uint32 value = 0;
 				switch (size) {
 					case CC_Param_Size::CC_Param_Size_Byte:
 						value = data[pos++];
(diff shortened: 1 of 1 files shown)


Fishwaldo:
I’d prefer to initialize value at the declaration instead of a default case statement. 

Default case statement will mean we won’t get a warning if we add additional ENUM’s later and don’t handle them in switch statements. I’ll push that fix soon.

travishaagen:
I updated the PR to initialize the value at declaration, and to log an error on the `default` case.

However, if you want to address this yourself, then no problem.

Fishwaldo:
Thanks. I'll still drop the default, as I would prefer a compile warning rather than a log message.

PR #2157, opened 2020-03-14
Adds North American ID information for Aeotec ZW164-A

mpbelhorn:
Adds identifier information for the North American frequency region version of the Aeotec Siren 6 (ZW164-A) to the existing device configuration metadata.

This information was fetched from the [Z-Wave Alliance data for the Aeotec Siren 6 ZW164-A](https://products.z-wavealliance.org/products/3284).


Fishwaldo:
Thanks!

PR #2164, opened 2020-03-23
Add Honeywell 39348-ZW4005 In-Wall Smart Switch

dbajric:


Fishwaldo:
OZW 1.4 is unmaintained now.

Issue #2129, opened 2020-02-20
"IsNodeZWavePlus" flag is set to false for Z-Wave Plus controller node


kpine:
The IsNodeZWavePlus flag for the primary controller node is always false. Is it possible to set it to true for controllers that are actually z-wave plus?

In Home Assistant, the flag is exported as an attribute for the zwave entity. A common support question is something along the lines of "Why is my controller reporting that it doesn't support Z-Wave Plus, do I need to enable something"?

If it were possible to have the flag set correctly, it would solve a lot of confusion.

Fishwaldo:
Technically - OZW is correct, but due to OZW not exposing the Z-Wave+ CommandClass on the Controller Node. 
(ZW+ is both a hardware and a software thing). 

The other challenge is the controller behaves differently from other devices on the network and we can not query 90% of the capabilities via the Serial Port (if you had another controller on the Network, you would be able to, for instance, turn off the disco lights etc).

I understand the confusion. I might look into "fudging" that flag for Controllers in the Future, if i can tell how to differentiate between ZW+ and ZW+v2)

kpine:
I noticed the MQTT data is showing some "NodePlusType" information:
```json
{
    "NodeManufacturerName": "Unknown: id=0000",
    "NodeProductName": "Unknown: type=0003, id=0008",
    "NodeBasicString": "Static Controller",
    "NodeBasic": 2,
    "NodeGenericString": "Static Controller",
    "NodeGeneric": 2,
    "NodeSpecificString": "Static PC Controller",
    "NodeSpecific": 1,
    "NodeManufacturerID": "0x0000",
    "NodeProductType": "0x0003",
    "NodeProductID": "0x0008",
    "NodeBaudRate": 100000,
    "NodeVersion": 4,
    "NodeGroups": 0,
    "NodeDeviceTypeString": "Unknown Type (0x0000)",
    "NodeDeviceType": 0,
    "NodeRole": 0,
    "NodeRoleString": "Central Controller",
    "NodePlusType": 0,
    "NodePlusTypeString": "Z-Wave+ node",
}
```
Could `NodePlusType` value be used to "fudge" the flag for a controller? I only have 2 non-zwave-plus devices and they're part of my OZW 1.4 network, but I did notice that in the zwcfg*.xml file those nodes are missing the `nodetype="0"` attribute, while all the plus nodes have it.

Or, maybe the application should not even expose this flag for primary controllers.

[...]

PR #2168, opened 2020-03-24
Rename Serial Version to Firmware Version and add Communication and Chip Version

petergebruers:
1 - Motivation for the rename logging of "Serial API Version" to "Firmware Version":

a) "Serial API Version" is confusing, the serial IP does indeed have a "version" but that is not this one.

b) The spec calls it an "application version"... And that is what end-users commonly know as "firmware version"

INS12350-17 Serial API Host Appl. Prg. Guide
7.3 Capabilities Command
FUNC_ID_SERIAL_API_GET_CAPABILITIES

ZW->HOST: RES | 0x07 | SERIAL_APPL_VERSION | SERIAL_APPL_REVISION | SERIALAPI_MANUFACTURER_ID1 | SERIALAPI_MANUFACTURER_ID2 | SERIALAPI_MANUFACTURER_PRODUCT_TYPE1 | SERIALAPI_MANUFACTURER_PRODUCT_TYPE2 | SERIALAPI_MANUFACTURER_PRODUCT_ID1 | SERIALAPI_MANUFACTURER_PRODUCT_ID2 | FUNCID_SUPPORTED_BITMASK[ ]

SERIAL_APPL_VERSION is the Serial API application Version number.
SERIAL_APPL_REVISION is the Serial API application Revision number.

Notice it is not "Serial API Version number" but "Serial API *application* Version number"

That's why I would log it as "Firmware Version"

The result on 3 of my dongles:

- A somewhat older Z-Wave.me UZB1 prints 5.6 which is correct (though sometimes printed as 5.06 because the dot is not a decimal separator)
- A recent UZB7 prints 7.0 which I did not check but sounds reasonable enough
- A Reference Implementation of a controller prints 6.8 which matches its definition

2 - Motivation for printing "Communication Version"

As discussed in issue https://github.com/OpenZWave/open-zwave/issues/2129 the "Serial Version" is a number between 1 and 8 and can inform us of the capabilities of the controller (documented in INS12350)

For example... My 3 dongles...

- Z-Wave.me UZB1 prints

```
Info, contrlr, Received reply to FUNC_ID_SERIAL_API_GET_INIT_DATA:
Info, contrlr,     Communication Interface Version: 6
```

This means that with this firmware, my controller supports "Extended Status Information" but does not do "Smartwtart

-  UZB7 prints "8" so supports S2 and "Smart Start"
- My reference controller also prints "8" which matches the capabilities of the SDK it was based on...

ATM I do not have a series 300 controller to test with.

BTW this does not answer the question "is this a ZW-Plus" controller but I think this information is useful regardless.

### DIFF
--- a/cpp/src/Driver.cpp
+++ b/cpp/src/Driver.cpp
@@ -114,7 +114,7 @@ static char const* c_sendQueueNames[] =
 //-----------------------------------------------------------------------------
 Driver::Driver(string const& _controllerPath, ControllerInterface const& _interface) :
 		m_driverThread(new Internal::Platform::Thread("driver")), m_dns(new Internal::DNSThread(this)), m_dnsThread(new Internal::Platform::Thread("dns")), m_initMutex(new Internal::Platform::Mutex()), m_exit(false), m_init(false), m_awakeNodesQueried(false), m_allNodesQueried(false), m_notifytransactions(false), m_timer(new Internal::TimerThread(this)), m_timerThread(new Internal::Platform::Thread("timer")), m_controllerInterfaceType(_interface), m_controllerPath(_controllerPath), m_controller(
-				NULL), m_homeId(0), m_libraryVersion(""), m_libraryTypeName(""), m_libraryType(0), m_manufacturerId(0), m_productType(0), m_productId(0), m_rfregion(ZW_RFRegion::RF_REGION_UNKNOWN), m_initVersion(0), m_initCaps(0), m_controllerCaps(0), m_Controller_nodeId(0), m_nodeMutex(new Internal::Platform::Mutex()), m_controllerReplication( NULL), m_transmitOptions( TRANSMIT_OPTION_ACK | TRANSMIT_OPTION_AUTO_ROUTE | TRANSMIT_OPTION_EXPLORE), m_waitingForAck(false), m_expectedCallbackId(0), m_expectedReply(0), m_expectedCommandClassId(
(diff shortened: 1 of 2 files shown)


petergebruers:
I wanted to add the "Chip version" as a separate PR but think it makes sense to add the chip version right away... Give me a second...

petergebruers:
So my UZB1 and my "reference controller" are both ZW500 chips and it comes as no surprise that OZW logs:

```
Chip Type: 5
Chip Version: 0
```

My UZB7 says

```
Chip Type: 7
Chip Version: 0
```

Again, imho useful information but still does not 100% answer "is this a ZW Plus controller" because it is a combination of SDK version and hardware and even certification but I think it is pretty safe to assume >= 5 means Z-Wave Plus...

Table in INS12350-17

Controller Chip Types

petergebruers:
Results of a Fibaro HC2 controller... This is a ZW300 chip... There is something odd about this controller, it might be caused by "migrating" all my stuff to a HC3. It reports a "0" HomeID and also Silabs PC Controller won't fully initialise it. But I can fake homeID = "1" in code and make OZW continue.

```
contrlr,  Received reply to FUNC_ID_SERIAL_API_GET_CAPABILITIES
contrlr,     Firmware Version:     3.94
contrlr,     Manufacturer ID:      0x010f
```

That's correct.

```
contrlr, Received reply to FUNC_ID_SERIAL_API_GET_INIT_DATA
contrlr,     Communication Interface Version: 5
contrlr,     Chip Type: 3
contrlr,     Chip Version: 1
```

That is correct again, it is a ZW0301 chip and definitely not ZW Plus. Serial Communication Version 5 is something I cannot really verify, it might be that this was highest possible version they could get with ZW300 compatible SDK... I don't know.

Fishwaldo:
Thanks!

PR #2176, opened 2020-04-02
Add missing option for ZW162 (Aeotec Doorbell 6)

LordMike:
Add the parameter 0x60 (96), a boolean to specify if the action button stops the currently playing chime.

Also update the link to the Doorbell 6 technical docs.

Relevant part of the docs. It specifies the read/write option no. 96, with a default of "0", size of "1".

image


Fishwaldo:
Thanks!

PR #2178, opened 2020-04-06
Termux / Android compatibility 

araczkowski:
Android, and by extension, Termux, is missing pthread_cancel and pthread_setcancelstate.
A workaround is in this pull request - with this we are able to compile and run open-zwave on Android.

We are going to create termux-package with open-zwave to automatically build deb for termux.

Would you consider to marge this pull request, then we will be able to switch the termux-package to the official OpanZwave repo without patching?


Fishwaldo:
LGTM :)

PR #2191, opened 2020-04-24
Add additional identifier for the Fibaro FGT-001 Heat Controller

smokku:
There is a new revision of Fibaro FGT-001 Heat Controller with new id `0001`.


smokku:
```
##[error]/home/runner/work/open-zwave/open-zwave/cpp/build/support.mk:69: git describe returned an empty result, setting GITVERSION to VERSION_MAJ.VERSION_MIN.-1 and VERSION_REV to 0
```
I think this is nothing I can fix in my PR...

Fishwaldo:
The error is actually:
 @github-actions github-actions / ConfigTest

config/manufacturer_specific.xml#L1

But I'll fix it post merge. 
Revision Number Was Not Bumped - Error Code 8

PR #2196, opened 2020-04-26
Door Condition per 1.6 specs, adds DoorSense support

quinnhosler:
Implements Door Condition per 1.6 specs under COMMAND_CLASS_DOOR_LOCK

Additionally includes updated xml files for Door Condition and includes August DoorSense™ support


config/august/asl-03.xml
```diff
@@ -33,8 +33,21 @@ time.</MetaDataItem>
...
+    <Value genre="config" type="list" index="7" instance="1" label="DoorSense" min="0" max="3" size="1" value="2">
+      <Help>
+       State of Door and Bolt
+      </Help>
+      <Item label="Locked/Open" value="0" />
+      <Item label="Locked/Closed" value="1" />
+      <Item label="Unlocked/Open" value="2" />
+      <Item label="Unlocked/Closed" value="3" />
+    </Value>
+  </CommandClass>
```
Fishwaldo, line 50: This is not a Configuration CC id - Why is it required?

cpp/src/ValueIDIndexesDefines.def
```diff
@@ -2156,7 +2156,8 @@ ENUM(ValueID_Index_DoorLock,
 		System_Config_Minutes = 3,
 		System_Config_Seconds = 4,
 		System_Config_OutsideHandles = 5,
-		System_Config_InsideHandles = 6
+		System_Config_InsideHandles = 6,
+		Door_Condition = 7
```
Fishwaldo, line 2160: Please follow the naming conventions

cpp/src/command_classes/DoorLock.cpp
```diff
@@ -85,6 +85,18 @@ namespace OpenZWave
...
+				DoorLockCondition_Unlatched_Locked_Open = 0x00,
+				DoorLockCondition_Unlatched_Locked_Closed = 0x01,
+				DoorLockCondition_Unlatched_Unlocked_Open = 0x02,
+				DoorLockCondition_Unlatched_Unlocked_Closed = 0x03,
+				DoorLockCondition_Latched_Locked_Open = 0x04,
+				DoorLockCondition_Latched_Locked_Closed = 0x05,
+				DoorLockCondition_Latched_Unlocked_Open = 0x06,
+				DoorLockCondition_Latched_Unlocked_Closed = 0x07,
+			};
+
```
Fishwaldo, line 99: Why is this required?

cpp/src/command_classes/DoorLock.cpp
```diff
@@ -194,6 +206,11 @@ namespace OpenZWave
 						value->OnValueRefreshed(lockState);
 						value->Release();
 					}
+					if (Internal::VC::ValueList* value = static_cast<Internal::VC::ValueList*>(GetValue(_instance, ValueID_Index_DoorLock::Door_Condition)))
+					{
+						value->OnValueRefreshed(_data[3]);
+						value->Release();
+					}
```
Fishwaldo, line 213: Few issues here:
1) this ValueID is not created by the Class. You need to determine if the device in question supports this ValueID (probably based upon the CommandClass Version) and then create the ValueID, and subsequently, in this specific block, only do this update if the CommandClass Version is equal/above the minimum version and you should also check if index 3 is valid in the data recieved.

config/august/asl-03.xml
quinnhosler: You're right this isn't needed and I didn't understand why I was making that change. I removed the changes to this file.

cpp/src/command_classes/DoorLock.cpp
quinnhosler: I was looking through the ZW docs and it appears DoorCondition is defined as far back as version 1 of the protocol and hasn't changed sense. I'm assuming this negates the need for any version checking? I added a simple validation of index 3.

quinnhosler:
Hey @Fishwaldo, thanks for your comments. I went ahead and made changes addressing them. Let me know if there are any other changes required!

quinnhosler:
What is the current thought process with this pull request? Does it need more work?

perosb:
I assume the conflicts that now occured need to be resolved first.

I'd be very interested in DoorCondition as well to test on ID Lock.
Good stuff!

elgeniskogen:
Would this fix the problem with ID-Lock 150 where I cannot see if door is closed or open? That would be fantastic. Now a I have to use an additional  sensor  as I am not able use the one in the lock. I have a lock to test on if needed. Your contributions are highly appreciated :-)

perosb:
> Would this fix the problem with ID-Lock 150 where I cannot see if door is closed or open? That would be fantastic. Now a I have to use an additional sensor as I am not able use the one in the lock. I have a lock to test on if needed. Your contributions are highly appreciated :-)

That is my understanding. I also have the ID Lock 150 with an additional sensor to solve this.

labaland:
When will this be added?  =)

drgreenbaum:
Please merge this! Would be a great addition!

labaland:
> Please merge this! Would be a great addition!

AGREE!!

drgreenbaum:
@Fishwaldo what is needed to make this happen? Do we need testing?

johnniemalm:
I have this to,

PR #2207, opened 2020-05-04
Adds config for devolo Home Control Shutter FM MT2761

ProtoFoo:
The Roller Shutter MT2761 is basically identical to the Qubino ZMNHCD1 Shutter (which is already supported). Just the firmware might be different. Short testing shows that the MT2761 works fine in my z-wave network and the relays respond to "up" and "down" commands.

Note: After runnning `make dist-update` as described in the guidelines, some other files were added to `testconfigversions.cfg`. I suppose they are from a previous commit?


Fishwaldo:
Thanks!

PR #2224, opened 2020-05-19
Adds config for devolo Home Control Switch FM MT2759

nicoh88:
The Switch FM MT2759 is basically identical to the Qubino ZMNHAD1 Flush (which is already supported). Just the firmware might be little different.

Note: After runnning make dist-update as described in the guidelines, some other files were added to testconfigversions.cfg. I suppose they are from a previous commit?


Fishwaldo:
Thanks!

PR #2226, opened 2020-05-19
Add support for different MCO Home MH9 and extend config options

arjenhiemstra:
Add support for MCO Home MH9 multisensor with different type and id than currently in de DB and improve config file to include all config options.


Fishwaldo:
Thanks - Next time please try to fix up the checks that do not pass

Issue #2205, opened 2020-05-01
OpenZWave reinterviewing all devices at every restart


akamming:
on request of @gizmocuz i  post this issue here...

for some unknown reason for some users at domoticz OpenZWave is re-interviewing all devices at every restart, even if the information is in the ozwcache.xml file (as far as i can see, but my knowledge does not go very far). This is an issue, cause at every reboot you have to wake up all battery powered devices manually before they work again. 

additional information can be found in the [issue on domoticz](https://github.com/domoticz/domoticz/issues/4119) and [the forum thread](https://www.domoticz.com/forum/viewtopic.php?f=24&t=32401&sid=711107c8048a9eff304fc5dac678210a)

Let me know what additional info you need for analysis, then i can try to gather it...

akamming:
Additional info. I just found out that
- After a power cycle (taking my pi of the power): all devices are loaded from ozwcache
- after a restart of the application without power cycle: 3 devices are loaded from cache. domoticz (with openzave 1.6) starts re-interviewing all other devices.

Fishwaldo:
Sleeping devices need to be woken up after restart to get the initial data that is needed to validate the cache is valid. They will not "operate" correctly till they are woken up (and proceed past QueryStage_CacheLoad.

How are you determining that devices are being re-interviewed. For any device that has proceeded past QueryStage_CacheLoad, we still get the "dynamic" values from the device - That is the latest value (eg, temp, status of a switch) after restarting.

Fishwaldo:
By the way - Looking at the domoticz issue - If you are experiencing a crash - Then it might be at a point where caching is not completed. I can't diagnose that issue.

akamming:
> How are you determining that devices are being re-interviewed. For any device that has proceeded past QueryStage_CacheLoad, we still get the "dynamic" values from the device - That is the latest value (eg, temp, status of a switch) after restarting.

i see this activity happening in the control panel.  most devices start with status "probe" and get several statuses before they are "ready"  again. The battery powered devices start with "probe (sleeping). After i manually wake them up them up, they go through several statuses till they are " sleeping"  again

> By the way - Looking at the domoticz issue - If you are experiencing a crash - Then it might be at a point where caching is not completed. I can't diagnose that issue.

I already did some more testing on this. The crash is unrelated. I found out that
1 If i completely power down my system and then restart domoticz. All my devices are read from the cache and everything works as it should
[...]

PR #2227, opened 2020-05-20
Fix issue for truncating cache file at closedown of application

akamming:
patch for issue #2205 

issue is confirmed by several users 

fix has been succesfully tested on rpi4b with domoticz


Fishwaldo:
Thanks - But I put in a alternative solution for calling WriteCache during shutdown.

PR #2230, opened 2020-05-24
Adds config for devolo Home Control Dimmer FM MT2760

nicoh88:
The Dimmer FM MT2760 is basically identical to the Qubino ZMNHDD1 Dimmer (which is already supported). Just the firmware might be little different.


Fishwaldo:
Thanks!

PR #2231, opened 2020-05-24
Add config for Honeywell ZW4106 Plug-in Smart Switch

alexander-bauer:
Hi all, this is my first contribution to Open Z-Wave, intends to add support for the [Honeywell ZW4106 Plug-in Smart Switch](https://products.z-wavealliance.org/products/3740?selectedFrequencyId=2), which I own one of.

My development environment isn't up to snuff, and I've had trouble with Perl dependencies for running the `xmllint`, and a too-old Git version for the `makedist`. I think this PR will require an additional commit by someone with a happier development environment, or some updates on my part.

Recommendations welcome; thank you all for the great work you do!


alexander-bauer:
This gave me the kick I needed to get Git upgraded on my dev VM, so I was able to run the `xmltest`. I had to bump the revision number in `manufacturer_specific.xml`, but it appears that the GitHub tests pass now. I'm still not able to run `make dist-upgrade` because one of my Perl libraries is out of date.

Fishwaldo:
Thanks!

PR #2232, opened 2020-05-28
Updates for Zooz Zen27 and ZSE29

bepsoccer:
- Zen27 update to reflect the min value of 0 for parameter 9
- ZSE29 config parameters added and are valid for firmware 2.0


Fishwaldo:
Thanks

PR #2238, opened 2020-06-06
Create XML for ZEN24 v3 hardware

cybergrimes:
Zooz ZEN24 has a v3 hardware revision that has many more features than the basic v2 functions.
Change log with differences: https://www.support.getzooz.com/kb/article/298-zen24-toggle-dimmer-change-log/
Settings derived from manufacturer page: https://www.support.getzooz.com/kb/article/318-zen24-toggle-dimmer-ver-3-08-advanced-settings/

I will submit for a new entry in manufacturer_specific.xml,
<Manufacturer id="27a" name="Zooz">
<Product type="b112" id="261c" name="ZEN24 Dimmer Switch" />

PR #2239, opened 2020-06-06
Add ZEN24 Dimmer Switch v3

cybergrimes:
Please see: https://github.com/OpenZWave/open-zwave/pull/2238
Zooz ZEN24 v3 hardware offers advance settings not available in original hardware. Created zen24v3 XML for these config settings.


Fishwaldo:
OZW 1.4 is no longer maintained.

PR #2251, opened 2020-06-12
Zooz ZEN21v3 corrections

cybergrimes:
Config XML for v3 hardware incorrectly labeled v2, manufacturer specific entry was never pointed to config XML after created. Everything was pointing to the original v1/v2 XML.


cybergrimes:
Not sure if I should be worried about the revision number for manufacturer_specific.xml but I updated it. I see in history its sometimes bumped by project owners.

PR #2256, opened 2020-06-14
fixed kwikset 914c alarm message index for refresh

steve28:
Changed the index value of the alarm message that should trigger a refresh.


Fishwaldo:
THanks!

PR #2261, opened 2020-06-18
Update ZEN21v3 parameter 12 description

cybergrimes:
Update description for parameter 12, this switch is not a dimmer and incorrectly references a value not applicable to this switch


Fishwaldo:
Thanks!

PR #2263, opened 2020-06-19
add Zooz ZEN23v3 config

cybergrimes:
Adding config for Zooz ZEN23 with v3+ hardware/firmware
https://www.support.getzooz.com/kb/article/301-zen23-toggle-switch-change-log/

This has the same node identify codes as v1/v2 versions (027a:b111:251c), how do we add this to manufacturer_specific.xml? If codes are pointed to this new config, older devices would get config options they are incapable of using. Conversely newer devices are not getting access to all features of the device, requiring additional config by users. How do you recommend this be handled here?


Fishwaldo:
Thanks!

cybergrimes:
> Thanks!

Did you have any guidance on how to include in manufacturer_specific? 
Please see original comment re: this device sharing node identity codes with previous versions

Fishwaldo:
Bitch loudly at the vendor. There is no way to distinguish if they use the same id’s.

cybergrimes:
Hah, yeah. Unfortunately it sounds like other projects/products are using the firmware version as an additional identifier.

PR #2268, opened 2020-06-23
Create DZPA1.xml

kaijk:
The Leviton DZ15S in-wall switch config was previously changed almost completely to the DZPA1 plug-in device, yet retained the DZ15S product type and zwave alliance "1957" ID so the DZ15S config didn't match either the DZ15S or the DZPA1 device. 

This PR copies the DZ15S config to a new entry for DZPA1, changing the metadata that continued referencing the DZ15S to reference the equivalent DZPA1 information.  Note, I do not have a DZPA1 to test, but, I think this PR reflects the goal of modifying the DZ15S metadata probably so as to clone it to a new config for the DZPA1.  

I will next work to back out or otherwise change the DZ15S so its metadata will match its definition.


config/leviton/DZPA1
```diff
@@ -0,0 +1,72 @@
+<!--
```
Fishwaldo, line 1: Shouldn't this be a DZPA1.xml file? (missed the extension)

config/leviton/DZPA1
```diff
@@ -0,0 +1,72 @@
...
+  </CommandClass>
+  <!-- Association Groups -->
+  <CommandClass id="133">
+    <Associations num_groups="1">
+      <Group index="1" label="Lifeline" max_associations="5"/>
+    </Associations>
+  </CommandClass>
+  <CommandClass id="134">
+    <Compatibility>
+      <ClassGetVersionSupported>false</ClassGetVersionSupported>
```
Fishwaldo, line 69: Why are we not getting a version? this indicates a "broken" device so just want to confirm

config/leviton/DZPA1
kaijk: Yes, changed to DZPA1.xml.

config/leviton/DZPA1
kaijk: This is a new device, finishing someone's copying of the DZ15S.xml to a new DZPA1.xml. Or, am I misunderstanding?

kaijk:
I'll need to start over. Sorry.  Noob

PR #2272, opened 2020-06-24
Add VerifyChange to Homeseer HS-WD100+ to resolve out-of-sync state

grizzlyjere:
OZW is not reporting the final dimming value of the HS-WD100+.  This change adds the `VerifyChanged` compatibility setting to resolve the issue.  Tested against HS-WD100+ firmware 5.19.


Fishwaldo:
Thanks!

PR #2273, opened 2020-06-24
Update for revised Zooz ZEN23 config

cybergrimes:
Use the ZEN23v3 config by default for advanced config settings. Vendor used same node identity codes as old hardware that can't be updated; changing to new config since this is what is still being sold today. As far as I can see having new config options on old hardware doesn't cause any problems, they are just ignored / don't do anything.


Fishwaldo:
Thanks!

PR #2274, opened 2020-06-24
Create Leviton DZPA1.xml and image, Revise DZ15S and Image

kaijk:
2nd try at adding the DZPA1.xml and its image.  
@Fishwaldo  you asked in my first attempt. "Why are we not getting a version? this indicates a "broken" device so just want to confirm".  I'm sorry but I don't understand. I set the `dzpa1.xml` to `revision 1`.  Is there another versioning to do?
I did increment the revision of the `manufacturer_specific.xml` file.


Fishwaldo:
I was talking about this:
```
  <CommandClass id="134">
    <Compatibility>
      <ClassGetVersionSupported>false</ClassGetVersionSupported>
    </Compatibility>
  </CommandClass>
```

kaijk:
Nine of the 12 Leviton device specifications in ozw have the "134"
`CommandClass` defined this way - as `false`.  The other three don't
include the 134 `CommandClass` at all. Also, no device from any other
manufacturer includes this `CommandClass` either based on a recursive
`grep` of the Class.

I can change the value to `true` for the DZPA1, but it will be the only
one.  Is that more appropriate?  I used the DZ15S as the DZPA1 template
since the metadata currently reflects the DZPA1 more than the DZ15S itself.

Fishwaldo:
You would need to test without that block in your config file and see if the device works as expected.

kaijk:
Removes CommandClass 134 from new dzpa1 and from revised dz15s following successful local testing following CommandClass removal

kaijk:
Tested locally w/o CommandClass 134 and revised the PR.
Thanks

Fishwaldo:
Thanks!

PR #2276, opened 2020-06-25
add FGRGBW-442

rubenfiszel:
not sure why there was no PR upstream. I do not know @brouwer but this seems to do the job.


brouwer:
Got tagged here, so just wanted to clarify that I don't think this is going to work. I switched away from openzwave, never managed to get the FGRGBW-442 to work with openzwave, as it uses different command classes and cc versions that didn't seem to work with openzwave.

rubenfiszel:
@brouwer thanks for clarifying, I will close the PR then. Seems odd that this device would be so radically incompatible with openzwave?

rubenfiszel:
In this thread, people are mentionning that the ZEN31 is basically the same device: https://github.com/home-assistant/core/issues/31253

Maybe it would be enough to copy the definition and adjust the metadata here?

PR #2260, opened 2020-06-17
Fix YRD446 alarm message index for refresh

DelusionalAI:
See https://github.com/OpenZWave/qt-openzwave/issues/105, and https://github.com/OpenZWave/open-zwave/pull/2256

Proper refresh lock status when lock is manually operated.

PR #2277, opened 2020-06-26
Fix YRD216 alarm message index for value refresh

mpbelhorn:
Manual lock and unlock operations for the Assa Abloy YRD216 Touch Button Deadbolt do not automatically refresh the values for the `COMMAND_CLASS_DOOR_LOCK` status values:

CommandClass 98 Instance 1 Index 0 "Locked" Boolean
and
CommandClass 98 Instance 1 Index 1 "Locked (Advanced)" list

This commit adds a value refresh trigger on the `COMMAND_CLASS_NOTIFICATION` Instance 1 Index 6 "Access Control" value which changes on manual, keypad, and wireless lock/unlock operations to update the overall lock status.

This change is similar in nature to #2260, #2256 and https://github.com/OpenZWave/qt-openzwave/issues/105.

### DIFF
--- a/config/assa_abloy/PushButtonDeadbolt.xml
+++ b/config/assa_abloy/PushButtonDeadbolt.xml
@@ -1,4 +1,4 @@
-<Product Revision="6" xmlns="https://github.com/OpenZWave/open-zwave">
+<Product Revision="7" xmlns="https://github.com/OpenZWave/open-zwave">
   <MetaData>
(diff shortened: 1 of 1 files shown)


mpbelhorn:
I've tested this on my own Z-Wave network and it's working great. This change now allows the lock status to be correctly updated regardless of how the lock was manipulated and regardless of the lock security state. Previously, lock status was only updated when using Z-Wave commands to manipulate the lock.

Fishwaldo:
Thanks!

PR #2283, opened 2020-06-28
Add support For Ge 12720

mike240se:
This is the older non-plus revision of the ZW4201 GE/Jasco Outdoor Plugin Smart Switch.  The product images are exactly the same, the casing of the product was the same for both revisions.

The manual is not available on Zwave alliance DB, I added another one I found though.


Fishwaldo:
Thanks!

PR #2289, opened 2020-07-03
[config] Add Intermatic PE653 and PE953 pool/spa controller and remote.

j9brown:
I'm not sure about the VerifyChanged compatibility declarations.  The ozw config file only seems to pick up instance 1 and doesn't actually set the verify changed flags.

This device requires polling and verify changes to work correctly.  Calling "enablePoll" and "setChangeVerified" at runtime for each value of interest does the trick.  What's the best way to encode that here?


Fishwaldo:
>This device requires polling and verify changes to work correctly

It shouldn't. Both are basically doing the same thing - polling continuously polls the device for changes. VerifyChanges will poll the device for changes after we get a update - if two reports are the same it publishes the value. 

I don't accept config files that enable polling as thats more of a Policy Decision than a config decision. 

VerifiedChanges can be enabled as per https://github.com/OpenZWave/open-zwave/wiki/CommandClass-Compatibility-Flags#all-commandclasses

config/intermatic/pe653.xml
```diff
@@ -0,0 +1,62 @@
...
+  </MetaData>
+
+  <!-- This device must be polled.  It also has a strange habit of reporting
+       old state when queried immediately after setting a switch and then
+       reporting the correct state when queried a few moments later. -->
+
+  <!-- This device reports 6 switches but there are only 5 circuits.
+       The switch on the default endpoint is redundant with endpoint 1 so
+       we omit it from the list of instances. -->
+  <CommandClass id="32">
```
Fishwaldo, line 23: This should be auto discovered. Its not needed.

config/intermatic/pe653.xml
```diff
@@ -0,0 +1,62 @@
...
+    <Instance index="1" label="Circuit 1" />
+    <Instance index="2" label="Circuit 2" />
+    <Instance index="3" label="Circuit 3" />
+    <Instance index="4" label="Circuit 4" />
+    <Instance index="5" label="Circuit 5" />
+  </CommandClass>
+  <CommandClass id="37">
```
Fishwaldo, line 30: This should be auto discovered. Its not needed.

config/intermatic/pe653.xml
```diff
@@ -0,0 +1,62 @@
...
+    <Instance index="1" label="Circuit 1" />
+    <Instance index="2" label="Circuit 2" />
+    <Instance index="3" label="Circuit 3" />
+    <Instance index="4" label="Circuit 4" />
+    <Instance index="5" label="Circuit 5" />
+    <Compatibility>
+      <VerifyChanged index="1">true</VerifyChanged>
```
Fishwaldo, line 37: index refers to the ValueID Index - Not instance Index. I don't think you need a entry for each "instance"

config/intermatic/pe653.xml
```diff
@@ -0,0 +1,62 @@
...
+      <VerifyChanged index="2">true</VerifyChanged>
+      <VerifyChanged index="3">true</VerifyChanged>
+      <VerifyChanged index="4">true</VerifyChanged>
+      <VerifyChanged index="5">true</VerifyChanged>
+    </Compatibility>
+  </CommandClass>
+
+  <CommandClass id="49">
```
Fishwaldo, line 45: This should be auto discovered. Its not needed.

config/intermatic/pe653.xml
```diff
@@ -0,0 +1,62 @@
...
+    <Instance index="1" label="Water Temperature" />
+    <Value type="decimal" genre="user" instance="1" index="1" label="Water Temperature" units="F">
+      <Help>Water Temperature Sensor Value</Help>
+    </Value>
+  </CommandClass>
+
+  <CommandClass id="67">
```
Fishwaldo, line 52: This should be auto discovered. Its not needed.

config/intermatic/pe953.xml
```diff
@@ -0,0 +1,15 @@
...
+    <MetaDataItem name="OzwInfoPage">http://www.openzwave.com/device-database/0005:0953:5045</MetaDataItem>
+    <MetaDataItem name="ProductPic">images/intermatic/pe953.png</MetaDataItem>
+    <MetaDataItem id="0953" name="ZWProductPage" type="5045">https://products.z-wavealliance.org/products/598/</MetaDataItem>
+    <MetaDataItem id="0953" name="FrequencyName" type="5045">U.S. / Canada / Mexico</MetaDataItem>
+    <MetaDataItem id="0953" name="Identifier" type="5045">PE953</MetaDataItem>
+    <ChangeLog>
+      <Entry author="Jeff Brown - jeff.brown@gmail.com" date="29 June 2020" revision="1">Initial creation</Entry>
+    </ChangeLog>
+  </MetaData>
+</Product>
```
Fishwaldo, line 15: No Config Params or Association Groups?

config/intermatic/pe653.xml
j9brown: Per the comment, this is intentional.  When auto-discovered, the device presents 6 switch instances however physically there are only 5 of them.  By listing them explicitly I am able to suppress the redundant switch instance and label each of them in the same manner as they appear in the product documentation.

FWIW, when I first set up the device, I got fooled by the redundant switch instance and was really confused why the switches weren't controlling the correct loads.

config/intermatic/pe653.xml
j9brown: Per prior comment, I'm listing the switching explicitly so as to suppress a redundant instance that should not be enumerated.

config/intermatic/pe653.xml
j9brown: This sensor is auto-discovered but it ends up having the incorrect label and units so I'm listing it explicitly to ensure the correct information is presented to the user.

config/intermatic/pe653.xml
j9brown: Ok, I will remove the others.

config/intermatic/pe653.xml
j9brown: Unfortunately this is not the case.  Auto discovery only detects instance index 1.  Also, it ends up with the wrong label and units so I'm listing both thermostats explicitly to ensure the correct information is presented to the user.  It took me some time to reverse engineer this.

config/intermatic/pe953.xml
j9brown: Unfortunately the config params are very complicated and undocumented.  This device is intended to be configured using the associated PE953 remote control.  Having watched the traffic between the two, there's a fair bit of manufacturer specific communication going on during that process.

The device doesn't support associations.

j9brown:
Hmm, so I just noticed something strange after making these changes and refreshing the node info.  I see 10 switches showing up now instead of 5.

I can resolve this by adding "endpoint" to the Instance declarations but it fails make xmltest.

    <Instance index="1" endpoint="1" label="Circuit 1" />
    <Instance index="2" endpoint="2" label="Circuit 2" />
    <Instance index="3" endpoint="3" label="Circuit 3" />
    <Instance index="4" endpoint="4" label="Circuit 4" />
    <Instance index="5" endpoint="5" label="Circuit 5" />

I must have removed these attributes prior to upload and failed to retest my change (sorry!).

What's the right way of solving this problem while also eliminating the redundant 6th switch endpoint from the enumeration?

j9brown:
Any thoughts on this?

Fishwaldo:
You probably need to add https://github.com/OpenZWave/open-zwave/wiki/CommandClass-Compatibility-Flags#multiinstance-commandclass MapRootToEndpoint in the config file - Without all the other CommandClass Info you have right now. Otherwise, please post a log file of the interview after issuing a refreshNodeInfo command.

j9brown:
Sounds good, I’ll try it but what does MapRootToEndpoint actually do?

PR #2290, opened 2020-07-03
[config] Fix Leviton Vizia RF+ device family.

j9brown:
This config merits discussion on how to handle SWITCH_MULTILEVEL_START_LEVEL_CHANGE and SWITCH_MULTILEVEL_STOP_LEVEL_CHANGE commands sent by the controller to the associated actuator.

More information here: https://groups.google.com/forum/#!topic/openzwave/PbCrQyQODTo

I've tentatively added COMMAND_CLASS_SWITCH_MULTILEVEL together with State/AfterMark compatibility directives to VRCS4 and VRCZ4 to indicate that the device sends these commands (but doesn't handle them itself).  These declarations fail "make xmltest" validation but they do work at runtime.  Moreover, OpenZWave doesn't know how to handle the incoming level changes (it just logs them).

I think we should break this PR into two parts.  (I'm including everything here for context.)

1) Submit the overall fixes to the Vizia RF+ family of devices, omitting the multilevel switch compatibility directives.
2) Figure out how we should handle those messages properly and amend the configs as required once that's done.

Thoughts?


config/leviton/vrcs2.xml
```diff
@@ -1,21 +1,45 @@
...
   </MetaData>
-  <CommandClass id="112"/>
+
+  <!-- Association Groups -->
+  <CommandClass id="133">
+    <Associations num_groups="2">
+      <!-- Each button toggles an individual scene on or off.  Multiple scenes can be active at once.
+           By default the device will associate groups 1 and 2 with the local loads.
+           Do not auto associate the controller with group 1. -->
+      <Group index="1" max_associations="232" label="Button 1" auto="false"/>
```
Fishwaldo, line 25: If you do not auto associate, how does the controller get updates about state change? Maybe all Groups should be auto=true (group 1 is implicit)

config/leviton/vrcs2.xml
```diff
@@ -1,21 +1,45 @@
...
+      <Group index="2" max_associations="232" label="Button 2"/>
+    </Associations>
+  </CommandClass>
+
+  <!-- Indicator Lights -->
+  <!-- Use these messages to override the indicator lights.
+
+       91 00 1D 0D 01 00 00 : Reset LEDs to locally controlled operation (default behavior).
```
Fishwaldo, line 33: Assuming I'm reading this correctly - its using the Manufactuter_Proprietary CC for Leds?

config/leviton/vrcs4.xml
```diff
@@ -0,0 +1,111 @@
...
+    </ChangeLog>
+  </MetaData>
+
+  <!-- Association Groups -->
+  <CommandClass id="133">
+    <Associations num_groups="4">
+      <!-- Each button toggles an individual scene on or off.  Only one scene can be active at once.
+           If the device has a switch, then by default it will associate group 1 with the local load.
+           Do not auto associate the controller with group 1. -->
+      <Group index="1" max_associations="232" label="Button 1" auto="false"/>
```
Fishwaldo, line 25: same comment as above.

config/leviton/vrcs4.xml
```diff
@@ -0,0 +1,111 @@
...
+       devices should reply with SCENE_ACTUATOR_CONF_REPORT (2C 03 <sceneId> FF 00) to confirm
+       the scene change.  Ignoring the message works too although the controller seems to ask
+       several times after each scene change until it gets the expected reply.
+
+       Scene messages interact with the indicator light change messages described above.
+       Turning on and off button LEDs determines whether the controller considers the
+       corresponding scenes to be active.
+  -->
+  <CommandClass id="38">
+    <State>
```
Fishwaldo, line 107: We need to port AfterMark to Compatibility rather than state.

config/leviton/vrcz4.xml
```diff
@@ -1,18 +1,111 @@
...
+    </ChangeLog>
+  </MetaData>
+
+  <!-- Association Groups -->
+  <CommandClass id="133">
+    <Associations num_groups="4">
+      <!-- Each button toggles an individual zone on or off.  Multiple zones can be active at once.
+           If the device has a switch, then by default it will associate group 1 with the local load.
+           Do not auto associate the controller with group 1. -->
+      <Group index="1" max_associations="232" label="Button 1" auto="false"/>
```
Fishwaldo, line 25: same comment as others.

config/leviton/vrcs4.xml
```diff
@@ -0,0 +1,111 @@
...
+       When no scene is assigned (default behavior), the controller sends BASIC_SET messages
+       to association groups 1 to 4 corresponding to the active button.
+       When a scene is assigned, the controller sends scene activations and level change messages
+       that offer richer control of the device's behavior.
+
+       The controller has 8 scene groups:
+         - Groups 1-4 are triggered when buttons 1-4 are pressed when their LED is not lit.
+         - Groups 5-8 are triggered when buttons 1-4 are pressed when their LED is already lit.
+
+       To configure the scenes, send SCENE_CONTROLLER_CONF_SET to associate scenes with
```
Fishwaldo, line 67: we don't support this CC.... so your saying a user should use the SendRaw to get this working?

config/leviton/vrr15.xml
```diff
@@ -0,0 +1,15 @@
...
+    <MetaDataItem name="OzwInfoPage">http://www.openzwave.com/device-database/001D:0334:1603</MetaDataItem>
+    <MetaDataItem name="ProductPic">images/leviton/vrr15.png</MetaDataItem>
+    <MetaDataItem id="0334" name="ZWProductPage" type="1603">https://products.z-wavealliance.org/products/812/</MetaDataItem>
+    <MetaDataItem id="0334" name="FrequencyName" type="1603">U.S. / Canada / Mexico</MetaDataItem>
+    <MetaDataItem id="0334" name="Identifier" type="1603">VRR15-1LZ</MetaDataItem>
+    <ChangeLog>
+      <Entry author="Jeff Brown - jeff.brown@gmail.com" date="29 June 2020" revision="1">Initial creation</Entry>
+    </ChangeLog>
+  </MetaData>
+</Product>
```
Fishwaldo, line 15: No Config/Association Groups?

Fishwaldo:
Regarding Aftermark - Only a few CommandClasses support it. You would need to update the COMMAND_CLASS_SWITCH_MULTILEVEL to also property support Aftermark:

1) Only expose ValueID's that are appropriate for reporting in a Aftermark CC. 
2) Make sure they are exposed as Read Only
3) override CommandClass::HandleIncommingMsg to decode SET messages and update appropriate ValueID's

Does this device advertise this as a Controlling CommandClass (rather than Controlled). If it does, you don't need to add it to the config file as a Aftermark Class. 

You can refer to the Basic CC to get some idea - Although that is also handling a lot of other corner cases as well.

config/leviton/vrcs2.xml
j9brown: That's correct.  Unfortunately it doesn't support the "Indicator" command class.

config/leviton/vrcs2.xml
j9brown: The scene controller and the loads are represented as different zwave nodes using multichannel encapsulation.

The scene controller has factory default associations with the local loads (relays) on the device.  There's no need to associate the primary controller unless you want the primary controller to respond to scene changes itself.  In my case, I just want the VRCS2 loads to be locally controlled (so they're very responsive).

The loads appear as separate binary switches.  They seem to get auto-associated (somehow?) with the primary controller so they do report their state changes as expected.

Aside: My old Vera automation system had a nasty habit of purging factory default associations when setting up a new device.  In the case of the VRCS2, it would remove the associations for the local loads and effectively break the switch behavior unless manually reconfigured.  It took ages to figure out what was happening there.

config/leviton/vrcs4.xml
j9brown: Similarly, the VRCS4 presents its load as a separate device and auto-associated it to button 1 as the factory default.  That said, 3 of the 4 buttons are useless out of the box unless additional associations are made.

So we could argue that auto-association with the primary controller is more essential for this one.  I've made those associations manually for my case along with other configurations documented in this file.

Unfortunately, openzwave currently doesn't know how to configure the scenes such that it can distinguish between each button and it doesn't handle incoming scene actuator commands either.  With that in mind, I don't think auto-association is a good idea just yet but I will defer to your judgement on the matter.

config/leviton/vrcs4.xml
j9brown: Yep.  At least for now.

config/leviton/vrcs4.xml
j9brown: That sounds good.  How about I remove this section from this patch for now and reintroduce it once the necessary command classes are implemented?

Alternatively, what would you think of encoding the "after mark" bit as an attribute of the CommandClass element itself because it fundamentally changes how the command class is interpreted?

config/leviton/vrr15.xml
j9brown: The configs are undocumented.  I haven't needed to poke them, assuming there are any.

The association groups are automatically detected just fine.  There's only one group and it reports the state of the controlled load.

j9brown:
> Regarding Aftermark - Only a few CommandClasses support it. You would need to update the COMMAND_CLASS_SWITCH_MULTILEVEL to also property support Aftermark:

Yeah, that makes sense.  I'll take a look at that project in a bit.  In the meantime, I managed to work around the problem by handling these unhandled messages outside of openzwave.  It works surprisingly well.  ;)

> Does this device advertise this as a Controlling CommandClass (rather than Controlled). If it does, you don't need to add it to the config file as a Aftermark Class.

Unfortunately it doesn't.  I checked.

> You can refer to the Basic CC to get some idea - Although that is also handling a lot of other corner cases as well.

Thanks!

j9brown:
Sorry I've been slow to take care of this because my local workarounds are working well enough.  ;)  I'll get on it in a bit.  After some reflection, I think I may actually implement some of the missing command classes.  We'll see.

PR #2294, opened 2020-07-06
Added support for WT00Z5-1 

natekspencer:
This is to add support for Linear/Nortek/GoControl WT00Z5-1 smart 3-way accessory dimmer switch to the device database.


Fishwaldo:
Thanks!

PR #2296, opened 2020-07-08
Added Nexia TH100NX Temperature and Humidity Sensor

Ankhislife:
Added TH100NX.xml under Nexia devices using information from https://products.z-wavealliance.org/products/2223.


Fishwaldo:
Thanks!

PR #2304, opened 2020-07-11
[Config Update] Kwikset Locks: Update default trigger refresh index

firstof9:
Unit testing shows 512/513 are the `alarm_type` and `alarm_level` reports that this lock uses instead of proper Notifications.

### DIFF
--- a/config/kwikset/smartcode.xml
+++ b/config/kwikset/smartcode.xml
@@ -33,6 +33,7 @@ SmartKey: Re-key the lock in seconds</MetaDataItem>
       <Entry author="Justin Hammond - Justin@dynam.ac" date="02 Jun 2019" revision="12">Updated Metadata Import from Z-Wave Alliance Database - https://products.z-wavealliance.org/products/2188/xml</Entry>
       <Entry author="Justin Hammond - Justin@dynam.ac" date="02 Jun 2019" revision="13">Updated Metadata Import from Z-Wave Alliance Database - https://products.z-wavealliance.org/products/2237/xml</Entry>
       <Entry author="Justin Hammond - Justin@dynam.ac" date="02 Jun 2019" revision="14">Updated Metadata Import from Z-Wave Alliance Database - https://products.z-wavealliance.org/products/2369/xml</Entry>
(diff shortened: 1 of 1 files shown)


mike240se:
just a heads up that on some locks (in my case the 888 which has its own xml so this wont cause an issue) the index of 512/513 are not always sent.  index 6 is sent 100% of the time no matter how the lock is unlocked/locked.  index 512/513 have a few cases where they are not triggered, not sure if that is a bug with my lock or not but just keep in mind incase you get reports of random issues where these locks arent updating after this fix.

firstof9:
> just a heads up that on some locks (in my case the 888 which has its own xml so this wont cause an issue) the index of 512/513 are not always sent. index 6 is sent 100% of the time no matter how the lock is unlocked/locked. index 512/513 have a few cases where they are not triggered, not sure if that is a bug with my lock or not but just keep in mind incase you get reports of random issues where these locks arent updating after this fix.

Right this is for the 910 and others that are using 512/513. If people report their kwikset lock isn't updating properly, I'll happily work with them to get it sorted and get another XML going for another model.

kphonik:
I'm having the same problem with the 914 - much tinkering with XML hasn't produced a working config so far. Unfortunately also having the same problem still with my YRD446 even after the recent fix.

firstof9:
> I'm having the same problem with the 914 - much tinkering with XML hasn't produced a working config so far. 

Have you tried adjusting the TriggerRefresh on your 914 in the cache? It'll be under `<CommandClass id="113" name="COMMAND_CLASS_NOTIFICATION">`

rccoleman:
My two Kwikset 914s are working fine with the change here (trigger on 513).  It was pretty flaky with ozwdaemon 1.53, but I now have my locks working with 1.50.

Fishwaldo:
So this has a conflicted with the recently merged #2297 - Seems we need to split out the config file for different devices?

firstof9:
> So this has a conflicted with the recently merged #2297 - Seems we need to split out the config file for different devices?

Ya that's why I put in the 916.xml PR, #2297 will break TriggerRefresh for 910 and 914 Kwikset locks. I can rebase and adjust, not a problem.

firstof9:
Please see #2326

PR #2307, opened 2020-07-12
Add support for FGDW-002 US model and DMEX1 extender

Madj42:
Add support for FGDW-002 US model and DMEX1 extender


Fishwaldo:
Thanks!

PR #2316, opened 2020-07-17
Fix TriggerRefreshValue for Schlage BE468 locks

cgarwood:
Index should be 6 instead of 0. This was previously fixed for the BE469 locks but not the BE468.


Fishwaldo:
Thanks!

PR #2319, opened 2020-07-20
[Config Update] Kwikset 916

firstof9:
## Changes
* Split the 916 locks from the smartcode
* Updated the 916 TriggerRefresh to index 6

Replace PR #2297


Fishwaldo:
Thanks!

PR #2320, opened 2020-07-22
Update Parameters for Zen20 2.03 

DelusionalAI:
The new 2.03 firmware adds some feature and new config parameters. There is unfortunately a slight hardware difference between 2.02 and 2.03, but they share the same ID.


Fishwaldo:
Thanks!

PR #2323, opened 2020-07-23
Added support for Graber MCZ1

natekspencer:
This is to add support for Graber/Bali/Springs Window Fashions/Somfy MCZ1 Multi-Channel Remote Control to the device database.


Fishwaldo:
Thanks!

PR #2326, opened 2020-07-24
[Config Update] Kwikset Locks: New 910/914 XML (non-zwave+ versions)

firstof9:
Replacement PR for #2304 

Unit testing shows 512/513 are the `alarm_type` and `alarm_level` reports that this lock uses instead of proper Notifications.

### DIFF
--- a/config/kwikset/910.xml
+++ b/config/kwikset/910.xml
@@ -1,7 +1,7 @@
 <!-- 
 Kwikset smartcode 
 http://s7d5.scene7.com/is/content/BDHHI/z-wave-configuration
---><Product Revision="15" xmlns="https://github.com/OpenZWave/open-zwave">
+--><Product Revision="1" xmlns="https://github.com/OpenZWave/open-zwave">
   <MetaData>
     <MetaDataItem name="OzwInfoPage">http://www.openzwave.com/device-database/0090:0001:0001</MetaDataItem>
     <MetaDataItem name="ProductPic">images/kwikset/smartcode.png</MetaDataItem>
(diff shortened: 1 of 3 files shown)


firstof9:
Please don't merge yet.

firstof9:
Re-verified this works on my 910 smartcode

rtrox:
This change, as discussed in #2304 and #2297, is really intended for the Smartcode 910. The 888, 914, and obsidian can all use index 6, which is a bit more standard, and has been more stable for me on the obsidian, and for @mike240se on the 888.

While I'm willing to create a separate config for the obsidian if needed (as we discussed in #2297), it's probably not good to just overwrite that change, which works for other devices, with this change. @Fishwaldo, do you mind helping determine what the right path forward is here? Do we need to create separate files for each device?

firstof9:
@rccoleman confirmed it worked with is 914 which uses the same product code as my 910. `0090 0001 0001` if the product code for the obsidian locks have a different product code, you'll want a separate XML file for it.

rccoleman:
To add to the fun, I replaced my older Z-Wave 914 locks with new Z-Wave Plus 914 locks and they report 512, 513, *and* 6 on each lock/unlock.  It looks like the new ZW+ 914 versions use 0x0090/0x0440/0x0003 for the "ProductCodes".

rtrox:
Right, this is the point I'm making here. I'm not sure where this comment is coming from:

> . 0090 0001 0001 if the product code for the obsidian locks have a different product code, you'll want a separate XML file for it.

There are 9 device ids that are using smartcode.xml:

https://github.com/OpenZWave/open-zwave/blob/master/config/manufacturer_specific.xml#L1073-L1081

And the majority of these devices (including the newer 914) accept index 6. While many of them will also accept 512, it isn't quite as stable, as pointed out in your first PR. While I don't mind creating a new file for the obsidian, it seems as though only one of the 9 device ids, and possibly only older versions of that device, require this change.

firstof9:
Looks like the XML test might need a little tweaking @Fishwaldo sorry

nechry:
thank @firstof9

PR #2328, opened 2020-07-24
added config params to kwikset 888

mike240se:
@MagmaiKH provided a manual with the config params and i used that to add the missing param block.


nechry:
thank @mike240se

PR #2330, opened 2020-07-28
Initial support for Leviton VRPD3-1LW Vizia RF+ 300 Watt Scene Capabl…

DaMadOne:


kelchm:
I have a VRPD3-1LW with a type id of 1705 (rather than 1706).

Fishwaldo:
Thanks!

PR #2331, opened 2020-07-29
[Config Update] Kwikset 910 - Add battery level update compatibility flag

firstof9:
Enable compatibility flag for battery level updates.


firstof9:
Looks like we can't make flags a default.

PR #2332, opened 2020-07-31
Several Config Parameter Fixes for LZW30, LZW30-SN, LZW31, LZW31-SN, LZW36

InovelliUSA:


stshontikidis:
Any status on this?

InovelliUSA:
The request seems to still be in queue to be looked at. I don't think there is anything else Inovelli needs to do, but let me know if this is not the case.

InovelliUSA:
I've added an update to the lzw36 config file to add config options for parameter 51 that is in firmware 1.36+

Myztillx:
Is there something else that needs to happen to get this merged? Would really like this update 😁

judeibe:
Looking forward to seeing the pull request merged in.

Fishwaldo:
Thanks!

PR #2343, opened 2020-08-03
Fixed iterator deletion

RDols:
After erase the iterator is undefined. Found these while looking for something else. Fixed them.


Fishwaldo:
Thanks!

PR #2352, opened 2020-08-11
Add initial support for Sunricher ZV2835RAC

zyberzero:
This will add support for the Sunricher ZV2835RAC.

For some reason this product is neither listed on Z-wave alliance webpage nor Sunrichers official webpage. I do however have one, and the data in the config file is  based upon the manual I got with the dimmer.


Fishwaldo:
Thanks!

PR #2353, opened 2020-08-12
add support for Iblinds 

mike240se:
The manufacturer recommends disabling getsupport:

"Most of the open-source controllers, including Home Assistant poll (send Z-Wave GET) immediately following a Z-Wave position command (Z-Wave SET) (see Z-Wave packet sniffer image below).   While this is fine for most devices such as dimmable light bulbs issuing a poll command immediately following the position command can cause iblinds to experience undesired behavior.  "

"Disable Get support by adding the getsupported="false" to the Command_Class_Switch_Multilevel Node Element"

From: https://support.myiblinds.com/community/home-assistant/configuring-home-assistant-and-iblinds/

But the great people in #zwave on the HA discord found an undocumented compatibility flag "NoRefreshAfterSet" so I changed the xml to use that instead of completely disabling get support as the issue is related to doing a get right after a set.  

The ST device handler solves hte issue by doing the get 15 seconds after the set.

WIth this new change it will allow someone to add to HA a check after set on a delay so that blinds position can still be "Get"

I also removed the auto generated values that i had in the original PR for the purpose of clarifying the help text as you said they shouldnt be included and auto generated instead, i only had them in there to improve the help text.

these blinds will function as a dimmer as their specific type is dimmer in their firmware but i cant fix that, if someone wants them to show as a cover in HA they would need to edit the cache file and change specific type to 6.


habhomegit:
@mike240se this is Eric from HAB Home Intel.  Great work finding NoRefreshAfterSet.  Please contact us when you have a chance.  We are working on version 3 of iblinds and would like see if we can get you to help us do some beta testing.   

Eric 
support@myiblinds.com

PR #2252, opened 2020-06-13
New Devices from TechniSat (shutter and single-switch)

mklooss:
Added these two devices:

Single-Switch/Relay
https://products.z-wavealliance.org/products/3677
Link (German): https://www.technisat.com/de_DE/Ausschalter-Einsatz,-BJ/352-22702-22935/

Shutter
https://products.z-wavealliance.org/products/3778
Link (German): https://www.technisat.com/de_DE/Rollladenschalter-Einsatz,-BJ/352-22704-22938/

PR #2357, opened 2020-08-15
Add New Device from TechniSat (Dimmer)

Technikte:
This is my first ever real commit, so I have no idea what I do here 100%. I did everything manual in Github Desktop and did no test or so, but calc the SHA512 of the xml and pasted it into the distfiles.mk. I hope everything looks fine and to bad I dont quite yet understand how to do the test in a windows env.
It is based on https://github.com/OpenZWave/open-zwave/pull/2252 from mklooss.

Dimmer
https://products.z-wavealliance.org/products/3798
Link (German): https://www.technisat.com/de_DE/Dimmer-Einsatz,-BJ/352-22705-22937/

It seems like I can't get it right with the revision sheme and the changelogs and so on. Maybe someone could correct that or if there a way to restart at revision 1 with correct hash and so on.


Fishwaldo:
Thanks!

Fishwaldo:
Thanks!

frsegers:
I'm working on the device configuration of the related TechniSat Double Switch. I compared my draft with your configuration file.
Unfortunately the type of "Parameter 2" and "Parameter 3" is incorrect.
The original manual indicates a size of 2 bytes so the type should be "short" in stead of "int" (which indicates 4 bytes).
It should have been noticed while testing it against the device. Using an "int" it will not process value updates of "Parameter 2" and "Parameter 3".

Technikte:
@frsegers I checked the doc at https://products.z-wavealliance.org/products/3798/configs and noticed that Parameter 2 + 3 are indeed wrong..that means we should check all technisat configs since I copied most part by @mklooss single switch.

This should be the answer when I understand the open-zwave docs correct. 
https://github.com/OpenZWave/open-zwave/wiki/Adding-Devices#creating-configuration-variables
2 Byte  Technisat == 16 Bit == 0 ... 65.535 == short  datatype open-zwave

PR #2361, opened 2020-08-16
Add new device Shenzhen Saykey Curtain Motor Controller

KingOfIce77:
Add new device Shenzhen Saykey Curtain Motor Controller
https://products.z-wavealliance.org/products/2662


KingOfIce77:
I don't understand the ConfigTest check error...

nechry:
thank @KingOfIce77

Issue #1239, opened 2017-06-02
Invalid thermostat set point on Secure SRT322 (or Horstmann HRT4-ZW)


dsoulayrol:
Hi,

I observe that, sometimes, changing the thermostat set point on Secure SRT322 results in the minimal value being displayed on the object (5°C) and a negative value returned on the network.

My application exposes the thermostat to the user as an object with *plus*, *minus* and *set* buttons. The first two augment or diminish the current set point by 0.5 steps, the last one sets a value provided by the user. Hence, in the attached logs, it is possible to see a lonely `ThermostatSetpointCmd_Set` command queued when the *set* button is used, or a series of them when *plus* or *minus* are pressed multiple times in a row.

[horstmann_setpoint.txt](https://github.com/OpenZWave/open-zwave/files/1047904/horstmann_setpoint.txt)

In the logs, the object has the ID 89. Its *Wake Up Interval* is set to 15 minutes. I have first successfully changed the set point to 28, and then pressed the minus button until I got 23. **open-zwave** accurately sends all the successive values until the last one:

> 2017-06-02 10:52:33.239 Info, Node089, Sending (WakeUp) message (Callback ID=0x61, Expected Reply=0x13) - ThermostatSetpointCmd_Set (Node=89): 0x01, 0x0c, 0x00, 0x13, 0x59, 0x05, 0x43, 0x01, 0x01, 0x21, 0xe6, 0x25, 0x61, 0x7c

But the returned value is -2.6:

> 2017-06-02 10:52:33.300 Detail, Node089,   Received: 0x01, 0x0c, 0x00, 0x04, 0x00, 0x59, 0x06, 0x43, 0x03, 0x01, 0x22, 0xff, 0xe6, 0xd2
> 2017-06-02 10:52:33.300 Detail, 
> 2017-06-02 10:52:33.300 Info, Node089, Response RTT 35 Average Response RTT 35
> 2017-06-02 10:52:33.300 Detail, Node089, Refreshed Value: old value=28.0, new value=-2.6, type=decimal

I then reproduced once the problem and, later, I used the *set* button to fix the same value. This time, the value is encoded differently in the request:

[...]

PR #2362, opened 2020-08-18
Honeywell TH8320ZW1000 thermostat requires 2 bytes for precision

automaton82:
I own this thermostat and trying to send the value with only 1 byte results in it dropping to the lowest value (10 degrees Celsius) if you send a decimal value (such as 23.5).

Setting the precision to 2 solves it. See #1239 for more information.


nechry:
thank @automaton82

PR #2364, opened 2020-08-19
Add Support for HEATIT Z-TRM3 / Thermofloor heatit058

Hatzl:
Add Support for Heatit Z-TRM3 wall thermostat.


sahansen:
Can't change set point.
Got it to work with Fibaro HC2 by enabling following associations (using internal temp. sensor):
EP0, GR1, SCA.
EP0, GR3, MCA.
EP2, GR2, MCA
If using external sensor EP3, GR2, MCA has to be enabled and with floor sensor EP4, GR2, MCA. 
If that helps.

Hatzl:
It's a little bit weird..
Changing set point via config parameters should work fine. Changing set point via user parameters is not working.
I think there is a bug in the firmware of the thermostat.
There is an unofficial fixed firmware [here](https://www.hjemmeautomasjon.no/forums/topic/6330-z-trm3-target-temperature-får-ikke-satt/?&page=3#comments) (commented by Chrske)

The flashing tutorial is for Z-TRM2 but it's the same on Z-TRM3.
After flashing you maybe need to change your temperature-sensor to the external one (in config parameters), and back to your internal.

For me that fixed it and I'm now able to change my set point via Home-Assistant.

I hope I could help you.

nechry:
thank @Hatzl

PR #2367, opened 2020-08-21
Add User code parameter to RF locking and unlocking notification event

zyberzero:
First of all; I'm not entirely sure that this is according to Z-Wave standards. However, my Id Lock 150 Z-Wave module sends the RFID user code this way (please see ID Locks documentation [here](https://idlock.se/wp-content/uploads/2019/08/IDLock150_ZWave_UserManual_v3.02.pdf) ).

### DIFF
--- a/config/NotificationCCTypes.xml
+++ b/config/NotificationCCTypes.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<NotificationTypes xmlns='https://github.com/OpenZWave/open-zwave' Revision="11">
-	<!-- Last Updated with Version 11 - 1/1/20 -->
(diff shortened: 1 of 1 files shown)


Fishwaldo:
According to the specifications, this is not supported, so I will not merge this. You should ask the vendor to update per the Specs. 

image

zyberzero:
> According to the specifications, this is not supported, so I will not merge this. You should ask the vendor to update per the Specs.
> 
> image

I haven't found exactly your source, could you please share where I can find it? (I do believe you, I just want to have that accessible for future reference).

kpine:
SDS13713 Notification Command Class, list of assigned Notifications

PR #2370, opened 2020-08-27
SwitchBinary CC to include payload length before assuming V2 message content

l3arcf:
Leviton DZ15S and maybe other switches that support Version 2, don't send a V2 report, but still send what appears to be a V1 report without the Target vale.  This checks the payload length as well as version. (the pattern currently used in Basic CC).

### DIFF
--- a/cpp/src/command_classes/SwitchBinary.cpp
+++ b/cpp/src/command_classes/SwitchBinary.cpp
@@ -107,7 +107,7 @@ namespace OpenZWave
 					// data[1] => Switch state
 					if (Internal::VC::ValueBool* value = static_cast<Internal::VC::ValueBool*>(GetValue(_instance, ValueID_Index_SwitchBinary::Level)))
 					{
-						if (GetVersion() >= 2)
+						if (GetVersion() >= 2 && _length == 4)
 							value->SetTargetValue(_data[2] != 0, _data[3]);
 						value->OnValueRefreshed(_data[1] != 0);
 						value->Release();
--- a/cpp/src/command_classes/SwitchBinary.cpp
+++ b/cpp/src/command_classes/SwitchBinary.cpp
@@ -107,7 +107,7 @@ namespace OpenZWave
 					// data[1] => Switch state
 					if (Internal::VC::ValueBool* value = static_cast<Internal::VC::ValueBool*>(GetValue(_instance, ValueID_Index_SwitchBinary::Level)))
 					{
-						if (GetVersion() >= 2 && _length == 4)
+						if (GetVersion() >= 2 && _length >= 4)
 							value->SetTargetValue(_data[2] != 0, _data[3]);
 						value->OnValueRefreshed(_data[1] != 0);
 						value->Release();
--- a/cpp/src/command_classes/Basic.cpp
+++ b/cpp/src/command_classes/Basic.cpp
@@ -148,7 +148,7 @@ namespace OpenZWave
 					if (!m_com.GetFlagBool(COMPAT_FLAG_BASIC_IGNOREREMAPPING) && m_com.GetFlagByte(COMPAT_FLAG_BASIC_MAPPING) != 0)
 					{
 						/* update our Mapped Class with the Target Values */
-						if (GetVersion() >= 2 && _length == 4)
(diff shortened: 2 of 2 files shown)


l3arcf:
I believe this will help solve the issue reported in qt-openzwave here: [https://github.com/OpenZWave/qt-openzwave/issues/146](url) which is wat I experienced, and possibly hassio-addons here: [https://github.com/home-assistant/hassio-addons/issues/1528](url).

My first PR, so I won't be surprised if I got some process thing wrong, but I tried!

cpp/src/command_classes/SwitchBinary.cpp
```diff
@@ -107,7 +107,7 @@ namespace OpenZWave
 					// data[1] => Switch state
 					if (Internal::VC::ValueBool* value = static_cast<Internal::VC::ValueBool*>(GetValue(_instance, ValueID_Index_SwitchBinary::Level)))
 					{
-						if (GetVersion() >= 2)
+						if (GetVersion() >= 2 && _length == 4)
```
kpine, line 110: I think the check should be:
```c++
if (GetVersion() >= 2 && _length >= 4)
```
If versions after 2 are ever introduced, the length will also increase. This would hold true for the code in `Basic.cpp` as well. See `Alarm.cpp` in contrast.

l3arcf: That makes sense.  I was hoping to keep my PR to one file to reduce the risk of a conflict with others.  I changed Basic CC too, and there was more than one place that needed >= rather than ==.  I built locally and seem to be OK, but I'm confident that Basic CC code is exercised more than just the issue I was tripping over in SwitchBinary.

All for the better though.

whitepa:
@l3arcf thank you for this... Home Assistant is now usable again for me!  I had updated to the OpenZWave beta, but alas I have dozens of the problematic Leviton DZ15S.  Rebuilding with this PR has stopped the infinite refresh loops.

l3arcf:
np @whitepa , I had enough DZ15S's and support from the community to be motivated! I'm hoping this, or an equivalent fix is in the next release so I can keep up-to-date too.

hyperbolic2346:
Do we know when this might make it in to a release? I'd prefer not building my own openzwave2mqtt if I can avoid it, but based on the age of this PR I'm not sure I'll be able to avoid it. What is the release schedule for ozw?

billimek:
I'm experiencing the same issue with a different model Leviton zwave switch (`DZPA1`).

When toggling the power on->off, the zwave controller starts dumping similar logs as above.

In addition,

* Even with zwave logging disabled, the following gets logged repeatedly without end in the zwave2mqtt docker container - it's about 5 entries/second
* Even with zwave logging disabled, the overall zwave controller performance starts to degrade such that the other devices stop responding or respond very slowly.
* The last known working version, as far as I know, was the openzwave 1.6.0 release.  Reverting to this (vis the `robertslando/zwave2mqtt` image tag `3.4.0`) is the only remedy that I could find.

image
image

whitepa:
Seems like a lot of people are independently running into this issue, and the more skilled of us in debugging are able to arrive at this PR and discussion thread.  I'm guessing there are a lot more people out there who have been impacted by this and do not have the skills to debug and end up abandoning OpenZWave (and potentially Home Assistant) as being completely unusable. (I personally know one individual in this situation).  Given the relative simplicity of this fix, I strongly urge the maintainers to take a look at this and consider for the next release, as it is no doubt having a widespread negative impact.

ChrisLizon:
> 
> 
> Seems like a lot of people are independently running into this issue, and the more skilled of us in debugging are able to arrive at this PR and discussion thread. 

I'm one of those people. I was almost ready to switch back to Home Assistants integrated OZW1.4. I was luckily able to exclude my Leviton DZPA1 since I only use it for Christmas lights, but I could see it being a big issue for others. I'm just hoping it's fixed by December...

OverZealous:
> I only use it for Christmas lights

I'm in the exact same boat, it's the switch I use to toggle the tree on and off. I don't use it the rest of the year right now.

Fishwaldo:
Thanks!

PR #2374, opened 2020-08-28
Add support for additional Sunricher devices

mgjbroadbent:
Support added for:
* SR-ZV9001-CCT-EU (Touch panel to control CCT controllers)
* SR-ZV9101FA (LED Strip Controller)
  This covers 3 variants that report as the same type, DIM, CCT and RGBW


nechry:
thanks' @mgjbroadbent

PR #2381, opened 2020-09-08
add support for Fibaro FGRGBW442 (RGBW Controller 2)

nouknouk:
Based on the (almost done) work of @brouwer : [add FGRGBW-442](https://github.com/OpenZWave/open-zwave/pull/2276)

## my setup at home: 
  - wired three independants white led strips on R,G,B channels ; W channel unused
  - no usage of inputs & nothing wired on them.

## works so far (for me):
- setting a light level for R,G,B works.
- running a 'programmed color sequence' (Fireplace, Storm, Rainbow, Police, ...) works.
- basic association with a nodon octan remote works.
- meter reporting works

#### switch_multilevel instances:
- 1 = ???
- 2 = main light level
- 3 = red level
- 4 = green level
- 5 = blue level
- 6 = white channel level __i guess__

To switch on light, I must set level of 'colors' via instances 3,4,5 + global level via instance 2.

#### energy reports:

They are reported on commclass 'meter', instance __2__ for me.
- Electric - kWh (comclassid: 0x32, index: 0, instance: 2)
- Electric - W (comclassid: 0x32, index: 2, instance: 2)

Instance 1, after a few hours, seem to remain at zero.

## todo:

- shouldn't config values 30 ... 34 rather be defined as bitsets ?

- not sure about several other comclasses that are present in zw alliance product page but don't appear as values added to my node by ozw lib. Normal ? (`Application Status`, `Association Group Information V2`, `Association V2`, `CRC16 Encapsulation`, `Device Reset Locally`, `Firmware Update Meta-Data V4`, `Multi-Channel Association V3`, `Multi-Channel V4`, `Notification V8`, `Supervision`, `Switch Color V3`, `Transport Service V2`)

## also:

- thanks btw, for the awesome work made by all contributors to this marvelous project
- also, please be patient & comprehensive will all the stuff I probably did poorly: it's my first PR ;-)


Whuzz:
"To switch on light, I must set level of 'colors' via instances 3,4,5 + global level via instance 2."

Shouldn't it be possible to set the colors with in one command? RRGGBBWW, so send #50505050 to set all colors to 50% level?

nouknouk:
> Shouldn't it be possible to set the colors with in one command? RRGGBBWW, so send #50505050 to set all colors to 50% level?

Setting the color in one command (so here the equivalent of the 4 calls to `switch_multilevel`, instances 3+4+5+6?) is the purpose of the `Switch Color` command class, as described in the [zwave specs](https://www.silabs.com/documents/login/miscellaneous/SDS13781-Z-Wave-Application-Command-Class-Specification.pdf) (§4.27).

So far with my first draft of device config, the Switch Color` command class is not exposed by open-zwave yet ; it's probably because a mistake or something missing in my XML file.

Regarding the brightness of the color itself, it needs to be managed via a separate command to the `switch_multilevel` command class according to zwave specs:

> The Color Switch Command Class manipulates the color components of a device. Each color component is scaled by the brightness level previously set by a Multilevel Switch Set or Basic Set command.


I have to dig a bit more into this. I suppose the FGMRGBW44**1** could be a good starting point to investigate.
As I'm brand new to adding device support in open-zwave, any help would be appreciated.

Regards.

nouknouk:
Hi, I found some interresting informations in the github repo of [Jeedom home automation software](https://github.com/jeedom), that seems to have implemented its own version of device config for openzwave + its own declaration of module's features in their app (thanks @nechry for this work, btw).

- the [custom openzwave device config](https://github.com/jeedom/plugin-openzwave/blob/beta/resources/openzwaved/config/fibaro/fgrgbwm442.xml)

- the [additional declarations](https://github.com/jeedom/plugin-openzwave/blob/beta/core/config/devices/fibaro_271/271.2306.4096_fgrgbwm442.rgbw.controller.json)

I'll dig into that.

Whuzz:
Ok, I did some testing with your config. First off: Great work! Was able to more in HA than ever before!

The steps I took: 
Made a clean docker instance (I use the QT allinone wrapper in docker with VPN to the GUI) and removed all Zwave integrations from HA. Edited the manufacrurer_specific.xml with your changes, so the devices would be recognised and pointed to correct XML. Then I added the fgrgbw422.xml (should't this be fgrgbwm.xml though? the "m" seems part of the device type) you made.

Devices recognised in OpenZwave. I can change the color of the led strips by selecting one of the "Color Index" presets (Blue, Magenta, Red, etc) under User Values  or by selecting one of the pre-programmed sequences (Police, Storm, etc) in the Config Values section. I can change the brightness of all of these colors by using Level in the User Values. 
I can't find any way to control the channels (R, G, B, W) separately though? Maybe this is a limitation in the QT wrapper's GUI? The only way to change any color is by using those presets. 

Then, I tried the HA OpenZwave(Beta) integration. Lo and behold... I can use the color picker there! Chosing any color results in the correct color in the strip with the R-G-B channels all being used. Also, I can set the white-channel with the "White Value" slider. This overrides the colorpicker though, so I can't set both a color AND the white value at the same time. Perhaps that is a limitation in HA? 
I can also control the brightness of either the RGB color, or the White by using the brightness slide.

Massive step forward 👍

nechry:
thanks' @nouknouk

nouknouk:
Thanks @nechry for the merge.
I will refine with pending open points as soon as I have some time.

PR #2382, opened 2020-09-08
Support for iblinds V3

chancehiblinds:
Added iblinds v3 to products and Version3.xml


danielbrunt57:
I would suggest line 854 should be:
<Product type="0004" id="0071" name="iblinds V3" config="hab/iblinds_v3.xml"/>
and the XML renamed and put under "hab" folder so the naming structure is consistent with other entries.

chancehiblinds:
Not sure why this was so difficult. Revision number kept being replaced with previous number when committing changes

nechry:
thanks' @chancehiblinds and @danielbrunt57

PR #2389, opened 2020-09-09
Update idlock150.xml

into-void:
Update as per 1.5.6 / 1.6 FW (https://idlock.no/kundesenter/idl150updater-version/)

Menu item nr 10 no longer mentioned in Z-wave manual so that is removed (https://idlock.no/wp-content/uploads/2020/08/User-manual-Z-Wave-modul_EN_1.1.pdf). 

First try on Github, no programming skills, I've most likely done something wrong.


nechry:
mostly good, you just mist to change the Revision element and the author attribut is require

nechry:
thanks' @into-void

PR #2390, opened 2020-09-09
Update Honeywell T6 Pro th6320zw2003 Precision to 2

kylezimmerman:
I own this thermostat and had to make this change in my config file to stop the thermostat from setting the temperature to 10 degrees Celsius when setting a decimal temperature like 21.5 degrees.

Very similar to this recent PR https://github.com/OpenZWave/open-zwave/pull/2362 that also applies to this model. See also https://github.com/OpenZWave/open-zwave/issues/1239.


nechry:
thanks' @kylezimmerman

PR #2391, opened 2020-09-10
Add GE 46202 Smart Switch

putnamnation:
This is the "GE Enbrighten Z-Wave Plus Smart Switch with QuickFit and SimpleWire" in classic switch form. It's the
new version of the GE 14292 smart switch.

I've used the 46201 config as a base. The config is the same between these versions, only the style of switch is different. Paddle vs the classic style.


nechry:
thanks' @putnamnation

PR #2392, opened 2020-09-15
Update GE 14294 device XML

kpine:
Remove duplicate metadata that was from a different product.

Add some undocumented configuration parameters. These are documented in newer revisions and work fine in this version. They should work with firmware version >= 5.26.

Interesting [discussion ](https://community.getvera.com/t/ge-14294-operation/197000/3) at the Vera community about these  undocumented parameters.

Thanks to KeithL in the HA community for [bringing this up](https://community.home-assistant.io/t/ozw-beta-can-t-set-parameter-not-in-device-xml/226588)!


nechry:
thanks' @kpine

PR #2395, opened 2020-09-18
updated fortrezz/wv01 and added econet/ezw1204

GeekGoneOld:
I didn't (couldn't) run xmltest or dist-update but ran an online xmllint.  All tested OK but HA can't test the links to product manual or product service.

Note: the fortrezz change (same device with different id and type) isn't actually in products.z-wavealliance.org with that id and type.  It is a brand labeled version sold as WaterCop by Dynaquip.  Maybe that version never got z-wave certification!

Note: econet addition is also available from 2 or 3 other manufacturers (brand labeled) as zw1404 (not ezw1404).  None of those manufacturers are in ozw.  I don't know who makes it but I put it under econet but had to change the mfg id for econet.


nechry:
thanks' @GeekGoneOld I just fix the error Number of Groups does not equal Group Entries

PR #2402, opened 2020-09-24
Fixed lifeline configuration of Luxy Switch (ZMNKAD)

petrovcicklemen:
An ordinary lifeline association was configured previously, whereas this product requires a multi channel lifeline configuration, for correct unsolicited message reporting.


petrovcicklemen:
Apologies. Tested the device on Home Assistant and thought, that it requires the added lines for the correct lifeline configuration. Only later realized it's running OpenZWave_v1.4, where those lines were required, but are not needed in OZW_v1.6.

PR #2404, opened 2020-09-25
Fix yrd110, yrd120 lock xmls

GeekGoneOld:
Moved these two devices out of PushButtonDeadbolt.xml and TouchDeadbolt.xml respectively.  These are zw300, not 500 and operate differently.


firstof9 (changes requested):
You need to bump the version number on line 2 of manufacturer_specific.xml.

https://github.com/OpenZWave/open-zwave/blob/cbdd3b91bbfc598e736a06fb2d8239976009639e/config/manufacturer_specific.xml#L2

GeekGoneOld:
Done.

At first I thought I could start the new file at revision at 1 but I thought about what you said and it seems obvious that the revision is not a file issue, it is a device (mfg, id, type) issue so I made the revision one higher than the higher of the two devices were previously.  Many thanks.

firstof9:
There you go, checks have passed 👍

firstof9 (approved):

Fishwaldo:
Thanks!

PR #2412, opened 2020-10-01
Fix config values for Aeotec Nano Shutter

matejdro:
* Trip times were decimal while they should be short
* Added actual list items to the repositioning and confirmation triggers

### DIFF
--- a/config/aeotec/zw141.xml
+++ b/config/aeotec/zw141.xml
@@ -58,10 +58,10 @@ Product Type ID: EU=0x00, US=0x01, AU=0x02 CN=0x1D
     <Value genre="config" index="22" label="Motor direction" size="1" type="list" units="" value="2">
       <Help>Set to toggle the motor running direction</Help>
     </Value>
-    <Value genre="config" index="34" label="Blade turn time" size="2" max="32767" min="10" type="decimal" units="0.01 seconds" value="150">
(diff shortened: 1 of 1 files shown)


matejdro:
Question: Is there a better way to define those one-shot triggers like repositioning and calibration configs here? Currently it is represented by two list items and to trigger reposition, you have to switch between those items and then switch again.

kpine:
> Question: Is there a better way to define those one-shot triggers like repositioning and calibration configs here? Currently it is represented by two list items and to trigger reposition, you have to switch between those items and then switch again.

I think that is usually done with the button type (which doesn't seem to be documented in the wiki). For example:
https://github.com/OpenZWave/open-zwave/blob/189b9050ee857bc177a26973cd99b81ebc9fcd92/config/aeotec/zw096.xml#L135

If you're using ozwdaemon, button values are not currently functioning properly.

matejdro:
Should I change it up to the button to be the most correct or leave it as list to keep compatibility with ozwdaemon?

nechry:
> Should I change it up to the button to be the most correct or leave it as list to keep compatibility with ozwdaemon?

Keep it as list even is not user-friendly, but rename default item to "Default".
Please also add a Change Log Entry for your revision

matejdro:
Made both updates.

Fishwaldo:
Thanks!

PR #2414, opened 2020-10-04
Include AU version of ZTS-500

JamesDodds:
The AU version has a type of 0202 instead 0200 but is otherwise identical. Added the necessary lines for it to be recognized.

I cannot find a differing product code or manual for the AU version.


Fishwaldo:
Thanks!

PR #2415, opened 2020-10-04
Initial support for Leviton ZW4SF

BrianWithAHat:
Adding initial support for the Leviton Decora ZW4SF 4 speed fan controller. 

Configuration values were taken from the z-wavealliance product listing and other preexisting Leviton Decora device configurations.


Fishwaldo:
Thanks!

PR #2422, opened 2020-10-17
Added Fibaro FGS-224

vifani:
I have bought the new Fibaro FGS-224 and I have seen it was not recognized on Home Assistant. So I have worked to add this device and I have also tested it. It works as expected


Fishwaldo:
Thanks

Issue #2403, opened 2020-09-24
Node values 'Lost' after 'Refresh Node Info' (Fibaro Wall Plug)


kvaster:
Recently I have added several Fibaro wall plugs in my system - FGWPE/F Wall Plug Gen5 (FIBARO System).

After initialization it was all OK - I was able to see following items under 'User':

* Switch
* Power
* Electric kWh
* Electric W
* Exporting

But after 'Refresh Node Info' I can see only following items:

* Switch
* Power

I'm using zwave2mqtt with open-zwave.

**open-zwave version:** commit 7eaae21b3248a82f841900a27ec9702d144f53cc

In ozwcache after adding node (before refresh):
```
<CommandClass id="50" name="COMMAND_CLASS_METER">
                                <Compatibility />
                                <State>
                                        <CCVersion>2</CCVersion>
                                        <InNif>true</InNif>
                                        <StaticRequests>2</StaticRequests>
                                </State>
                                <Instance index="1" />
                                <Value type="decimal" genre="user" instance="1" index="0" label="Electric - kWh" units="kWh" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="4.88">
                                        <Help></Help>
                                </Value>
                                <Value type="decimal" genre="user" instance="1" index="2" label="Electric - W" units="W" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="0.0">
                                        <Help></Help>
                                </Value>
                                <Value type="bool" genre="user" instance="1" index="256" label="Exporting" units="" read_only="true" write_only="false" verify_changes="false" poll_intensity="0" min="0" max="0" value="False">
                                        <Help></Help>
                                </Value>
                                <Value type="button" genre="system" instance="1" index="257" label="Reset" units="" read_only="false" write_only="true" verify_changes="false" poll_intensity="0" min="0" max="0">
                                        <Help></Help>
                                </Value>
                        </CommandClass>
```

And in ozwcache after node refresh command:

```
<CommandClass id="50" name="COMMAND_CLASS_METER">
                                <Compatibility />
                                <State>
                                        <InNif>true</InNif>
                                        <StaticRequests>2</StaticRequests>
                                </State>
                                <Instance index="1" />
                        </CommandClass>
```

Logs attached. Full one and filtered. Fibaro plug is node 16.
[...]

PR #2425, opened 2020-10-20
Update static request flags after receiving all secure command classes

kvaster:
We need to update static request flags only after receiving all secure command classes, cause sometime on node refres command class version is received at the end.

Fixes: #2403


Fishwaldo:
Thanks

PR #2433, opened 2020-10-24
Add support for Fibaro Door Window Sensor 2 Australia

nemccarthy:
Add support for Fibaro Door Window Sensor 2 Australia

https://products.z-wavealliance.org/products/2294?selectedFrequencyId=4


Fishwaldo:
Thanks!

PR #2439, opened 2020-10-28
Fix trane TZEMT400BB32MAA.xml

GeekGoneOld:
1. removed non-existent config params (id 161-199)
2. fixed name metadataitem of thermostat to correct model number
3. updated version and added revision history

It now boots quickly instead of timing out for the invalid config params.


Fishwaldo:
Thanks!

PR #2443, opened 2020-11-03
Add support for GE/Jasco 26932 In-Wall Smart Motion Dimmer

gongtao0607:
26932 Dimmer appears to be identical to 26933:
https://products.z-wavealliance.org/products/2093


Fishwaldo:
Thanks

PR #2444, opened 2020-11-03
Update Cooper RF9540-N device

gongtao0607:
Opposite to the spec, group 255 allows 0 associations is reported
from the device. Issue discussed here:
https://community.home-assistant.io/t/stuck-on-associations/136980/6


Fishwaldo:
Thanks

PR #2457, opened 2020-11-11
Jasco/GE 26931,26932: Add ForceInstance=false flag

gongtao0607:
Add ForceInstance=false flag to fix issue when adding association 1.0.
Without this flag, adding association '1.0' will become '1.1' silently.


gongtao0607:
Supposedly the same compatibility flags for 26932 (VerifyChanged and ForceInstance) are needed by 26933 as well, but I don't have a 26933 to test.

PR #2476, opened 2020-11-22
Add config for GE/Jasco 14322 In-Wall Smart Toggle Dimmer

brandon099:
This is adding support for the 14322 In-Wall Smart toggle dimmer. It is very similar to the already supported 14295, so if this is not the proper way to add support for this switch under these circumstances, let me know. Thanks!


hawk259:
I don't know if my PR will be approved, but it could tweak your PR as well if it is:

https://github.com/OpenZWave/open-zwave/pull/2473

This adds a COMMAND_CLASS_BASIC block to GE/Jasco switches that support double tap and makes Basic Set, usually treated as a Node Event, to be converted to a Value type.  Also defaults double tap association auto=true.

brandon099:
@hawk259 Thanks, I've subscribed to that PR as well so if that gets merged in first, I'll update this one.

PR #2478, opened 2020-11-23
Fix Zooz ZEN15 config value size

vvuk:
Zooz ZEN15 config param 152 is a 1-byte int (byte); in the config file it's listed as an int, and all changes to this param get ignored by the device because it's sent with a 4 byte size.  (This config file has `size` specified for all the byte/short/int values, even though it's ignored; I kept that in case it's convention?)

`make dist-update` also found some files that weren't in distfiles before, though I didn't touch them.  I can remove that commit if desired.


firstof9 (changes requested):
Bump the revision number in `manufacturer_specific.xml`

vvuk:
bumped -- missed that bit, sorry. also bonus DZPD3-1LW config.

firstof9:
Nice now just wait for the merge. 👍

PR #2488, opened 2020-12-01
Re-add Fibaro Flood Sensor with id 2003

rtclauss:
Commit 607dd7a931f20a747812d3009a9a51cb7db55f6a accidentally removed the newly added Fibaro Flood sensor with `id=2003` added in PR #2050. 

This commit re-adds the Fibaro Flood sensor variant to the `manufacturer_specific.xml`.


firstof9:
You'll want to bump the revision number in `manufacturer_specific.xml` too.

firstof9 (changes requested):
You'll want to bump the revision number in `manufacturer_specific.xml` too.

PR #2499, opened 2020-12-13
Update ZDB5100.xml

staal0:
According to: https://products.z-wavealliance.org/products/3399/configs

### DIFF
--- a/config/logicsoft/ZDB5100.xml
+++ b/config/logicsoft/ZDB5100.xml
@@ -25,8 +25,8 @@
       <Item label="No local operation of the dimmer" value="0"/>
       <Item label="Pushbutton 1 controls the dimmer" value="1"/>
       <Item label="Pushbutton 2 controls the dimmer" value="2"/>
-      <Item label="Pushbutton 3 controls the dimmer" value="3"/>
-      <Item label="Pushbutton 4 controls the dimmer" value="4"/>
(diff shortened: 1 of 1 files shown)


staal0 (commented):
According to: https://products.z-wavealliance.org/products/3399/configs

staal0:
@Fishwaldo - Question:  Device gets recognized as switch not as dimmer. Is it possible to ajust this in the xml? or is this device one of the 2-3% using non-standard implementation?

kpine:
> Question: Device gets recognized as switch not as dimmer

What's the context here? You don't see the dimmer command classes (multilevel switch) reported by the device, or your home automation software doesn't show it as a light?

staal0:
Home Assistant sees it as a switch - if generic value is changed from 24 (switch) to 17 (dimmer) inside the node tag it will become a dimmer. So my question is, if I can change anything in the xml to make it get recognized as a dimmer?

kpine:
> if generic value is changed from 24 (switch) to 17 (dimmer) inside the node tag it will become a dimmer.

You should submit an issue to the HA core project to support this. It's not changed in the XML.

staal0:
Yes, that sounds right - I thought that might be the case. It's been far too long since I've looked at this integration. I'll submit an issue to HA.

PR #2510, opened 2020-12-20
Update fgs224.xml

Romke10:
Added parameter 151 so it is possible to change the operating mode of the second channel


Romke10:
Why is the check failing?

PR #2515, opened 2020-12-24
Add support for Eaton RF9640-N Dimmer

sharmav39:


config/manufacturer_specific.xml
```diff
@@ -489,7 +489,9 @@
   <Manufacturer id="0132" name="DynaQuip Controls"></Manufacturer>
   <Manufacturer id="010D" name="e-Home AUTOMATION"></Manufacturer>
   <Manufacturer id="0247" name="EASY SAVER Co. Inc"></Manufacturer>
-  <Manufacturer id="001A" name="Eaton"></Manufacturer>
+  <Manufacturer id="001A" name="Eaton">
```
chensquared, line 492: Other Eaton devices (e.g. the 9540 predecessor) are under the Cooper manufacturer "001a" (lower case "a").  Consider putting the config there?

sharmav39: Eaton acquired Cooper in 2012, so the Z-Wave Alliance certification listings for newer products have no mention of Cooper.

https://products.z-wavealliance.org/products/3152/
vs
https://products.z-wavealliance.org/products/797/

I think it is more clear to have Eaton as the manufacturer and migrate older Cooper devices to the Eaton folder in the future as Eaton is now the owner of '001a'.

PR #2517, opened 2020-12-27
Add basic command class to Aeotec switches and dimmers

matejdro:
As discussed on https://github.com/OpenZWave/qt-openzwave/issues/60, only way to use get Basic events going forward is to add couple of flags to the devices that support this.

These three devices all support basic command independently from controlling the load (e.g. pressing the physical switch can trigger basic set independently of controlling the actual load connected to those switches) and thus I think it is useful for OZW to also report basic status independently from actual load status.

### DIFF
--- a/config/aeotec/zw111.xml
+++ b/config/aeotec/zw111.xml
@@ -2,7 +2,7 @@
 Aeotec ZW111 Nano Dimmer, base on Engineering Spec 3/6/2018 V2.02
 Product Type ID: EU=0x00, US=0x01, AU=0x02 CN=0x1D
 https://aeotec.freshdesk.com/helpdesk/attachments/6048080397
---><Product Revision="6" xmlns="https://github.com/OpenZWave/open-zwave">
+--><Product Revision="7" xmlns="https://github.com/OpenZWave/open-zwave">
   <MetaData>
     <MetaDataItem name="OzwInfoPage">http://www.openzwave.com/device-database/0086:006F:0103</MetaDataItem>
     <MetaDataItem name="ProductPic">images/aeotec/zw111.png</MetaDataItem>
@@ -21,6 +21,7 @@ The Nano Dimmer is also a security Z-Wave plus device and supports Over The Air 
       <Entry author="Justin Hammond - Justin@dynam.ac" date="02 Jun 2019" revision="4">Initial Metadata Import from Z-Wave Alliance Database - https://products.z-wavealliance.org/products/2095/xml</Entry>
       <Entry author="Justin Hammond - Justin@dynam.ac" date="02 Jun 2019" revision="5">Updated Metadata Import from Z-Wave Alliance Database - https://products.z-wavealliance.org/products/2135/xml</Entry>
       <Entry author="Justin Hammond - Justin@dynam.ac" date="02 Jun 2019" revision="6">Updated Metadata Import from Z-Wave Alliance Database - https://products.z-wavealliance.org/products/2149/xml</Entry>
+      <Entry author="Matej Drobnič - matejdro@gmail.com" date="27 Dec 2020" revision="7">Enable Basic command class</Entry>
     </ChangeLog>
     <MetaDataItem id="006F" name="ZWProductPage" type="0003">https://products.z-wavealliance.org/products/2135/</MetaDataItem>
     <MetaDataItem id="006F" name="Identifier" type="0003">ZW111-C</MetaDataItem>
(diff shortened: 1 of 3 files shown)


matejdro:
I'm not sure why PR check failed. From what I see it is caused by another file that is not changed in this PR. Can anyone with more knowledge of this system help here?

config/aeotec/zw111.xml
```diff
@@ -361,4 +362,10 @@ The Nano Dimmer is also a security Z-Wave plus device and supports Over The Air
       <Group index="4" label="Control Key2" max_associations="5"/>
     </Associations>
   </CommandClass>
+  <CommandClass id="32" name="COMMAND_CLASS_BASIC">
```
firstof9, line 365: Remove the `name="COMMAND_CLASS_BASIC"` part

config/aeotec/zw116.xml
```diff
@@ -325,4 +326,10 @@ The Nano Switch is also a security Z-Wave device and supports Over The Air (OTA)
       <Group index="4" label="Control Key2" max_associations="5"/>
     </Associations>
   </CommandClass>
+  <CommandClass id="32" name="COMMAND_CLASS_BASIC">
```
firstof9, line 329: Remove the `name="COMMAND_CLASS_BASIC"` part

config/aeotec/zw132.xml
```diff
@@ -355,4 +356,10 @@ The Dual Nano Switch is also a security Z-Wave device and supports Over The Air
       <Group index="4" label="Control Key2" max_associations="5"/>
     </Associations>
   </CommandClass>
+  <CommandClass id="32" name="COMMAND_CLASS_BASIC">
```
firstof9, line 359: Remove the `name="COMMAND_CLASS_BASIC"` part

firstof9 (changes requested):

config/aeotec/zw111.xml
matejdro: Any reason why this should be removed?

firstof9: `The attribute 'name' is not allowed.` per the check.

matejdro: Thanks for your help. Updated.

abmantis:
After getting this in Z2M, state reporting is completely broken for aeotec switches. When pressing the switch, on the log I get "No Valid Mapping for Basic Command Class and No ValueID Exported. Error?", and the state does not change.

kpine:
All of the recent changes to XMLs like this basically break any software that isn't ozwdaemon (qt-openzwave).

abmantis:
@kpine how would this work on qt-openzwave?

kpine:
It wouldn't work there either to be honest. Can you change your switch to send multilevel reports instead of basic?

matejdro:
I'm using this with qt-openzwave (whole reason for this change is that this is the only way to get qt-openzwave to support basic).

matejdro:
Huh, sorry, I had no idea this would break anything but qt-openzwave. 

@nechry should this be reverted or should other non-qt ozw instances be updated to support this?

kpine:
@matejdro It's not really your fault.

The thing with these Aeotecs in particular is that, as you originally said, the Basic Report signals the stage change of the actual switch. The Nano Switch for example, can use Basic CC to report off/on, which is mapped to the Binary Class. With these changes the switch value no longer changes state, instead only the basic values do. While you want to use basic set for the non-load switches, the override affects all the switches.

Another example, the other similar changes to the GE switches affect other non-ozwd software (z2m, Domoticz, is there anything else?) that would expect to be using Node Events instead of values. Only ozwd is written to utilize this behavior.

matejdro:
Okay, I see the issue now. It even happens on ozw-qt on my switches, so it's not limited to to the others.

Maybe the real solution here is to revert this change and make this into a per-device config option? That way anyone can pick which behavior they want from the switch.

kpine:
As of now there is no way to personally customize devices in this fashion outside of modifying the XML file. My preference has just been to maintain my own local copies. You can do this permanently by setting the Revision number in the XML to something like 999, so OZW won't download a version to replace it.

The same can be done to quickly revert to the previous behavior if you have access the the config files. Simply download the previous version (such as [this](https://github.com/OpenZWave/open-zwave/raw/72aebb704546d40f419180fd1dd39f789ef83152/config/aeotec/zw116.xml) for the nano switch) and bump the Revision to 999, and replace the existing one.

matejdro:
Then I vote to revert my PR. Adding support to a niche use case is not worth requiring to switch all notifications from Basic to Hail.

PR #2518, opened 2020-12-27
Add TKB HOME TZE96 thermostat

mcartoixa:
Add the [TKB HOME TZE96 thermostat](http://www.tkbhome.com/Z-WAVE-Color-Touch-Weekly-Programming-Heating-Thermostat_012_200.html). I own the device but the documentation is quite scarce.
This is my first contribution so mistakes may have been made: please double check I did not break anything 😁


mcartoixa:
Check fails because of missing image, but I double checked it is indeed part of the pull request. 🤷🏻‍♂️ Please advise.

PR #2524, opened 2021-01-03
fix: Correct SetValue exception message on Schedule and Button

13rac1:
Change the exception to a unique message indicating an error in the
SetValue() function, rather than GetValueFloatPrecision().

My logs contain many instances of the previous message from a Button:
```
Warning, Exception: Manager.cpp:2499 - 102 - ValueID passed to GetValueFloatPrecision is not a Decimal Value
```
I'm not sure if the error is occurring in GetValueFloatPrecision() or
SetValue(), either way this line needs to be corrected.


nechry:
thank @13rac1

PR #2541, opened 2021-01-17
Ring Alarm Contact Sensor: minor fix

gongtao0607:
Config 6 is misdocumented in z-wave alliance. The true values are
read(tested) from the device.


nechry:
thank @gongtao0607

PR #2545, opened 2021-01-18
Updated metadata and photos for Shenzhen Neo light switches

SantaFox:


SantaFox:
The check shows an error in file config/logicsoft/ZDB5100.xml that were not updated by me. However I took a look to it and bitmask seems to be properly set

nechry:
thank @SantaFox

Issue #2214, opened 2020-05-11
ZW095 Home Energy Meter Gen5 (AEON Labs) not updating / wrong initialization


magpern:
**Version**
Z2M docker dev

Build/Run method
- [X] Docker

Zwave2Mqtt version: 3.1.0
OpenZwave 1.6.1115

**Describe the bug**
I have just bought a ZW095 Home Energy Meter Gen5 (AEON Labs)
It registers fine in Z2M, but it never updates. It only updates if I manually select to refresh node info. Nothing in the log, nothing in MQTT. Total silence.

**To Reproduce**
Just include the node and wait. Nothing happens

**Expected behavior**
The device to update as expected according to settings.

**Additional context**
I have found how to make the device update. After hours and hours of googling.
I found it [here](https://community.openhab.org/t/solved-home-energy-meter-hem-g3-aeon-labs-gen5-zw095-do-not-report-automatically/44579/22) Austris_V post from Dec '18 (now called @billiaz )

It is something with the association groups that is not set up correctly.
I downloaded Z-Wave PC Controller and manually set the association groups (I think that is what I did)

![image](https://user-images.githubusercontent.com/6589760/81509386-89424880-930a-11ea-9cbd-de775297fe70.png)

After sending the command "85 01 0X 01" (where X is the Grouping identifyer) the device started updating according to the set number of seconds in setting 111 to 113

Somehow the device is not initializing correctly or if the xml file for the device is wrong.
I figured, if I need to send ASSOCIATION_SET on group 1,2 and 3 should not the xml file contain these groups?
It just contains a LIFELINE
[Link to xml settings](https://github.com/OpenZWave/open-zwave/blob/master/config/aeotec/zw095.xml)

Link to device inclusion process [here](https://paste.ubuntu.com/p/vJDSnBXwTr/)

magpern:
@Fishwaldo  It just struck me... if sending "85 01 0X 01" helps, HEX 85 is DEC 133
133 is the command class for associations. Doesn't that simply mean the definition of the device is wrong.

I can change the xml file for the device, if I just can figure out what is means.

"85 01 00 01"
"85 01 01 01"
"85 01 02 01"
"85 01 03 01"

What in above is assosiation num_group and what is index.
The label? Is that important?

```json
  <CommandClass id="133">
    <Associations num_groups="1">
      <Group index="1" label="Lifeline" max_associations="5"/>
    </Associations>
  </CommandClass>`

[...]

PR #2546, opened 2021-01-18
Update zw095.xml to configure ordinary lifeline association instead of multi channel lifeline association

hambudi:
Please see the discussion here https://github.com/OpenZWave/open-zwave/issues/2214 for more about this issue.

I have made this change locally and then added the device to my network and tested that I am seeing updates on the device now.


nechry:
thank @hambudi I just add bump rev number and add your change entry

PR #2547, opened 2021-01-18
Fixing heatit push button devices

paalkr:
Adding support for the Future Home AS white labeled version of the HeatIt Z-push button 8 device. Also fixing association groups for the HeatIt z-push devices.


nechry:
thank @paalkr

PR #2549, opened 2021-01-18
Initial config for Assa Abloy/Yale YRD136/156 deadbolt locks

cybergrimes:
Device has different config parameters from PushButtonDeadbolt.xml, some values inverted or options reduced and some parameters removed so new config needed for YRD136/156 deadbolts.


nechry:
thanks @cybergrimes

PR #2550, opened 2021-01-19
Fixes build issue with recent VS and reduces maintenance

dotMorten:
Fixes this:
image

Needed to bring the .net repo up to date


nechry:
thanks @dotMorten

PR #2552, opened 2021-01-20
Additional parameters for zen16

cybergrimes:
Add parameters 21 through 24 for device firmware 1.03 (https://www.support.getzooz.com/kb/article/356-zen16-multirelay-change-log/)
------------------
Firmware: 1.03

    Added the ability to change relays to NC (normally closed) and customize on/off reports based on the switch / relay position
    Added DC motor control mode which prevents R1 and R2 to be turned on at the same time


nechry:
thanks @cybergrimes

PR #2553, opened 2021-01-21
Update Max Value of Inovelli LZW45 Parameter 30

nathanfiscus:
The range for this parameter is too narrow.  Currently the max value accepted by the device is 197375.


nechry:
thanks @nathanfiscus

PR #2555, opened 2021-01-25
Add config for mcohome mh-dt411

rrozema:
Added config for mcohome MH-DT411, a cheap in-wall touch panel dimmer switch. The dimmer works rather well without the config, however it is better to disable automatic meter reports as the device rather chatty. Plus, setting config option 12 to 0 improves behavior in the user interface.


rrozema:
Sorry, don't know what else to change to make it pass the xmltest.

PR #2558, opened 2021-01-26
Support new Dome Mouser id version 0104

joelmoses:
There's a new hardware variant of the Dome Mouser z-wave pest trap out there. The device is functionally the same as the old 0086 id device but is now id 0104.


joelmoses:
I did not increment the root revision of manufacturer_specific.xml; I'm unsure of the protocol of iterating this value on an individual pull request adding a device file -- I would presume the maintainer iterates this when generating a release?

PR #2560, opened 2021-01-28
Added RF9601

knightzac19:
I recently acquired an RF9601 and found that its config didn't exist, I've imported the config over from ZWave's Official XML. I've verified the config works perfectly.


nechry:
tank @knightzac19

PR #2561, opened 2021-01-28
Removed Invalid RF9501 Group

knightzac19:
RF9501 has an invalid group in the config that causes the ZWave controller to go into an endless loop. I've removed the invalid group.


nechry:
great, thank @knightzac19

PR #2564, opened 2021-01-29
Added entry in manufacturers.xml for Fibaro Smart Module FGS-214 in EU

msbg:
This was required to correctly map our newly purchased FGS-214 switch with OpenZWave.  

Thanks!

Monica


nechry:
thank @msbg

PR #2568, opened 2021-01-31
Add entries for Zooz ZEN34 with different type code

cybergrimes:
Manufacturer sent information that a mass production of ZEN34 devices now have a new type code. No other changes.


nechry:
thank @cybergrimes

PR #2574, opened 2021-02-05
Fix zen23v3.xml parameter 19 to 12

FuzzyMistborn:
Zooz documentation indicates that Parameter 12 is the one responsible for controlling the 3-way switch type: https://www.support.getzooz.com/kb/article/317-zen23-on-off-toggle-switch-ver-4-01-advanced-settings/

Parameter 19 is for the Zen24, which is where I think the error came from.


cybergrimes:
I submitted the ZEN23 and ZEN24 at the same time, probably too much copy/paste. Sorry if it caused you issues.

FuzzyMistborn:
No worries! I wish they'd kept the parameters consistent. Only took me a little bit to figure the issue out.

nechry:
@FuzzyMistborn can you add the missing Change Log Entry with a a incremented revision number

FuzzyMistborn:
Think I did that right, apologies first PR here.

nechry:
great thank @FuzzyMistborn

PR #2575, opened 2021-02-05
Add settings for Fibaro FGT-001 firmware 4.7

markruys:
Firmware 4.7 adds a few more configuration settings.


nechry:
thank @markruys

markruys:
FYI, here is changelog of firmware 4.7 for Heat Controller from HC3:

- Function for reporting the demand for the heating medium enabling optimal control of the furnace,
- Heating system diagnosis function. Informing about problems with reaching the set temperature,
- Improved operation of the window opening detection function,
- Added the parameter for changing the direction of the thermostat knob.

PR #2578, opened 2021-02-09
Add Zooz ZST10 700 series USB controller

cybergrimes:
The new Zooz stick with 700 series chip is reporting with manufacturer 0000, type 0004 and id 0004. That manufacturer ID is in the list already for Sigma Designs but they don't have anything with id/type 0004 so I created a new Zooz entry with that manufacturer ID. I don't know what is the preferred way to handle such a thing. Maybe they'll get the manufacturer ID corrected on future production runs.

2021-02-09 09:57:36.850 Info, contrlr,     Manufacturer ID:      0x0000
2021-02-09 09:57:36.851 Info, contrlr,     Product Type:         0x0004
2021-02-09 09:57:36.851 Info, contrlr,     Product ID:           0x0004


nechry:
thank @cybergrimes

PR #2581, opened 2021-02-09
Initial config entries for Zooz ZEN71, ZEN76 and ZEN77 products

cybergrimes:
Config XML, manufacturer_specific.xml entry and product image for Zooz ZEN76 S2 On Off switch


nechry:
thank a lot @cybergrimes

PR #2582, opened 2021-02-10
Update eur_spiritz.xml

JasperPlant:
changed default setting of the backlight to off, because of battery drain.


nechry:
@JasperPlant please Bump the Revision Number and add an change log entry for it

cybergrimes:
@nechry I think this is a dupe of https://github.com/OpenZWave/open-zwave/pull/2583
@JasperPlant might have resubmitted on the config test error

nechry:
ok in this case I close it

PR #2583, opened 2021-02-10
Update eur_spiritz.xml

JasperPlant:
changed default setting of the backlight to off, because of battery drain.


JasperPlant:
Changed default behavior of backlight to off because of battery drain.

nechry:
thank @JasperPlant

PR #2588, opened 2021-02-22
Another id for MH9 CO2

kvaster:
I have two versions of MH9 CO2 at my place.
One version with id=3102 and type=0901.
Second one with id=5102 and type=0901.


nechry:
thank @kvaster

PR #2592, opened 2021-03-01
Initial config for Zooz ZEN32 Scene Controller, Zooz ZEN17 Universal Relay

cybergrimes:
Initial configuration from Zooz; I will add missing metadata as it becomes available.


nechry:
thank @cybergrimes

PR #2600, opened 2021-04-10
Fix errors in 1.10 firmware

markruys:
Firmware 1.10 (10/24/2017) changed the size of configuration parameters 0x29, 0x31 and 0x32 from 3 to 4 bytes. As you can update the firmware of this device over the air, it makes sense to update the config xml to reflect these changes.


nechry:
thank @markruys

PR #2607, opened 2021-05-28
Added new product ID 0000 for Simon S100 10002020 dimmer.

JordiAndujar:
Simon S100 dimmer 10002020-13X must have a new line with id="0000" along with "023a", because some dimmers have this ID assigned (similar case to Simon S100 Socket iO 10002034-13X)


nechry:
thank @JordiAndujar

PR #2617, opened 2021-09-20
Supervision CC support for SwitchBinary

nebuohyrrah:
Following up on pull request #2584 and #2616.
Made additions to support the supervision CC for the switch binary CC.


Crypto2:
Is SwitchBinary supposed to be working in the current master? I can't get it to work with my switched outlets (it reads their status fine but they don't respond to commands to turn on/off.) They work with Simplicity Studio just fine so I'm thinking it has to be a software issue.

nebuohyrrah:
Yes, for me it works without any problems for wall plugs that support the supervision command class, but also wall plugs that don't support the supervision cc.

Crypto2:
Mine look like they do support Supervision, not sure what it could be. It detects and reports them and their status fine, just nothing happens when I try to command them.

Issue #2613, opened 2021-07-29
Build errors at AssociationCommandConfiguration.cpp


TheNetopyr:
Dear all,

Since a few weeks Open-zwave won' t build, it error's out with the following error on a clean git clone:

Building src/command_classes/Alarm.cpp
Building src/command_classes/ApplicationStatus.cpp
Building src/command_classes/AssociationCommandConfiguration.cpp
/home/user/open-zwave-read-only/cpp/src/command_classes/AssociationCommandConfiguration.cpp: In member function ‘virtual bool OpenZWave::Internal::CC::AssociationCommandConfiguration::HandleMsg(const uint8*, uint32, uint32)’:

/home/user/open-zwave-read-only/cpp/src/command_classes/AssociationCommandConfiguration.cpp:191:85: error: ‘this’ pointer is null [-Werror=nonnull]
  191 |                                                                 group->ClearCommands(nodeIdx);
      |                                                                 ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
In file included from /home/user/open-zwave-read-only/cpp/src/Driver.h:36,
                 from /home/user/open-zwave-read-only/cpp/src/command_classes/CommandClass.h:36,
                 from /home/user/open-zwave-read-only/cpp/src/command_classes/AssociationCommandConfiguration.h:31,
                 from /home/user/open-zwave-read-only/cpp/src/command_classes/AssociationCommandConfiguration.cpp:29:
/home/user/open-zwave-read-only/cpp/src/Group.h:138:30: note: in a call to non-static member function ‘bool OpenZWave::Group::ClearCommands(uint8, uint8)’
  138 |                         bool ClearCommands(uint8 const _nodeId, uint8 const _endPoint = 0x00);
      |                              ^~~~~~~~~~~~~
/home/user/open-zwave-read-only/cpp/src/command_classes/AssociationCommandConfiguration.cpp:199:82: error: ‘this’ pointer is null [-Werror=nonnull]
  199 |                                                                 group->AddCommand(nodeIdx, length, start + 1);
      |                                                                 ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/user/open-zwave-read-only/cpp/src/Driver.h:36,
                 from /home/user/open-zwave-read-only/cpp/src/command_classes/CommandClass.h:36,
[...]

PR #2647, opened 2023-04-18
Fix incorrect null pointer check

slowriot:
After calling `node->GetGroup()`, the current version checks if `group` is null, and if true, proceeds to derefecence the null pointer.

This amends the code to perform as intended, only dereferencing the pointer if `group` is **not** null.


slowriot:
Resolves #2613

